PR #1, opened 2025-02-25
[CI/Build] fix installation issue where subpackages are missing

sducouedic:
This fixes vllm_spyre installation issue, where packages in `vllm-spyre/vllm_spyre` where missing, for two reasons:
1. `__init__.py` files were missing in the subpackages
2. the packages needed to be listed in `setup.py`, `setup()` function. The `packages` argument is supposed to list all the packages **and** subpackages that need to be installed, but instead it was listing `vllm_spyre` only. 

Note that the issue doesn't appear when vllm-spyre is installed in editable mode (ie. `pip install --editable .`), and this is the reason why this issue didn't appear previously.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 import os
 from typing import List
 
-from setuptools import setup
+from setuptools import find_packages, setup
 
 ROOT_DIR = os.path.dirname(__file__)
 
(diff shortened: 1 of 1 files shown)


tdoublep (approved):
LGTM

PR #4, opened 2025-02-27
[Bugfix] Fix broken import

joerunde:
Post v0.7.3 this class was renamed from `LoraNotSupportedWorkerBase` to `LoRANotSupportedWorkerBase`. This causes our import of it to fail, so this PR implements a try/catch on the import for backwards compatibility so that this doesn't break on future vllm releases.

See https://github.com/vllm-project/vllm/pull/13831

Tested on a local install with vllm@b382a7f28


tdoublep (approved):
LGTM

tdoublep:
Looks like the `yapf` check is failing here

yannicks1 (approved):
LGTM

PR #5, opened 2025-02-28
allow loading models from huggingface

pmur:
Test whether the requested model is a path or not. If not, acquire and use a cached model from huggingface.


tdoublep (approved):
LGTM - thanks for fixing this!

PR #7, opened 2025-03-03
Update deeptools logging options default

sducouedic:
This fixes [issue #253](https://github.ibm.com/ai-foundation/aiu-app-sw-tracker/issues/253).
From my understanding, using `tempfile.TemporaryDirectory()` is a typical way to export to user's temporary directory.

### DIFF
--- a/vllm_spyre/model_executor/model_loader/spyre_setup.py
+++ b/vllm_spyre/model_executor/model_loader/spyre_setup.py
@@ -42,7 +42,7 @@ def spyre_setup(rank=0, world_size=1, local_rank=0, local_size=1, verbose=False)
         os.environ["DEEPRT_EXPORT_DIR"] += f"/{rank}"
         os.environ["DTCOMPILER_EXPORT_DIR"] += f"/{rank}"
         sys.pycache_prefix=os.getenv("DEEPRT_EXPORT_DIR")+"/py-" + str(rank)
-    os.environ.setdefault("DTCOMPILER_KEEP_EXPORT", "1")
+    os.environ.setdefault("DTCOMPILER_KEEP_EXPORT", "0")
 
     # Inform Flex of the size of this job
     os.environ.setdefault("FLEX_RDMA_WORLD_SIZE", str(world_size))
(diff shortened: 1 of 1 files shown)


vllm_spyre/model_executor/model_loader/spyre_setup.py
```diff
@@ -36,13 +37,14 @@ def spyre_setup(rank=0, world_size=1, local_rank=0, local_size=1, verbose=False)
     # Each rank needs a unique space to write its binaries
     # For both 'export' and '__pycache'
     # https://docs.python.org/3/library/sys.html#sys.pycache_prefix
-    os.environ.setdefault("DEEPRT_EXPORT_DIR", "export")
-    os.environ.setdefault("DTCOMPILER_EXPORT_DIR", "export")
+    with tempfile.TemporaryDirectory() as exportdir:
+        os.environ.setdefault("DEEPRT_EXPORT_DIR", exportdir)
+        os.environ.setdefault("DTCOMPILER_EXPORT_DIR", exportdir)
```
tdoublep, line 42: Are we sure that the `exportdir` does not get deleted once we leave the `with` scope here?

sducouedic: The directory is not created, it is just a generation of a new path string inside user's temporary directory, that doesn't exist: eg. `/tmp/tmprngshl`.

From the understanding of the code, deeptool will then create that directory `/tmp/tmprngshl` if it doesn't exist yet. I should test that to be sure.

tdoublep: ok, I don't really understand why the `with` scope is needed here in that case

sducouedic: Ok, sorry you were right, the `temfile.TemporaryDirectory()` does create a temporary directory, that is cleared and removed from the filesystem after the context: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

So it appears that what I observed (I cannot seem to find that post anymore), is a hack use of that simply to generate a new path

yannicks1: This works, I tested it.

PR #6, opened 2025-02-28
V1 integration

joerunde:
@tjohnson31415 and I spent a while hacking through a new V1-compatible worker , runner, and scheduler. This works on an AIU chip! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉 

The main blockers here are:
- The scheduler is not yet pluggable (Woosuk is working on it now https://github.com/vllm-project/vllm/pull/12544/files)
- It doesn't look like there's support for ignoring requests when they first come into the scheduler. (The V1 engine always tries to run them anyway)

We'll push up a draft PR with some temporary changes for vLLM for the above blockers so we can at least move development forward here on v1

PR #11, opened 2025-03-06
[V1] tests

yannicks1:
### V1 tests
Adding tests for V1 alongside the existing V0 tests.

-> This PR can be merged while still working on #17, #24 and #29

**Further TO DOs**
- [ ] Add V1 support for `test_spyre_seed.py`: Sampling temperature has to be passed instead of hard coded [here](https://github.com/vllm-project/vllm-spyre/pull/6/files#r1983384389) (open PR #24)
- [ ] Add V1 support for `test_spyre_embeddings.py` (open issue #17)
- [ ] Add V1 support for `test_spyre_max_prompt_length.py` (open issue #29)
- [x] Add V1 support for `test_spyre_tensor_parallel.py` (open issue #25)
- [x] Investigating issue where tokens are prepended with Ġ instead of a whitespace (see comments [1](https://github.com/vllm-project/vllm-spyre/pull/11#issuecomment-2706872622) and [2](https://github.com/vllm-project/vllm-spyre/pull/11#issuecomment-2711847556))
- [x] Investigating hanging V1 engine in pytests when compiling multiple configurations (see comments [1](https://github.com/vllm-project/vllm-spyre/pull/11#issuecomment-2711244246) and [2](https://github.com/vllm-project/vllm-spyre/pull/11#issuecomment-2711368047))

### DIFF
--- a/.github/workflows/test-spyre.yml
+++ b/.github/workflows/test-spyre.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Run Spyre tests within docker container
       run: |
         docker run -i --rm --entrypoint /bin/bash vllm-spyre -c '''
-          pip install pytest sentence-transformers && \
+          pip install pytest sentence-transformers pytest-timeout pytest-forked && \
           python -c "from transformers import pipeline; pipeline(\"text-generation\", model=\"JackFram/llama-160m\")" && \
           export VARIANT=$(ls /root/.cache/huggingface/hub/models--JackFram--llama-160m/snapshots/) && \
           mkdir -p /models && \
@@ -24,5 +24,5 @@ jobs:
           export MASTER_ADDR=localhost && \
           export DISTRIBUTED_STRATEGY_IGNORE_MODULES=WordEmbedding && \
           cd vllm-spyre && \
-          python -m pytest tests -v -k eager
+          python -m pytest --forked --timeout=300  tests -v -k eager
         '''
--- a/Dockerfile.spyre
+++ b/Dockerfile.spyre
@@ -16,7 +16,15 @@ RUN ln -sf $(which python${PYTHON_VERSION}) /usr/bin/python && \
     ln -sf $(which pip${PYTHON_VERSION}) /usr/bin/pip
 
 # Download and install vllm ###########################################################
-RUN pip install vllm==0.7.3
+RUN git clone --depth 1 https://github.com/vllm-project/vllm.git \
+    && cd vllm \
+    && git fetch origin pull/14242/head:spyre-workarounds \
+    && git checkout spyre-workarounds \
+    && python -m pip install --upgrade pip \
+    && pip3 install torch=="2.5.1+cpu" --index-url https://download.pytorch.org/whl/cpu \
+    && python use_existing_torch.py \
+    && pip install -r requirements-build.txt \
+    && SETUPTOOLS_SCM_PRETEND_VERSION=0.7.3 VLLM_TARGET_DEVICE=empty pip install --verbose . --no-build-isolation
 
 # Install vllm Spyre plugin ##################################################################
 RUN mkdir /workspace/vllm-spyre
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -47,7 +47,7 @@ def generate_spyre_vllm_output(model: str, prompts: List[str],
     for req_output in vllm_outputs:
         result = {}
         result['text'] = req_output.outputs[0].text
-        result['token_ids'] = req_output.outputs[0].token_ids
+        result['token_ids'] = tuple(req_output.outputs[0].token_ids)
         result['tokens'] = tuple([
             req_output.outputs[0].logprobs[i][t].decoded_token
             for i, t in enumerate(result['token_ids'])
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -19,8 +19,8 @@ def generate_spyre_vllm_output(model: str, prompts: List[str],
                                max_model_len: int, block_size: int,
                                sampling_params: Union[SamplingParams,
                                                       List[SamplingParams]],
-                               tensor_parallel_size: int,
-                               backend: str) -> List[Dict[str, Any]]:
+                               tensor_parallel_size: int, backend: str,
+                               vllm_version: str) -> List[Dict[str, Any]]:
 
     warmup_prompt_length = [t[0] for t in warmup_shapes]
     warmup_new_tokens = [t[1] for t in warmup_shapes]
@@ -33,6 +33,7 @@ def generate_spyre_vllm_output(model: str, prompts: List[str],
     os.environ['VLLM_SPYRE_WARMUP_BATCH_SIZES'] = ','.join(
         str(val) for val in warmup_batch_size)
     os.environ['VLLM_SPYRE_DYNAMO_BACKEND'] = backend
+    os.environ['VLLM_USE_V1'] = "1" if vllm_version == "V1" else "0"
 
     vllm_model = LLM(model=model,
                      tokenizer=model,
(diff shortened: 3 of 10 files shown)


joerunde:
@yannicks1 there are also the changes required to vLLM to make the scheduler pluggable and to support rejecting requests from the scheduler. We would need to somehow install those changes to run any tests that need to use our scheduler in order to pass. Or, we could keep those disabled until we get support merged upstream

yannicks1:
@joerunde I thought the idea was to point to your vLLM [fork](https://github.com/vllm-project/vllm/pull/14242) in both our development setup and the testing pipeline. We can simply switch between V0 and V1 via the env variable `VLLM_USE_V1` within the same pod/container without having to reinstall anything. The only caveat is that we should periodically sync your fork with vLLM main. Once the vLLM community supports a pluggable scheduler in V1 we will point to the main branch again.

sducouedic:
Hey @yannicks1, I believe we should keep V0 tests using the main branch of `vllm-project/vllm` and use @joerunde fork for V1 tests. Here is what  I would suggest to do:

- create a new dockerfile `Dockefile_v1.spyre` that builds with the @joerunde 's fork version of vllm
- in `.github/workflows` create a new yml file named `test-spyre-v1.yml`. This test suit will build an image with `Dockerfile_v1.spyre` and set the `VLLM_USE_V1=1`. See [how this is done](https://github.com/vllm-project/vllm-spyre/blob/8ec9c94167384709aebb5bbfd00b019bdbb2a2b2/.github/workflows/test-spyre.yml#L11) in `test-spyre.yml`

This allow us to keep V0 tests completely independent from V1 and still working with vllm installed from Pypi.

yannicks1:
I see your point, but my motivation is to ensure the same setup for our internal development environment and the testing pipeline. Otherwise commits would trigger tests run with a different vLLM version than we use internally for development.

yannicks1:
Copy pasting @joerunde 's findings here, as dealt with in this PR:

> I do see the tokens prepended with Ġ instead of a whitespace, but maybe that's an expected quirk of the vllm tokenizer?

```
   token id. token               logprob         token id. token               logprob
HF:       11 ','                 -1.742141  vLLM:       11 ','                 -1.735718  
HF:      358 ' I'                -1.747389  vLLM:      358 'ĠI'                -1.784092  
HF:     2846 "'m"                -1.320623  vLLM:     2846 "'m"                -1.332726  
HF:      264 ' a'                -2.389239  vLLM:      264 'Ġa'                -2.374791  
HF:     2697 ' little'           -2.544920  vLLM:     2697 'Ġlittle'           -2.516528  
HF:    22568 ' confused'         -0.916666  vLLM:    22568 'Ġconfused'         -0.902765  
HF:      922 ' about'            -0.970884  vLLM:      922 'Ġabout'            -0.959876  
HF:      279 ' the'              -1.210310  vLLM:      279 'Ġthe'              -1.203339  
HF:     6811 ' difference'       -3.076489  vLLM:     6811 'Ġdifference'       -3.043910  
HF:     1990 ' between'          -0.090222  vLLM:     1990 'Ġbetween'          -0.092823  
HF:      279 ' the'              -2.121253  vLLM:      279 'Ġthe'              -2.086390  
HF:     1403 ' two'              -2.014982  vLLM:     1403 'Ġtwo'              -2.029896  
HF:     3878 ' terms'            -1.944602  vLLM:     3878 'Ġterms'            -1.945674  
HF:      330 ' "'                -1.303640  vLLM:      330 'Ġ"'                -1.307465  
HF:       67 'd'                 -4.424416  vLLM:       67 'd'                 -4.442968  
HF:     1056 'iss'               -2.718370  vLLM:     1056 'iss'               -2.716669  
HF:     3294 'olution'           -0.849339  vLLM:     3294 'olution'           -0.865022  
HF:        1 '"'                 -0.324868  vLLM:        1 '"'                 -0.322343  
HF:      323 ' and'              -0.050900  vLLM:      323 'Ġand'              -0.053666  
HF:      330 ' "'                -0.015193  vLLM:      330 'Ġ"'                -0.014664  

logprob absolute differences: average=0.013078  maximum=0.036704
logprob relative differences: average=0.013513  maximum=0.054343
```

> Another thing I noticed is that the tests with multiple warmup shapes fail to compile.

yannicks1:
Regarding
> Investigating issue where tokens are prepended with Ġ instead of a whitespace

It looks like this is indeed an issue of the upstream **V1** detokenizer: The `outputs.logprobs.decoded_token` are off while the `outputs.token_ids` and `outputs.text` are correct. Since we only use the `outputs.logprobs.decoded_token` for printing, but do our comparison on based on`outputs.token_ids`, we can ignore it for now.

yannicks1:
Regarding

> Another thing I noticed is that the tests with multiple warmup shapes fail to compile.

I was able to reproduce the failure when more than one configurations are tested in the same pytest. The program hangs after the first compilation with V1. It looks like the (multi-threaded) vLLM V1 engine does not get shut down and cleaned up properly after it is not used anymore. 

Note: Funnily enough, the script `test_spyre_basic.py` (backend eager) works for multiple configurations with a V1 engine when only calling `generate_spyre_vllm_output()` (not computing hugging face model output), but in combination with `generate_hf_output()` it hangs itself after the first configuration.

joerunde:
> @joerunde I thought the idea was to point to your vLLM [fork](https://github.com/vllm-project/vllm/pull/14242) in both our development setup and the testing pipeline. We can simply switch between V0 and V1 via the env variable `VLLM_USE_V1` within the same pod/container without having to reinstall anything. The only caveat is that we should periodically sync your fork with vLLM main. Once the vLLM community supports a pluggable scheduler in V1 we will point to the main branch again.

Yup! I was just wondering how this would work with the public CI here since this runs with vllm 0.7.3: https://github.com/vllm-project/vllm-spyre/blob/main/Dockerfile.spyre#L18-L19
But I agree with your other points, it's probably fine to keep this repo installing off of released versions of vllm and disable tests here that require unreleased support. Our internal testing pipeline can continually test the newer vllm changes.

> I was able to reproduce the failure when more than one configurations are tested in the same pytest. The program hangs after the first compilation with V1. It looks like the (multi-threaded) vLLM V1 engine does not get shut down and cleaned up properly after it is not used anymore.

@dpatel-ops had a workaround for this, using `--forked` will run the tests in separate processes so that they don't pollute global state. There's a lot of work in-flight to make the v1 shutdown cleaner but it's been in limbo for a long time while most of the focus is on feature parity so that users can actually make the switch.

tjohnson31415:
> Investigating issue where tokens are prepended with Ġ instead of a whitespace
>
> The outputs.logprobs.decoded_token are off while the outputs.token_ids and outputs.text are correct

As I understand it, the `Ġ` replacement is a quirk of some BPE tokenizers, but it is equivalent to having the token prepended with a whitespace. If you look in the [tokenizer.json for granite-3.1-8b-instruct](https://huggingface.co/ibm-granite/granite-3.1-8b-instruct/raw/main/tokenizer.json), the "real" token texts have `Ġ` instead of whitespace. Depending on how you display/stringify the token, it may or may not convert `"Ġ"` to `" "`:
```
>>> tokenizer.decode([25697])
' IBM'
>>> tokenizer.convert_ids_to_tokens([25697])
['ĠIBM']
```
vLLM uses `convert_ids_to_tokens` within its [`detokenizer_utils.py`](https://github.com/vllm-project/vllm/blob/main/vllm/transformers_utils/detokenizer_utils.py).

yannicks1:
did someone look into `test_spyre_tensor_parallel.py` yet and knows what's required to make it work?

joerunde:
> did someone look into test_spyre_tensor_parallel.py yet and knows what's required to make it work?

I'm not sure, I can open an issue for it so we can keep track of it. Should we aim to merge this PR in before tensor parallel and embeddings are supported? I think embeddings might take a while, since that will require upstream support for embeddings in v1 as well.

yannicks1:
I updated the Dockerfile.spyre to install vllm from @joerunde upsream PR#14242  spyre-workarounds. I then made sure to exclude failing V1 test (max prompt length) and include running ones (tensor parallelism). There is now open issues for all remaining failing V1 tests (#17, #24, #29). 

-> ready to merge!

joerunde (approved):
Nice!

🤞  it won't be too long to get all the tests enabled

PR #22, opened 2025-03-10
:children_crossing: Port hf download fix to V1

joerunde:
This ports the fix from #5 to the V1 worker.

I tested that this can download and run a small model from HF, on both v0 and v1. Tested with `Cheng98/llama-160m` and `JackFram/llama-160m`


tjohnson31415 (approved):
Validated that it works now. LGTM!

Issue #10, opened 2025-03-06
[V1] Track sampling params correctly in the worker


joerunde:
In `prepare_model_input` in the v1 model runner, we currently hardcode the `SamplingMetadata`. We should instead save the sampling params that are provided when a request is first scheduled, and use them on subsequent decode iterations.

The model runners are now expected to save state about currently running requests, so that the scheduler does not need to continually broadcast the full state of a scheduled request. See the upstream v1 `gpu_model_runner` and how it saves info off of `scheduled_new_requests` and continually updates it on each iteration.

wallashss:
Hey @joerunde,

I could not find the `scheduled_new_requests` in `vllm/v1/worker/gpu_model_runner.py` on upstream. Perhaps the code changed or am I looking the wrong thing?

joerunde:
@wallashss `scheduler_output.scheduled_new_reqs` should be the new requests coming into the batch: https://github.com/vllm-project/vllm/blob/8ed5421aaa7da24051acdae53c860e6ce6598403/vllm/v1/worker/gpu_model_runner.py#L317-L338

For the spyre model runner, we currently don't support continuous batching so we just get the data about the requests in the batch once here: https://github.com/vllm-project/vllm-spyre/blob/main/vllm_spyre/v1/worker/spyre_model_runner.py#L171-L174

I already had to do a bit of state saving to get the logprobs working correctly, it would be good to have some encapsulation like the gpu runner's `CachedRequestState` and `InputBatch` to track data about the running requests

wallashss:
Merged on #24

PR #24, opened 2025-03-12
[V1] Fix track sampling params correctly in the worker

wallashss:
This PR remove the hardcoded sampling parameters for `v1/spyre_model_runner`.

I basically copied the gpu_model_input from the upstream V1 and did a basic integration to the spyre_model_runner.

I copied because, we may probably do updates to those new classes to add features to our specific needs and to not depend from upstream code that might break the model runner from the spyre side. Also I think it move toward the implementation to add continuos batch.

Address #10


tests/test_spyre_input_batch.py
```diff
@@ -0,0 +1,254 @@
...
+from typing import Optional
+
+import numpy as np
+import pytest
+import torch
+
+from vllm.sampling_params import SamplingParams
+from vllm.utils import is_pin_memory_available, make_tensor_with_pad
+from vllm.v1.sample.metadata import SamplingMetadata
+from vllm.v1.worker.gpu_input_batch import CachedRequestState, InputBatch
```
joerunde, line 13: I think this is importing the wrong `CachedRequestState` and `InputBatch`, we'd want the ones from the plugin

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -109,6 +113,20 @@ def __init__(
         # mapping of request ID to sampling params
         self._sampling_params_by_request: dict[str, SamplingParams] = {}
         self._max_logprobs: Optional[int] = None
+        
+        self.requests: dict[str, CachedRequestState] = {}
```
joerunde, line 117: These should be unused now, right?
```
        self._sampling_params_by_request: dict[str, SamplingParams] = {}
        self._max_logprobs: Optional[int] = None`
```

Also these three attributes are holding state information about the batch:
```
        # position_ids of all the sequences in current batch
        self._position_ids: torch.Tensor = None
        # attention masks of all the sequences in current batch
        self._mask: torch.Tensor = None
        # mapping: request id to index in batch
        self._req_ids2idx: dict = {}
```
It would be great to get that data from `InputBatch` so that we don't have to track state in multiple places

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -305,45 +319,180 @@ def prepare_model_input(
...
+                req_id, ())
+            if spec_token_ids:
+                start_index = end_token_index
+                end_token_index += len(spec_token_ids)
+                self.input_batch.token_ids_cpu[
+                    req_index, start_index:end_token_index] = spec_token_ids
+            # NOTE(woosuk): `num_tokens` here may include spec decode tokens.
+            self.input_batch.num_tokens[req_index] = end_token_index
+
+        # Check if the batch has changed. If not, we can skip copying the
```
joerunde, line 462: I don't know if any of this logic about continuous batching can be here since we don't support it 🤔 

We at least don't need to add it in while it's unsupported and the scheduler only gives us one batch at a time

tests/test_spyre_input_batch.py
wallashss: Good catch! I'll fix it. Thanks.

joerunde:
In general we might want to consider stripping out all of the kv-cache and continuous-batching related parts of `InputBatch` and `CachedRequestState` here, since they're not yet supported on spyre and I don't know what future support might look like. I have a slight preference for not including code that will be unused, to avoid confusion

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: Ok, I'll clean up. Thanks for the feedback.

yannicks1 (commented):
Hi guys, 
I don't have the bandwidth to do a full review reading through all the code, but I quickly tested with `tests/test_spyre_seed.py` for V1 after I fixed the `Dockerfile.spyre` yesterday. I basically did combine this branch `wallashss:track_sampling_params` with `ysc-V1-tests` and enabled V1 tests for `test_spyre_seed.py` (#30 ). The tests ran successfully  🎉 After someone has reviewed the code, we can merge safely!

tdoublep:
> I have a slight preference for not including code that will be unused, to avoid confusion

I agree with this, the less unused code the better.

joerunde:
@wallashss I think you'll need to update `tests/test_spyre_seed.py` to enable the test with `V1` as part of this PR

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -0,0 +1,467 @@
...
+@dataclass
+class CachedRequestState:
+
+    req_id: str
+    prompt_token_ids: list[int]
+    prompt: Optional[str]
+    sampling_params: SamplingParams
+    generator: Optional[torch.Generator]
+
+    block_ids: list[int]
```
joerunde, line 27: We're not doing any kv cache management right now, so I don't think we should pull in anything related to block tables here. We can always copy the code back in if we end up needing it, but I don't know if the continuous batching and kv caching logic that we will need will be identical to the gpu code or not

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -305,38 +305,103 @@ def prepare_model_input(
...
                                   input_masks=input_masks,
-                                  sampling_metadata=dummy_metadata,
+                                  sampling_metadata=sampling_metadata,
                                   is_prompt=is_prompt)
 
+    def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
+        """Update the cached states and the persistent batch with the scheduler
+        output.
+
+        The SamplingMetadata is updated and copied to the device if there is a
```
joerunde, line 321: nit: I don't think we're actually doing any "copying to device" operations here, right?

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -0,0 +1,467 @@
...
+                # Lazy allocation for this tensor, which can be large.
+                self.allowed_token_ids_mask = torch.zeros(self.max_num_reqs,
+                                                          self.vocab_size,
+                                                          dtype=torch.bool,
+                                                          device=self.device)
+            self.allowed_token_ids_mask[req_index][
+                sampling_params.allowed_token_ids] = True
+
+    def remove_request(self, req_id: str) -> Optional[int]:
+        """This method must always be followed by a call to condense()."""
```
joerunde, line 266: Trying to wrap my head around this- it looks like `remove_request` and `condense` are required if we want to update the sampling params after a request finishes, right? And the `condense` operation here is shuffling around these tensors so that they can be resized to be smaller to match the number of running requests.

If that's correct, then I think there's potentially a big problem here, where the ordering of requests in this condensed tensor will be different than the ordering of requests in the model runner (tracked by self._req_ids2idx) will be different. And I don't think we can actually manage memory on the AIU yet, so I don't think any reordering of requests is actually possible at all.

Looking into it a bit, diving through the code:

The batches of requests are padded out to the closest matching warmup shape's batch size (eg if we have 5 requests but the warmup shape's batch size is 8, then we pad out the batch with 3 fake empty requests), and then for padded or finished requests, we continually send in token id 0 while the batch finishes processing. The model then has this `indices` tensor which is a boolean, indicating if the request is actually running or not. So in this example that would look like `[T, T, T, T, T, F, F, F]` on the first pass, but if the second request finishes decoding before the others, then `model.indices` would be updated to look like `[T, F, T, T, T, F, F, F]`.  It looks like the model then only actually produces output for request indices that match `True` in that mask, so I think after the second request finishes, the outputs for requests 3-5 would be shifted down in the output tensor.

The fun news is that this is currently broken!
If you send in this request to create a batch of 2 prompts, where we force the first one to finish first (when it counts down to 0):
```
{
  "model": "/models/llama-194m/",
  "prompt": ["6 5 4", "This is a longe essay about courage: "],
  "max_tokens": 20,
  "stop": "0"
}
```
then we get an error because the scheduler tries to index past the end of the output tensor 🙃 


So I _think_ this all boils down to:
- We need to better understand what happens when one request finishes, and fix up the internal state and/or output appropriately so that we don't crash
- We need to make sure that ordering of these sampling metadata tensors is kept in sync with the internal state of the model

If I'm reading things right, the answer might be to shift all the request indices down when a request finishes, but I doubt it's that easy

wallashss:
Updates:

- Put it draft, since It still need to do couple of things
- Major refactoring considering #33 

TODO:
- Possibly minor changes
- Documentation
- New Tests from the scratch for spyre_input_batch
- Fix/update existing tests

tests/test_spyre_input_batch.py
```diff
@@ -0,0 +1,253 @@
...
+        block_ids=[],
+        generator=None,
+        num_computed_tokens=len(output_token_ids),
+        output_token_ids=output_token_ids,
+    )
+
+
+@pytest.mark.parametrize("device", ['cpu'])
+@pytest.mark.parametrize("batch_size", [1, 2, 32, 64])
+def test_sampling_metadata_in_input_batch(device: str, batch_size: int):
```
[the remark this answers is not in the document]
wallashss: Agree I'll do new tests from the scratch

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -0,0 +1,460 @@
...
+            top_p=None if self.no_top_p else self.top_p[indices_mask],
+            top_k=None if self.no_top_k else self.top_k[indices_mask],
+            min_p=None if self.no_min_p else self.min_p[indices_mask],
+            generators=generators,
+            max_num_logprobs=self.max_num_logprobs,
+            prompt_token_ids=prompt_token_ids,
+            frequency_penalties=self.frequency_penalties[indices_mask],
+            presence_penalties=self.presence_penalties[indices_mask],
+            repetition_penalties=self.repetition_penalties[indices_mask],
+            output_token_ids=cast(list[list[int]], self.req_output_token_ids),
```
joerunde, line 378: It's not clear to me whether this is updated while the batch is running to include all of the output tokens from each decode step

wallashss: Yeah... you're right. I accidentally removed the part of the code that do that in the last refactoring. I'll fix it soon. Thanks for the catch.

tests/test_spyre_input_batch.py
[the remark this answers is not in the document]
wallashss: Hey Joe. I'm sorry but I changed my mind on that! 😅 

While I like more readable code as well, after actually run these tests on the new input_batch I found those tests very useful. I found a lot of flaws and they helped to fix my implementation. I could totally write new fresh test, but I guess I would fall on more readable and less complete or complete but very similar to what we already have. 

I'm confident that the implementation is good now and I believe that has value we have these tests in this repo to keep sync with the contributions of vLLM community.

vllm_spyre/v1/worker/spyre_input_batch.py
wallashss: Solved

vllm_spyre/platform.py
```diff
@@ -81,7 +81,7 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         max_seq_len = 0
         for shape in cls.get_warmup_shapes():
             max_batch_size = max(max_batch_size, shape['batch_size'])
-            max_seq_len = max(max_batch_size,
+            max_seq_len = max(max_seq_len,
```
wallashss, line 84: FYI: Found this bug that was breaking the tests with my changes.

.github/workflows/test-spyre.yml
```diff
@@ -25,5 +25,5 @@ jobs:
           export DISTRIBUTED_STRATEGY_IGNORE_MODULES=WordEmbedding && \
           cd vllm-spyre && \
           python -m pytest --timeout=300  tests -v -k "V0 and eager" && \
-          python -m pytest --forked --timeout=300  tests -v -k "V1- and eager"
+          python -m pytest --forked --timeout=300  tests -v -k "(V1- and eager) or test_sampling_metadata_in_input_batch"
```
joerunde, line 28: 🤔 

hmmm, maybe we need to do something else here. So far all of the existing tests are "end to end" tests (they use vllm.entrypoints.*) and require parameterizing on both hardware and v0/v1 support. This new test isn't an end to end test, so it doesn't require that parameterization. Also, the v0 and v1 tests are currently split up because the v1 tests require `--forked` which runs each test in a new subprocess, while the v0 tests can run sequentially in a single worker. Since this test only tests `InputBatch` and doesn't load a model, I think it should be able to run without `--forked`.

Maybe we can repackage things around a bit here to make it clear which tests are doing what? 
Something like `tests/spyre/...` or `tests/e2e/...` for the existing tests? `tests/worker/...` for the new tests?

joerunde (approved):
I just spun this code up on a dev pod as well and tried out all the sampling params, and things seem to work okay (though that's definitely not a full correctness test).

The test setup is a bit wonky, but I don't think that has to be a merge blocker. I really like how this is integrated now with `InputBatch` handling all of the internal state for us, I'd like to get this in so we can start iterating on it for the upcoming continuous batching work

tests/test_spyre_input_batch.py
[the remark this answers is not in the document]
joerunde: Okay, good to hear they caught bugs!

PR #26, opened 2025-03-13
Use execute_model for warmup

### LINKED ISSUE #12: [V1] Use the "real" model forward pass for warmup
Currently we implement a separate dummy forward pass for the model warmup in `v1.worker.SpyreWorker.compile_or_warm_up_model`

If we can instead figure out how to use `execute_model` during the warmup, we can both reduce the code required here and ensure that the torch compilation that runs at warmup is building the correct graphs. 🤞 this is a simple refactor, but I have not looked into it to know about any potential blockers.

rafvasq:
PR to use `execute_model` in the model warmup `v1.worker.SpyreWorker.compile_or_warm_up_model` instead of a separate dummy forward pass. 

Closes #12


joerunde:
Can `_raw_model_forward` on the model runner be removed too?

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -68,8 +69,51 @@ def compile_or_warm_up_model(self) -> None:
...
-                                          self.restricted_tokens, batch_size)
+
+            num_scheduled_tokens: dict = {}
+            total_num_scheduled_tokens: int = 0
+            dummy_requests: list = []
+            for i in range(batch_size):
+                dummy_requests.append(
+                    NewRequestData(
+                        req_id=f"warmup-{i}",
+                        prompt_token_ids=[1] * prompt_len,
```
yannicks1, line 80: We previously sampled these tokens from a list of valid tokens: 
```
valid_token_ids = [ i for i in range(1, vocab_size) if i not in set(special_token_ids)]
```
where `special_token_ids` contains BOS, EOS and pad token ids. Not sure whether this was needed, or what happens if in your case any of the above special token ids are 1.

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -68,8 +69,51 @@ def compile_or_warm_up_model(self) -> None:
...
+                total_num_scheduled_tokens=total_num_scheduled_tokens,
+                scheduled_spec_decode_tokens={},
+                scheduled_encoder_inputs={},
+                num_common_prefix_blocks=0,
+                finished_req_ids=set(),
+                free_encoder_input_ids=[],
+            )
+
+            # Use execute_model for warm up
+            self.execute_model(scheduler_output)
```
yannicks1, line 107: As far as I understand this executes the model only for one step. In your case it does the prefill and generates the 1st token. We need to warm up not only for prefill, but also for `(num_decode_tokens - 1)` decode steps (since prefill already produced 1 token).

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -68,8 +69,51 @@ def compile_or_warm_up_model(self) -> None:
...
+
+            if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == "sendnn_decoder":
+                from torch_sendnn import torch_sendnn
+                ul_start_time = time.time()
+                torch_sendnn.update_lazyhandle()
+                ul_stop_time = time.time()
+                ul_total_t = ul_stop_time - ul_start_time
+                print(f"update_lazyhandle() done (duration: {ul_total_t}s)")
+
```
yannicks1, line 116: After `torch_sendnn.update_lazyhandle()` there is a second _**complete**_ warmup needed. 

To sum up: 

1. complete forward pass: prefill plus `(num_decode_tokens - 1)` decode steps
2. torch_sendnn.update_lazyhandle()
3. complete forward pass: prefill plus `(num_decode_tokens - 1)` decode steps

See comment below.

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -209,118 +253,6 @@ def load_model(self):
...
-            torch_sendnn.update_lazyhandle()
-            ul_stop_time = time.time()
-            ul_total_t = ul_stop_time - ul_start_time
-            print(f"update_lazyhandle() done (duration: {ul_total_t}s)")
-
-        # warmup 2nd forward pass
-        self._warmup_model_forward_pass(warmup_tokens_tensor,
-                                        valid_token_ids_tensor, prompt_len,
-                                        num_decode_tokens, batch_size,
-                                        extra_kwargs)
```
yannicks1, line 273: this has to happen for each warmup shape (combination of `prompt_len`, `num_decode_tokens`, `batch_size`)

yannicks1 (changes requested):

wallashss:
Unfortunately, I could not yet tried your changes because of problems in my dev environment. But, besides the feedback from other reviewers, I feel that the final code should have more comments, specially the review comments pointed out by @yannicks1 for better understanding later.

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -68,8 +69,51 @@ def compile_or_warm_up_model(self) -> None:
             print(f"[SpyreWorker] Warming up for prompt length {prompt_len}, "
                   f"decoding {num_decode_tokens} tokens with batch "
                   f"size {batch_size}")
-            self._warmup_spyre_fixed_size(prompt_len, num_decode_tokens,
```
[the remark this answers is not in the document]
joerunde: +1, specifically avoiding double-nested for loops is nice to do

vllm_spyre/v1/worker/spyre_worker.py
joerunde: Hah, for the very first dummy model I checked this is true: https://huggingface.co/JackFram/llama-160m/blob/main/config.json#L6

I think vllm uses a bunch of repeated token id 0 for profiling, since the input ids tensor is just initialized with `torch.zeros` and for text-only models it's not update for profiling. 

The general idea with setting a repeated token ID was to have the model continue the sequence, so it doesn't end up hitting an eos token early and stopping. But if we control the loop here that keeps invoking the model, maybe that doesn't matter.

joerunde:
@rafvasq Can you also update all the prints in this code to logging statements?

See https://github.com/vllm-project/vllm-spyre/pull/31 for context. I didn't update the prints in this warmup code because I knew you were rewriting it

vllm_spyre/v1/worker/spyre_worker.py
rafvasq: Thanks for clarifying this, learning as I go but it makes sense to me now.

I took another stab at it, still trying to use `execute_model` to avoid doing anything manually except the dummy data setup.

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -253,95 +250,99 @@ def _warmup_spyre_fixed_size(self, prompt_len, num_decode_tokens,
...
+
+        # update_lazyhandle
         if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == "sendnn_decoder":
             from torch_sendnn import torch_sendnn
             ul_start_time = time.time()
             torch_sendnn.update_lazyhandle()
             ul_stop_time = time.time()
-            ul_total_t = ul_stop_time - ul_start_time
-            print(f"update_lazyhandle() done (duration: {ul_total_t}s)")
+            logger.info("update_lazyhandle() done (duration: %.3fs",
```
joerunde, line 323:
```suggestion
            logger.info("update_lazyhandle() done (duration: %.3fs)",
```

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -253,95 +250,99 @@ def _warmup_spyre_fixed_size(self, prompt_len, num_decode_tokens,
...
-        # warmup 2nd forward pass
-        self._warmup_model_forward_pass(warmup_tokens_tensor,
-                                        valid_token_ids_tensor, prompt_len,
-                                        num_decode_tokens, batch_size,
-                                        extra_kwargs)
+        # Switch to cached requests to trigger decoding steps
+        scheduler_output.scheduled_new_reqs = []
+        scheduler_output.scheduled_cached_reqs = cached_requests
+
+        logger.info("[Warmup 2/2: Decoding steps...")
```
joerunde, line 336:
```suggestion
        logger.info("Warmup 2/2: Decoding steps...")
```

vllm_spyre/v1/worker/spyre_worker.py
rafvasq: Thanks for catching these!

joerunde (approved):
LPGTM!

I'll let @yannicks1 take another look through since he has a better understanding of the requirements

vllm_spyre/v1/worker/spyre_worker.py
yannicks1: yeah, I also suspect it does not matter as we force the decode steps in the loop, but rather save than sorry:)

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -253,95 +250,99 @@ def _warmup_spyre_fixed_size(self, prompt_len, num_decode_tokens,
...
+        self.execute_model(scheduler_output)  # Prefill step
+
+        # Switch to cached requests to trigger decoding steps
+        scheduler_output.scheduled_new_reqs = []
+        scheduler_output.scheduled_cached_reqs = cached_requests
+
+        logger.info("Warmup 1/2: Decoding...")
+        for _ in range(num_decode_tokens - 1):
+            self.execute_model(scheduler_output)
+
```
yannicks1, line 316: Personally, I am for the use of helper functions wherever they help reducing code duplication. I am aware its just 10 lines here, but they could be eliminated by reusing the `_warmup_model_forward_pass` we introduced in the original implementation.

yannicks1 (commented):
Other than the two minor comments this looks good to me! Thanks for contributing!

PS: I am mildly confused why the spyre-tests did not succeed. 
`tests/test_spyre_embeddings.py` fails with V0. But as your code changes only touch V1, the reason has got to be somewhere else...

vllm_spyre/v1/worker/spyre_worker.py
rafvasq: Sounds good, I re-introduced `_warmup_model_forward_pass` to handled the duplicate pass code

yannicks1 (approved):
LGTM! Thanks for contributing.

PR #28, opened 2025-03-14
[V1] Cleanup kv-caching apis and disable paged attention

### LINKED ISSUE #19: [V1] Disable paged attention when scheduling
The spyre plugin currently does not support paged attention and kv-caching in the way that upstream vllm does. The base scheduler that we extend though will attempt to do things like preempt a request if there are not enough available kv cache blocks to allocate for it.

For the v0 implementation, the worker API included `determine_num_available_blocks` and `initialize_cache`, and both the available number of blocks and the total number of the blocks in cache were pinned to `max_num_seqs`. The comments assert that this is equivalent to disabling paged attention:

```python
def determine_num_available_blocks(self) -> Tuple[int, int]:
        """Determine the number of available KV blocks.

        Swapping is not yet supported, so always return num_cpu_blocks=0.

        We configure num_gpu_blocks to be equal to max_num_seqs.
        """
        # Set the number of GPU blocks to be the same as the maximum number of
        # sequences that can be processed in a single batch. This is equivalent
        # to schedule without PagedAttention.
        num_gpu_blocks = self.scheduler_config.max_num_seqs

        # Swap not yet supported with Spyre backend.
        num_cpu_blocks = 0

        return num_gpu_blocks, num_cpu_blocks

    def initialize_cache(self, num_gpu_blocks: int,
                         num_cpu_blocks: int) -> None:
        """Initialize the KV cache.
        """

        # Different values are not tested.
        assert num_cpu_blocks == 0
        assert num_gpu_blocks == self.scheduler_config.max_num_seqs

        self.cache_config.num_gpu_blocks = num_gpu_blocks
        self.cache_config.num_cpu_blocks = num_cpu_blocks
```

This interface has changed for v1, and now workers are required to implement `get_kv_cache_spec` and `determine_available_memory` as well as `determine_num_available_blocks`. These currently have dummy outputs which have only been verified to work for single small requests. I have not looked very deep into how these apis are used by vLLM at setup or during scheduling, so we should investigate to see what else may need to be done to ensure that the scheduler doesn't attempt to do any paged attention operations.

As an alternative, we could fully write our own scheduler with much more basic functionality, but I understand our long term goal is to be able to fully support the stock scheduler.

joerunde:
This PR
- Cleans up unused kv-cache related APIs from the v1 worker
- Adds detailed docstrings about how the new kv-cache related v1 apis work
- Overrides the necessary values to copy v0's behavior of 1 block per sequence to avoid any preemption, prefix caching, or other kv cache block related logic in the scheduler
- Does a small refactor to use `super().__init__` in the model runner

Closes #19

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -85,7 +85,7 @@ class SpyrePlatform(Platform):
                               shape['prompt_length'] + shape['new_tokens'])
 
         if envs.VLLM_USE_V1:
-            # The v0 scheduler will run out of blocks if this is overriden
+            # The v0 scheduler will run out of blocks if this is overridden
             scheduler_config.max_num_seqs = max_batch_size
 
         cache_config = vllm_config.cache_config
@@ -104,7 +104,7 @@ class SpyrePlatform(Platform):
             cache_config.block_size = model_config.max_model_len
             cache_config.num_gpu_blocks_override = scheduler_config.max_num_seqs
             logger.info(
-                "Overriding configurations based on warmup shaped. "
+                "Overriding configurations based on warmup shapes. "
                 "max_model_len=%d, max_num_seqs=%d, block_size=%d, "
                 "num_gpu_blocks_override=%d", model_config.max_model_len,
                 scheduler_config.max_num_seqs, cache_config.block_size,
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -39,6 +39,7 @@ class SpyrePlatform(Platform):
     def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         parallel_config = vllm_config.parallel_config
         scheduler_config = vllm_config.scheduler_config
+        model_config = vllm_config.model_config
 
         if scheduler_config.is_multi_step:
             raise NotImplementedError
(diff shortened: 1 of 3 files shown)


vllm_spyre/platform.py
```diff
@@ -73,12 +74,32 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+
+            # To disable any paged attention ops in the base scheduler, we both:
+            # - Set the block size (in tokens) to the maximum sequence length
+            #       so that the scheduler thinks an entire sequence will fit in
+            #       one single block.
+            # - Set the number of blocks to the maximum number of sequences, so
+            #       the scheduler always thinks there's a block available
+            model_config.max_model_len = max_seq_len
+            cache_config.block_size = model_config.max_model_len
+            cache_config.num_gpu_blocks_override = scheduler_config.max_num_seqs
```
wallashss, line 102: Probably it is redundant in most scenario where users knows how spyre works, but what do you think log to the user that we are override the system configuration with those values?

wallashss:
LGTM, and I really like the way you document the decisions, very good to read the code later.

vllm_spyre/platform.py
joerunde: 👍 
When I added this log message I realized we needed https://github.com/vllm-project/vllm-spyre/pull/31 as well

vllm_spyre/platform.py
```diff
@@ -73,12 +74,41 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+        if cache_config and model_config:
+            # Cache and model config aren't set in the individual worker procs
+            # These are set in the main engine process
+
+            # To disable any paged attention ops in the base scheduler, we both:
+            # - Set the block size (in tokens) to the maximum sequence length
+            #       so that the scheduler thinks an entire sequence will fit in
+            #       one single block.
+            # - Set the number of blocks to the maximum number of sequences, so
+            #       the scheduler always thinks there's a block available
```
tdoublep, line 102: Where does this logic with `num_gpu_blocks_override` happen?

vllm_spyre/platform.py
```diff
@@ -73,12 +74,41 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+        max_batch_size = 0
+        max_seq_len = 0
+        for shape in cls.get_warmup_shapes():
+            max_batch_size = max(max_batch_size, shape['batch_size'])
+            max_seq_len = max(max_batch_size,
+                              shape['prompt_length'] + shape['new_tokens'])
+
+        if envs.VLLM_USE_V1:
+            # The v0 scheduler will run out of blocks if this is overridden
+            scheduler_config.max_num_seqs = max_batch_size
```
tdoublep, line 89: Why will it run out of blocks? I guess it's not that important.

tdoublep (approved):
LGTM - have some minor questions, but nothing that blocks us merging

vllm_spyre/platform.py
joerunde: Yeah... I didn't debug too far on that one because it's v0, and I assume our plan is still to focus development only on v1

PR #31, opened 2025-03-17
:recycle: Use real logger

joerunde:
I noticed that we weren't seeing some of the `logger.warning` messages that I had put into the v1 code, because we never initialized the logging package correctly for the plugin.

This PR reconfigures the logging package on import, and copies the vllm logging config for `vllm_spyre` so that all the usual logging configurations still work. For example, `VLLM_LOGGING_LEVEL=WARNING` will turn the level down to warning for the vllm_spyre logs too. This allows users to configure logging the same way as they do with other hardware backends.

I went ahead and ported some of the prints in the v1 codepath to use logging as well. This now logs the file and line number where the logs came from, so we no longer need to maintain manual prefixes like `[SchedulerConfig]`  to denote where the log is from. An example snippet:

```
INFO 03-17 18:51:10 [spyre_worker.py:200] load model...
WARNING 03-17 18:51:10 [config.py:3600] Current VLLM config is not set.
INFO 03-17 18:51:10 [platform.py:114] VLLM_SPYRE_WARMUP_PROMPT_LENS = [64]
INFO 03-17 18:51:10 [platform.py:115] VLLM_SPYRE_WARMUP_NEW_TOKENS = [20]
INFO 03-17 18:51:10 [platform.py:116] VLLM_SPYRE_WARMUP_BATCH_SIZES = [1]
WARNING 03-17 18:51:10 [topk_topp_sampler.py:46] FlashInfer is not available. Falling back to the PyTorch-native implementation of top-p & top-k sampling. For the best performance, please install FlashInfer.
ignoring module=WordEmbedding when distributing module
INFO 03-17 18:51:10 [spyre.py:200] NOTICE: Adjusting torch._dynamo.config.cache_size_limit from 8 to 256 to accommodate prompt size of 64 and decode tokens of 20
INFO 03-17 18:51:10 [spyre_worker.py:215] load model took 0.237s
INFO 03-17 18:51:10 [kv_cache_utils.py:537] GPU KV cache size: 1,073,741,824 tokens
INFO 03-17 18:51:10 [kv_cache_utils.py:540] Maximum concurrency for 4,096 tokens per request: 262144.00x
INFO 03-17 18:51:10 [spyre_worker.py:53] Start warming up 1 different prompt/decode/batchsize-shape combinations.
INFO 03-17 18:51:10 [spyre_worker.py:68] Warmup 1/1 prompt/decode/batchsize-shape combinations...
```

### DIFF
--- a/vllm_spyre/__init__.py
+++ b/vllm_spyre/__init__.py
@@ -1,3 +1,30 @@
+from logging.config import dictConfig
+
+from vllm.logger import DEFAULT_LOGGING_CONFIG
+
+
 def register():
     """Register the Spyre platform."""
     return "vllm_spyre.platform.SpyrePlatform"
+
+
+def _init_logging():
+    """Setup logging, extending from the vLLM logging config"""
+    config = {**DEFAULT_LOGGING_CONFIG}
+
+    # Copy the vLLM logging configurations for our package
+    config["formatters"]["vllm_spyre"] = DEFAULT_LOGGING_CONFIG["formatters"][
+        "vllm"]
+
+    handler_config = DEFAULT_LOGGING_CONFIG["handlers"]["vllm"]
+    handler_config["formatter"] = "vllm_spyre"
+    config["handlers"]["vllm_spyre"] = handler_config
+
+    logger_config = DEFAULT_LOGGING_CONFIG["loggers"]["vllm"]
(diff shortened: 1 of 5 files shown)


tdoublep (approved):
LGTM - thanks!

PR #32, opened 2025-03-17
:construction_worker: Only fork v1 tests

joerunde:
The v0 tests shouldn't need forking, since they clean up properly after themselves. This PR splits the pytest command into two separate runs, using `--forked` only for the v1 tests. 

This recent run took 12 minutes: https://github.com/vllm-project/vllm-spyre/actions/runs/13910223043/job/38922610694?pr=28
Let's see if this is faster without forking half the tests 🤞


joerunde:
This run took 10m28s, compared to 11m49s.

🤷  Not a huge improvement, but not bad either

tdoublep (approved):
LGTM

PR #37, opened 2025-03-19
✨  Reject requests, and upgrade vllm install to 0.8.0

### LINKED ISSUE #16: [V1] Reject requests that don't fit a warmup shape
See current V1 workarounds here: https://github.com/vllm-project/vllm/pull/14242

Requests must have both a prompt length and a requested number of tokens that is less than or equal to those same settings on a single warmup shape. If a request matches no warmup shape in this way, it must be rejected.

In the V0 implementation, this constraint is checked by the scheduler and the scheduler marks the request as ignored if it matches no warmup shapes. In V1, this currently does not work because the engine logic does not have any logic to handle requests that are immediately rejected.

We could implement this logic in the engine, or we could explore extending the platform api to allow it to validate requests as they are added, instead of at schedule-time. That alternate approach may allow rejecting requests with a 400-type error instead of returning empty results

joerunde:
This PR implements plugin-side rejection of requests that don't fit a warmup shape

Closes #16 

The reason I did it this way instead of opening an upstream PR for vllm to allow the scheduler to reject requests is:
- There's an open PR (https://github.com/vllm-project/vllm/pull/11737 and https://github.com/vllm-project/vllm/pull/14048) that should allow us to actually raise an error from `add_request` so that the user gets a real 400-type error, which IMHO would be much better than a 200 with empty text
- We're accumulating a lot of drift with the v1 implementation, and I'd prefer us getting a fix in the plugin now rather than waiting on upstream reviews for a feature that the stock v1 scheduler doesn't support, so that we can resume testing against vllm main (or at least the latest release). There's already a blocker where the plugin won't boot with 0.8.1 that we need to address :(

This PR also updates the vllm install to 0.8.0, which is the latest working release for the plugin

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -1,14 +1,15 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections import deque
-from typing import Deque, Optional
+from typing import Deque
 
-from vllm.config import (CacheConfig, LoRAConfig, ModelConfig, SchedulerConfig,
-                         SpeculativeConfig)
 from vllm.logger import init_logger
 from vllm.platforms import current_platform
+from vllm.sampling_params import SamplingParams
 from vllm.v1.core.scheduler import Scheduler
 from vllm.v1.core.scheduler_output import SchedulerOutput
+from vllm.v1.engine import EngineCoreOutput, EngineCoreOutputs, FinishReason
+from vllm.v1.outputs import ModelRunnerOutput
 from vllm.v1.request import Request, RequestStatus
 
 logger = init_logger(__name__)
@@ -20,22 +21,9 @@ class SpyreScheduler(Scheduler):
     - Only schedules batches of requests that fit a common warmup shape
     """
 
-    def __init__(
-        self,
-        scheduler_config: SchedulerConfig,
-        model_config: ModelConfig,
-        cache_config: CacheConfig,
-        lora_config: Optional[LoRAConfig],
-        speculative_config: Optional[SpeculativeConfig],
-        log_stats: bool,
-    ) -> None:
+    def __init__(self, *args, **kwargs) -> None:
         # Initialize vLLM scheduler
-        super().__init__(scheduler_config=scheduler_config,
-                         model_config=model_config,
-                         cache_config=cache_config,
-                         lora_config=lora_config,
-                         speculative_config=speculative_config,
-                         log_stats=log_stats)
+        super().__init__(*args, **kwargs)
 
         # Add our own state for handling Spyre constraints
 
@@ -48,6 +36,48 @@ class SpyreScheduler(Scheduler):
         # waiting queue. This lets us ensure that the set of requests the base
         # scheduler sees have at least one common warmup shape.
         self.holdback_queue: Deque[Request] = deque()
+
+        self.rejected_requests_this_iteration: set[str] = set()
+
+    def add_request(self, request: Request) -> None:
+        """This override rejects requests that fit no warmup shape"""
+        if len(
+                self._get_matching_warmup_shapes(
+                    request, list(self.spyre_warmup_shapes))) == 0:
+            logger.warning(
+                "No applicable warmup shape exists for "
+                "combination of prompt length (%d tokens) "
+                "and maximum number of output tokens to be "
+                "generated (%d tokens)", request.num_prompt_tokens,
+                request.sampling_params.max_tokens)
+            # TODO: There are open PRs that should enable raising an error for
+            # a single request like this, which will gracefully return an error
+            # for the request, instead of shutting down the engine.
+            # See https://github.com/vllm-project/vllm/pull/11737
+            # raise ValueError("Request does not fit any spyre warmup shape")
+
+            # For now, we'll insert a dummy request and manually reject it when
+            # we construct the outputs later
+            self.rejected_requests_this_iteration.add(request.request_id)
+            request.prompt_token_ids = [0]
+            request.num_prompt_tokens = 1
+            request.sampling_params = SamplingParams(max_tokens=1)
+
+        # delegate to super
+        super().add_request(request=request)
+
+    def update_from_output(
+        self,
+        scheduler_output: SchedulerOutput,
+        model_runner_output: ModelRunnerOutput,
+    ) -> EngineCoreOutputs:
+        """Temporary override to handle rejected requests that were too large
+        to schedule."""
+        reject_outputs = self._handle_rejects()
+        outputs = super().update_from_output(scheduler_output,
+                                             model_runner_output)
+        outputs.outputs.extend(reject_outputs)
+        return outputs
 
     def schedule(self) -> "SchedulerOutput":
         """This override adds constraints and then delegates most of the work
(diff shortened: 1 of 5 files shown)


joerunde:
Dang, simple manual tests are working but the warmup shapes test that flexes this is failing, so I think there's still a problem :(

Will keep digging...

joerunde:
This should be working and good to review now 👍

vllm_spyre/v1/core/scheduler.py
```diff
@@ -119,3 +131,59 @@ def schedule(self) -> "SchedulerOutput":
...
+            shape for shape in warmup_shapes
+            if request.num_prompt_tokens <= shape['prompt_length']
+            and max_tokens <= shape['new_tokens']
+            and current_batch_size < shape['batch_size']
+        ]
+
+    def _handle_rejects(self) -> list[EngineCoreOutput]:
+        """Temporary solution to reject requests that were too large to
+        schedule. This removes the rejected requests from the scheduler, and 
+        returns empty outputs for them with finish reason `abort`.
```
tdoublep, line 154: The behaviour in V0 was that if the request didn't match any warmup shape, it would be added to the ignore queue. This results in finish reason 'length' and empty string as output. While we didn't really like this behaviour, it was consistent with how vLLM handled similar cases (e.g., prompt length being longer than max model len) in V0. Has this changed in V1?

tdoublep (commented):
Looks generally good, just have one comment about *how* the requests get rejected since this seems to be change vs. what we were doing in V0.

vllm_spyre/v1/core/scheduler.py
joerunde: @tdoublep Yeah, as far as I can tell the v1 scheduler doesn't have the ability to ignore requests. The new flow in the v1 engine is that the scheduling is broken down into two passes, `schedule()` and `update_from_output()`. The engine expects all requests to schedule at least one token eventually from `schedule()`, and if no tokens are scheduled for an iteration it'll just go idle waiting for more inputs. Once the model has been invoked, `update_from_output()` is used to construct the outputs for the engine's callers. If we look at the simple case where we get only one single request and it doesn't fit a warmup shape, we can't just not schedule it or the engine will go idle and we never get a chance to pass a result back to the caller. So we have to schedule at least one dummy token, and then pass back an empty result from `update_from_output()`, which is a bit sad.

As far as I can tell, all of the request validation has been moved ahead of the scheduler, so that for cases like `len(prompt) > max_model_len`, the user gets an error and the scheduler is never invoked for that request. For online requests, that's handled in the api server for both V0 and V1:
```
{
  "object": "error",
  "message": "This model's maximum context length is 100 tokens. However, you requested 153 tokens (137 in the messages, 16 in the completion). Please reduce the length of the messages or completion.",
  "type": "BadRequestError",
  "param": null,
  "code": 400
}
```

But the offline entrypoint doesn't do as much request validation, with V1 the engine handles this during "input processing" and will now raise a `ValueError` on any invalid inputs:
```
model = LLM("/models/llama-194m", max_model_len=100)
model.generate("hello 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 hello 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 hello 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 hello 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/utils.py", line 1080, in inner
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/entrypoints/llm.py", line 462, in generate
    self._validate_and_add_requests(
  File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/entrypoints/llm.py", line 1310, in _validate_and_add_requests
    self._add_request(
  File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/entrypoints/llm.py", line 1328, in _add_request
    self.llm_engine.add_request(
  File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/engine/llm_engine.py", line 164, in add_request
    request = self.processor.process_inputs(request_id, prompt, params,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/engine/processor.py", line 141, in process_inputs
    self._validate_model_inputs(processed_inputs)
  File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/engine/processor.py", line 266, in _validate_model_inputs
    raise ValueError(
ValueError: Prompt length of 137 is longer than the maximum model length of 100.
```

IIUC, the input processing step was removed from the main engine loop in V1 so that it could better handle these cases. Previously with V0 any errors would kill the engine completely, now V1 can raise any validation errors it wants before the request gets to the main engine loop. But I do see with Robert's incoming changes [here](https://github.com/vllm-project/vllm/pull/14048/files) that there should soon be per-request failure handling from inside the engine loop that may allow us to raise a value error from our scheduler, which would be a lot better than this hack.

However, I think the best approach would be to let the plugins hook into the v1 engines' input processing step so that we can properly enforce the new assumption that every request that makes it to the scheduler is schedulable. I can start talking to maintainers about that, but I did still want to get this temporary fix in so we can keep moving forward.

joerunde:
I just removed the backwards compatibility for 0.7.3, since remembering that 0.8.0+ is required anyway for the pluggable scheduler

vllm_spyre/v1/core/scheduler.py
```diff
@@ -119,3 +131,59 @@ def schedule(self) -> "SchedulerOutput":
...
+
+        for request in rejected_requests:
+            queue.remove(request)
+            reject_outputs.append(
+                EngineCoreOutput(request.request_id,
+                                 new_token_ids=[],
+                                 finish_reason=FinishReason.ABORT,
+                                 stop_reason="Request did not fit any warmup "
+                                 "shape"))
+            request.status = RequestStatus.FINISHED_ABORTED
```
yannicks1, line 185: @joerunde Would your solution still work when you put: 
```
            reject_outputs.append(
                EngineCoreOutput(request.request_id,
                                 new_token_ids=[],
                                 finish_reason=FinishReason.LENGTH,
                                 stop_reason="Request did not fit any warmup "
                                 "shape"))
            request.status = RequestStatus.FINISHED_IGNORED
```
here? As @tdoublep mentioned, this would be more consistent with how we used to handle it previously. 
A question regarding your [comment](https://github.com/vllm-project/vllm-spyre/pull/37/files#r2007979062): Does this mean that for both online and offline inference the `RequestStatus` and `FinishReason` are never set but an error is raised before? Are therefore the `FinishReason.LENGTH` and `RequestStatus.FINISHED_IGNORED` obsolete [here](https://github.com/vllm-project/vllm/blob/main/vllm/v1/request.py)? If that is the case, I don't have any preference...

yannicks1 (commented):
LGTM in general, dropped a small comment on the choice of `FinishReason` and `RequestStatus` when rejecting request.

vllm_spyre/v1/core/scheduler.py
yannicks1: Sounds good. Also no strong opinion here, just asked for better understanding.

joerunde:
Slack thread about this here: https://vllm-dev.slack.com/archives/C087RA55P0D/p1742828751236509

joerunde:
I'll go ahead and merge this in so we can upgrade, will keep pulling on that slack thread and figure out how to iterate on this behavior

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -323,7 +323,7 @@ def prepare_model_input(
             min_tokens={},
             logit_bias=[None for _ in range(num_reqs)],
             allowed_token_ids_mask=None,
-        )
+            bad_words_token_ids={})
```
wallashss, line 326: Have we update vllm?

PR #39, opened 2025-03-20
FMS Wrapper for static batching

yannicks1:
### FMS Wrapper for static batching

changes: 
- introducing env var `VLLM_SPYRE_USE_CB` to switch between continuous batching and static batching
- introducing `FmsModelPseudoWrapper` which mimics the static batching kv cache handling alongside the existing `FmsModelWrapper` for continuous batching kv cache handling
- both wrapper classes inherit from a base class `FmsModelBaseWrapper` to avoid code duplication.


nikolaospapandreou (approved):
LGTM

PR #44, opened 2025-03-24
[GPTQ] fix imports of moved add ons 

### LINKED ISSUE #42: [GPTQ] fails to start with upstream aiu-fms repository
aiu-fms has been made public, and the upstream seems to have split up into two repos:

https://github.com/foundation-model-stack/aiu-fms-testing-utils

and

https://github.com/foundation-model-stack/fms-model-optimizer

When starting with a vllm with a gptq model it does not find `aiu_as_addon`

yannicks1:
### [GPTQ] fails to start with upstream aiu-fms repository

closes #42 

changes:
- fix import in model loader ([here](https://github.com/vllm-project/vllm-spyre/commit/5773b735a433a2477ec9dd85c001e60b57490b31))

Tested successfully on AIU Spyre with granite 8b.

### DIFF
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -133,6 +133,7 @@ class SpyreCausalLM(nn.Module):
                 logger.info("Loaded `aiu_addons` functionalities")
             else:
                 linear_type = "gptq_cpu"
+                logger.warning("GPTQ is not expected to work on CPU.")
 
             quant_cfg = model_config._parse_quant_hf_config()
 
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -121,11 +121,6 @@ class SpyreCausalLM(nn.Module):
                 model_config.dtype, self.dtype)
 
         if model_config.quantization == "gptq":
-
-            # note, we have to find a better way to package this
-            # shouldn't it be part of FMS?
-            sys.path.append("/home/senuser/aiu-fms")
-
             if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == "sendnn_decoder":
                 from fms_mo.aiu_addons.gptq import (  # noqa: F401
                     gptq_aiu_adapter, gptq_aiu_linear)
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -127,7 +127,8 @@ class SpyreCausalLM(nn.Module):
             sys.path.append("/home/senuser/aiu-fms")
 
             if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == "sendnn_decoder":
-                from fms_mo.aiu_addons.gptq import gptq_aiu_adapter, gptq_aiu_linear  # noqa: F401
+                from fms_mo.aiu_addons.gptq import (  # noqa: F401
+                    gptq_aiu_adapter, gptq_aiu_linear)
                 linear_type = "gptq_aiu"
                 logger.info("Loaded `aiu_addons` functionalities")
             else:
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -1,6 +1,5 @@
 """Utilities for selecting and loading Spyre models."""
 import os
-import sys
 from typing import Optional
 
 import torch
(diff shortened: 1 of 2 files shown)


vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -127,13 +127,12 @@ def load_weights(self, model_config: ModelConfig, max_prompt_length: int,
...
             if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == "sendnn_decoder":
-                from aiu_as_addon import aiu_adapter, aiu_linear  # noqa: F401
+                from fms_mo.aiu_addons.gptq import (  # noqa: F401
+                    gptq_aiu_adapter, gptq_aiu_linear)
                 linear_type = "gptq_aiu"
-                logger.info("Loaded `aiu_as_addon` functionalities")
+                logger.info("Loaded `aiu_addons` functionalities")
             else:
-                from cpu_addon import cpu_linear  # noqa: F401
                 linear_type = "gptq_cpu"
```
tdoublep, line 135: Should we perhaps just raise an exception here (since we don't expect it to work)?

yannicks1: I inserted a logger warning. This way it does not directly raise an exception if this should work in the future due to some changes in the fms repos.

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -127,13 +127,13 @@ def load_weights(self, model_config: ModelConfig, max_prompt_length: int,
             sys.path.append("/home/senuser/aiu-fms")
 
             if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == "sendnn_decoder":
-                from aiu_as_addon import aiu_adapter, aiu_linear  # noqa: F401
+                from fms_mo.aiu_addons.gptq import (  # noqa: F401
```
joerunde, line 130: Is the above `sys.path.append("/home/senuser/aiu-fms")` still required now that the `fms_mo` package exists? It's not clear from the comment what we're actually importing from that path

joerunde: Also, we should ideally be tracking `fms_mo` as a dependency, but I'm assuming that it's not published to pypi yet, right?

dpatel-ops: I have not tried gptq but yes I agree we may not need `aiu-fms` anymore as everything there should be either in `fms_mo` or in `fms`.

vllm_spyre/model_executor/model_loader/spyre.py
tdoublep: If it doesn't work now, I would still prefer to raise an exception rather than warning if it is unsupported configuration.  Exception is much better feedback to the user.

vllm_spyre/model_executor/model_loader/spyre.py
tdoublep: Agree we need to add `fms_mo` as a dep.

joerunde:
@yannicks1 it looks like fms-mo only supports python <=3.11, but our build uses python 3.12

We could downgrade our build to 3.11, or work with the fms-mo team to build for 3.12 as well. Usually that's a simple build config setting, but I don't know enough to know if fms-mo specifically has problems supporting 3.12.

I would prefer to get a release of fms-mo supporting 3.12, but if we need to get this change in with some README updates that mention gptq support is only on py311 that would be okay as well.

joerunde:
Shoot, it looks like with torch 2.5.1 the install of `triton` might brick model loading with vllm 0.8.2 :(

tjohnson31415:
> Shoot, it looks like with torch 2.5.1 the install of triton might brick model loading with vllm 0.8.2 :(

I reset my dev environment and installed vllm 0.8.2 with fms_mo (hacked to install with Python 3.12) and it doesn't raise the `TypeError: must be called with a dataclass type or instance` error. `fms_mo` pins `triton < 3.2` and it is actually `triton==3.2` with `torch==2.5.1` that seem to be incompatible.

> I don't know enough to know if fms-mo specifically has problems supporting 3.12.

I did see a [note in the README](https://github.com/foundation-model-stack/fms-model-optimizer/blob/5902c24906a1bf48e13ddf544fdf0cab3cdc647d/README.md?plain=1#L41):
>     📋 Python 3.12 is currently not supported due to PyTorch Dynamo constraint

joerunde:
> Though that may be an out-of-date note. https://github.com/pytorch/pytorch/issues/120233#issuecomment-2089257409 indicates that it was made compatible in PyTorch 2.4.

Interesting, I had actually assumed that some part of fms-mo that used dynamo required updating based on changes in python 3.12. But you know what happens when you assume lol

@yannicks1 I opened an issue for handling dependencies here: https://github.com/vllm-project/vllm-spyre/issues/61

If you want to, I think I'd be fine merging this in without the dependencies specified, but I'll leave it up to you.

joerunde (approved):
LGTM given that we have issues open to add tests and specify dependencies correctly. Looks like it'll be a bit more work to get that all set up, I'm fine if we merge first and follow up with those

yannicks1:
Thanks @joerunde and @tjohnson31415 for digging a bit deeper here. Looks like a non-trivial thing to solve, will merge this and we deal with the dependencies in this issue #61 (thanks for opening this Joe).

PR #47, opened 2025-03-25
Ensure that warmup shapes are available after forking.

tdoublep:
I need to do a demo using V0 online serving today, and I noticed that it is currently not working using latest `aiu-vllm-dev` image:

Trying to deploy the following:
```
python3 -m vllm.entrypoints.openai.api_server --model /models/llama-194m/ --max-model-len=2048 --block-size=128
```
produces:
```
[SpyreWorker] load model...
ERROR 03-25 09:16:21 [engine.py:448] type object 'SpyrePlatform' has no attribute 'spyre_warmup_shapes'
ERROR 03-25 09:16:21 [engine.py:448] Traceback (most recent call last):
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/engine/multiprocessing/engine.py", line 436, in run_mp_engine
ERROR 03-25 09:16:21 [engine.py:448]     engine = MQLLMEngine.from_vllm_config(
ERROR 03-25 09:16:21 [engine.py:448]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/engine/multiprocessing/engine.py", line 128, in from_vllm_config
ERROR 03-25 09:16:21 [engine.py:448]     return cls(
ERROR 03-25 09:16:21 [engine.py:448]            ^^^^
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/engine/multiprocessing/engine.py", line 82, in __init__
ERROR 03-25 09:16:21 [engine.py:448]     self.engine = LLMEngine(*args, **kwargs)
ERROR 03-25 09:16:21 [engine.py:448]                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/engine/llm_engine.py", line 280, in __init__
ERROR 03-25 09:16:21 [engine.py:448]     self.model_executor = executor_class(vllm_config=vllm_config, )
ERROR 03-25 09:16:21 [engine.py:448]                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/executor/executor_base.py", line 52, in __init__
ERROR 03-25 09:16:21 [engine.py:448]     self._init_executor()
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/executor/uniproc_executor.py", line 47, in _init_executor
ERROR 03-25 09:16:21 [engine.py:448]     self.collective_rpc("load_model")
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/executor/uniproc_executor.py", line 56, in collective_rpc
ERROR 03-25 09:16:21 [engine.py:448]     answer = run_method(self.driver_worker, method, args, kwargs)
ERROR 03-25 09:16:21 [engine.py:448]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/utils.py", line 2216, in run_method
ERROR 03-25 09:16:21 [engine.py:448]     return func(*args, **kwargs)
ERROR 03-25 09:16:21 [engine.py:448]            ^^^^^^^^^^^^^^^^^^^^^
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm_spyre/worker/spyre_worker.py", line 149, in load_model
ERROR 03-25 09:16:21 [engine.py:448]     spyre_warmup_shapes = current_platform.get_warmup_shapes()
ERROR 03-25 09:16:21 [engine.py:448]                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 03-25 09:16:21 [engine.py:448]   File "/opt/vllm/lib64/python3.11/site-packages/vllm_spyre/platform.py", line 167, in get_warmup_shapes
ERROR 03-25 09:16:21 [engine.py:448]     return cls.spyre_warmup_shapes
ERROR 03-25 09:16:21 [engine.py:448]            ^^^^^^^^^^^^^^^^^^^^^^^
ERROR 03-25 09:16:21 [engine.py:448] AttributeError: type object 'SpyrePlatform' has no attribute 'spyre_warmup_shapes'
```

It looks like the engine process is getting forked **after** we parse the warmup shapes in the main process, and thus in the engine process they don't exist. 

It doesn't look like the platform is really the best place to store "state". Why don't we just use the config instead? I understand it is a little bit nasty since we can't "change" that class in the plugin, but it works very nicely.

### DIFF
--- a/vllm_spyre/core/scheduler.py
+++ b/vllm_spyre/core/scheduler.py
@@ -28,6 +28,8 @@ from vllm.sequence import (Sequence, SequenceData, SequenceGroup,
                            SequenceStatus)
 from vllm.utils import Device, PyObjectCache
 
+from vllm_spyre.platform import SpyrePlatform
+
 logger = init_logger(__name__)
 
 
@@ -143,6 +145,12 @@ class SpyreScheduler:
         for i in range(1, self.scheduler_config.max_num_partial_prefills + 1):
             self.partial_prefill_budget_lookup_list[i] = (
                 scheduler_config.max_num_batched_tokens // i)
+
+        # SPYRE SPECIFIC CODE BLOCK START
+        self.spyre_warmup_shapes = SpyrePlatform.get_warmup_shapes(
+            self.scheduler_config
+        )
+        # SPYRE SPECIFIC CODE BLOCK END
 
     @property
     def next_cache_id(self):
@@ -668,8 +676,7 @@ class SpyreScheduler:
         seq_groups: List[ScheduledSequenceGroup] = []
 
         # SPYRE SPECIFIC CODE BLOCK START
-        spyre_warmup_shapes = self.scheduler_config.spyre_warmup_shapes
-        applicable_spyre_warmup_shapes = list(spyre_warmup_shapes)
+        applicable_spyre_warmup_shapes = list(self.spyre_warmup_shapes)
         # SPYRE SPECIFIC CODE BLOCK END
 
         waiting_queue = self.waiting
--- a/vllm_spyre/core/scheduler.py
+++ b/vllm_spyre/core/scheduler.py
@@ -669,7 +669,7 @@ class SpyreScheduler:
         seq_groups: List[ScheduledSequenceGroup] = []
 
         # SPYRE SPECIFIC CODE BLOCK START
-        spyre_warmup_shapes = current_platform.get_warmup_shapes()
+        spyre_warmup_shapes = self.scheduler_config.spyre_warmup_shapes
         applicable_spyre_warmup_shapes = list(spyre_warmup_shapes)
         # SPYRE SPECIFIC CODE BLOCK END
 
--- a/vllm_spyre/core/scheduler.py
+++ b/vllm_spyre/core/scheduler.py
@@ -22,7 +22,6 @@ from vllm.core.scheduler import (ARTIFICIAL_PREEMPTION_MAX_CNT,
                                  scheduler_running_outputs_builder,
                                  seq_group_metadata_builder)
 from vllm.logger import init_logger
-from vllm.platforms import current_platform
 # SPYRE SPECIFIC CODE BLOCK END
 from vllm.sequence import (Sequence, SequenceData, SequenceGroup,
                            SequenceGroupMetadata, SequenceGroupMetadataDelta,
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -75,10 +75,10 @@ class SpyrePlatform(Platform):
 
         # Override --max-num-seqs to the biggest warmup batch size
         # And override --max-model-len to the biggest warmup sequence
-        cls.set_warmup_shapes(scheduler_config)
+        spyre_warmup_shapes = cls.get_warmup_shapes(scheduler_config)
         max_batch_size = 0
         max_seq_len = 0
-        for shape in scheduler_config.spyre_warmup_shapes:
+        for shape in spyre_warmup_shapes:
             max_batch_size = max(max_batch_size, shape['batch_size'])
             max_seq_len = max(max_batch_size,
                               shape['prompt_length'] + shape['new_tokens'])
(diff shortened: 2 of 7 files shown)


tdoublep:
Hmm this change doesn't sit nicely with the V0 unit tests because we only parse the environment variables once at import time, but the tests rely on changing them throughout the excecution.

maxdebayser:
So  the warmup shapes are being parsed in the main server process while the first worker process that also contains the engine in V0 is already up and running?

yannicks1 (approved):
I just fixed formatting and added the same logic to V1 classes. LGTM!

tjohnson31415:
> So the warmup shapes are being parsed in the main server process while the first worker process that also contains the engine in V0 is already up and running?

No, the worker process comes up after the main server process parses the warmup shapes, but the shapes aren't re-parsed in the worker. This problem is related to `VLLM_WORKER_MULTIPROC_METHOD`. For the initial v1 integration @joerunde put in code to force using `VLLM_WORKER_MULTIPROC_METHOD=fork`:
https://github.com/vllm-project/vllm-spyre/pull/6/files#diff-16ac04c4e75668ccde20cd2cfb82fa496d5d98dbe169ffede15652dc60a16066R54-R62
(this code could be removed now in this PR)

The default of `spawn` meant that the `spyre_warmup_shapes` added to the `SpyrePlatform` class instance did not exist in the spawned worker (the worker process doesn't call `set_warmup_shapes`). With `fork` the modifications to the class persist in the worker.

Moving the warmup shapes to the `scheduler_config` works because the `vllm_config` is serialized to the worker class during initialization regardless of spawn vs fork.

tdoublep:
@tjohnson31415 thank you for the correction + detailed explanation! I wrongly assumed that the default V0 behaviour was fork too. I will try it quickly now on main using fork. 

Either way, I still think we should consider moving the shapes into the config. In fact, I wonder why we are really using env variables at all. Wouldn't we rather be passing them as proper arguments? If vLLM does not provide a way for plugins to add their own custom args/config then perhaps this is something we could change upstream?

tjohnson31415:
Hmm, yeah, then this may be something else, not the serialization problem that I'm familiar with 🤔

joerunde:
> If vLLM does not provide a way for plugins to add their own custom args/config then perhaps this is something we could change upstream?

+1 on bringing that up upstream. I'd rather not have to do the entrypoint hijacking that the vllm tgis adapter does to add cli args. 

For this PR though, do we need to store the warmup shapes on a config object at all? It seems like they can be rebuilt in each worker from the environment variables, and I'd prefer doing that now for robustness until we work out some upstream changes for plugins to be able to add their own arguments and config properly.

joerunde:
Also.... we should probably have at least a single test that runs the openai server so we can catch these problems too!

tdoublep:
>Also.... we should probably have at least a single test that runs the openai server so we can catch these problems too!

Agreed. @dpatel-ops and I had discussed this before.

tdoublep:
>For this PR though, do we need to store the warmup shapes on a config object at all? It seems like they can be rebuilt in each worker from the environment variables, and I'd prefer doing that now for robustness until we work out some upstream changes for plugins to be able to add their own arguments and config properly.

@joerunde We can do it like that for now, yeah. I had an earlier attempt where I was doing that, just felt that cleaning in the config was cleaner since they are kept in "one place". It is not a huge difference though tbh.

joerunde:
Added a followup issue to correctly handle cli args and configs here: https://github.com/vllm-project/vllm-spyre/issues/51

PR #48, opened 2025-03-25
:sparkles: re-enable v1

joerunde:
Adds the new `supports_v1` platform api added upstream here: https://github.com/vllm-project/vllm/pull/15417

This removes one more blocker to moving past vllm v0.8.0


wallashss (approved):
LGTM!

PR #52, opened 2025-03-26
[Continuous batching] Initial cb test

nikolaospapandreou:
Initial CB implementation (for vLLM V1). Works with FMS model wrapper. 
Test with `offline_inference_spyre_cb_test.py`, set `VLLM_SPYRE_USE_CB` to 1 for continuous batching or 0 for static batching.


vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -70,21 +70,20 @@ def __init__(
...
         is_prompt: bool,
+        tkv: int,
+        active_pages: list[int],
     ) -> torch.Tensor:
 
-        if is_prompt:
-            self.tkv = 0
-            if not envs_spyre.VLLM_SPYRE_USE_CB:
-                self.model.past_key_value_states = None
+        self.tkv = tkv
```
tdoublep, line 83: Why do we set `self.tkv` here? It looks like it is not used.

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -155,16 +154,18 @@ def forward(
...
-        logits, self.tkv = self.model(
+        logits = self.model(
             input_ids,
             position_ids=positions,
             mask=masks,
             use_cache=True,
             only_last_token=True,
             tkv=self.tkv,
-            active_pages=[i for i in range(input_ids.shape[0])],
+            #active_pages=[i for i in range(input_ids.shape[0])],
```
tdoublep, line 164: nit: remove commented out lines

vllm_spyre/v1/core/scheduler.py
```diff
@@ -109,6 +113,88 @@ def schedule(self) -> "SchedulerOutput":
         outputs = super().schedule()
         return outputs
 
+    def schedule_cb(self) -> "SchedulerOutput":
```
tdoublep, line 116: I would propose have separate classes `StaticBatchingSpyreScheduler` and `ContinuousBatchingSpyreScheduler` and just implementing the `schedule` function differently, rather than having two functions. This could be addressed in the PR to main though, rather than this one.

vllm_spyre/v1/core/scheduler.py
```diff
@@ -109,6 +113,88 @@ def schedule(self) -> "SchedulerOutput":
...
+                if request.sampling_params is not None and\
+                        request.sampling_params.max_tokens is not None:
+                    max_tokens = request.sampling_params.max_tokens
+
+                available_warmup_shapes = [
+                    shape for shape in available_warmup_shapes
+                    if request.num_prompt_tokens <= shape['prompt_length']
+                    and max_tokens <= shape['new_tokens']
+                    and len(self.waiting) < shape['batch_size']
+                ]
```
tdoublep, line 148: I think that the continuous batching logic should not depend on the warmup shapes in this way?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -110,6 +112,24 @@ def __init__(
...
+        self._req_ids2idx_prompt: dict = {}
+        self._req_ids2idx_decode: dict = {}
+        self._decode_batch_size = 0
+        self._active_pages = []
+        self._free_page_idxs = []
+        self._position_ids_prompt: torch.Tensor = None
+        self._mask_prompt: torch.Tensor = None
+        self._tkv: int = 0
+        self._tkv2fms: int = 0
+        self._prev_step_dec = False
```
tdoublep, line 124: In my opinion, the management of pages here is implemented at the wrong level. We should not be trying to maintain all of this state in the model runner itself. We should be looking at how vLLM (V1) is implemented on GPU as a guide (e.g., we should be using something like the [InputBatch](https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/gpu_input_batch.py) class to maintain this state).  I think as a first attempt it is fine, and we can iteratively improve it from here.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -110,6 +112,24 @@ def __init__(
...
+
+        warmup_shapes = current_platform.get_warmup_shapes()
+        max_prompt_length = max(shape["prompt_length"]
+                                for shape in warmup_shapes)
+        max_batch_size = max(shape["batch_size"] for shape in warmup_shapes)
```
tdoublep, line 129: Continuous batching should not use warmup shapes

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -126,6 +146,155 @@ def load_model(self, prompt_lens: Iterable[int],
     def vocab_size(self) -> int:
         return self.model.model.model.config.src_vocab_size
 
+    def _prepare_prompt_cb(
+        self,
+        new_requests: List[NewRequestData],
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, List[int]]:
```
tdoublep, line 152: Similar to scheduler, I would suggest splitting into two model runner classes `SpyreModelRunnner` and `ContinuousBatchingSpyreModelRunner` or similar

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -346,13 +580,18 @@ def execute_model(
...
+            model_input = self.prepare_model_input_cb(scheduler_output)
+        else:
+            model_input = self.prepare_model_input(scheduler_output)
 
         hidden_states = self.model(
             input_ids=model_input.input_tokens,
             positions=model_input.input_positions,
             masks=model_input.input_masks,
             is_prompt=model_input.is_prompt,
+            tkv = self._tkv2fms,
```
tdoublep, line 593: Why is `self._tkv2fms` needed? Can't we just use `self.tkv`?

examples/offline_inference_spyre_cb_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+os.environ["VLLM_SPYRE_WARMUP_PROMPT_LENS"] = '64'
+os.environ["VLLM_SPYRE_WARMUP_NEW_TOKENS"] = str(max_tokens)
+os.environ['VLLM_SPYRE_WARMUP_BATCH_SIZES'] = '4'
+
+# defining here to be able to run/debug directly from VSC (not via terminal)
+os.environ['VLLM_SPYRE_DYNAMO_BACKEND'] = 'eager'
+os.environ['VLLM_SPYRE_USE_CB'] = '1'
+os.environ['VLLM_USE_V1'] = '1'
+
```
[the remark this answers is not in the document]
tdoublep: I don't understand why we need these new environment variables? They should just be controlled using normal vLLM arguments (`max-model-len` and `max-num-seqs` respectively).

vllm_spyre/platform.py
```diff
@@ -74,8 +78,22 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
-            # The v0 scheduler will run out of blocks if this is overridden
-            scheduler_config.max_num_seqs = max_batch_size
+            if envs_spyre.VLLM_SPYRE_USE_CB:
+                # For continuous batching we use max_num_seqs to control
+                # the max batch size respecting AIU Spyre KV cache size
+                scheduler_config.max_num_seqs =\
+                    envs_spyre.VLLM_SPYRE_MAX_BATCH_SIZE
+                # ToDo: this function check_and_update_config is called twice:
+                # 1st time scheduler_config.max_num_seqs is what user sets
+                # 2nd time scheduler_config.max_num_seqs is 128
```
tdoublep, line 88: I don't think we need to override `max-num-seqs` at all?

PR #46, opened 2025-03-24
[CI/BUILD] Use build cache for Spyre tests

ckadner:
**Changes in this PR:**

- [x] Update GH workflow for Spyre tests to ...
  - [x] build and cache a test image
  - [x] use the cached test image for both v0 and v1 tests
  - [x] run v0 and v1 tests in parallel
- [x] Add build stages to Dockerfile, with separate stage for tests
- [x] Add .dockerignore
- [x] [Delete untagged cache images](https://github.com/actions/delete-package-versions)
- [ ] AFTER MERGE: Update branch protection rules for `test-spyre` as merge requirement
  - <details><summary>Remove <code>test-spyre</code>:</summary>
    
    image
    </details>
  - <details><summary>Add <code>test-spyre-v0</code> and <code>test-spyre-v1</code>:</summary>
    
    image  
    </details>

**Optional:**
- [ ] Setup registry for large build cache images (Quay.io)
  - [ ] Setup Github secrets to authenticate to image registry (requires repo admin permissions)
- [x] Reconsider running tests sequentially if image uplodad/download time outweighs overall time savings
  - parallel wins if one of the 2 test suites runs for more than 3 mins (90s upload + 90s download/load)
- [ ] Consider building cache images for branches (release, dev, nighlty, feature-*, ...)

Resolves #34

PR #53, opened 2025-03-26
:alembic: Add tests for online server

### LINKED ISSUE #50: [tests] Add online test
### vLLM online mode testing

We currently only test vLLM in offline mode, hence don't catch issues with online (server/client) mode like for example [this](https://github.com/vllm-project/vllm-spyre/pull/47). 
Need to add at least one test in `vllm-spyre/tests` testing the online mode. 

template: [examples/spyre_warmup_online_client.py](https://github.com/vllm-project/vllm-spyre/blob/main/examples/spyre_warmup_online_client.py)

joerunde:
This PR adds tests for the online server, re-using some test setup from vllm to boot a server with `vllm serve ...` and use an `openai` client to interact with it. This already caught a bug with online serving with the latest vllm main as I wrote it 😎

I also started a `requirements-test.txt` file, I'll still need to get that hooked up in the gha workflow

Closes #50


tdoublep (approved):
LGTM - awesome thanks!

joerunde:
@yannicks1 Yes, and I think this PR is aiming to split the dockerfile up a bit so that there's a test-specific stage to address that issue: https://github.com/vllm-project/vllm-spyre/pull/46

tests/test_spyre_online.py
```diff
@@ -0,0 +1,173 @@
...
+import openai
+import pytest
+import requests
+from vllm.entrypoints.openai.cli_args import make_arg_parser
+from vllm.utils import FlexibleArgumentParser, get_open_port
+
+from tests.spyre_util import get_spyre_backend_list, get_spyre_model_list
+
+
+class RemoteOpenAIServer:
```
tjohnson31415, line 16: I think it is fine for now, but I'd like there to be a better way to use some of vLLM's test utilities instead of copying. Maybe the vllm test framework/utilities could become its own package for use in plugins.

tests/test_spyre_online.py
```diff
@@ -0,0 +1,173 @@
...
+        assert len(completion.choices) == 1
+        assert len(completion.choices[0].text) > 0
+
+        completion = client.completions.create(model=model,
+                                               prompt="Hello World!",
+                                               max_tokens=5,
+                                               temperature=1.0,
+                                               n=2)
+        assert len(completion.choices) == 2
+        assert len(completion.choices[0].text) > 0
```
wallashss, line 173: My 2 cents: I think it would be nice to have some 'unhappy' path as well. Something that would cause an error in the client, for instance, a bad input or something.

tests/test_spyre_online.py
joerunde: 👍  yeah that would be super cool

tests/test_spyre_online.py
joerunde: Yeah, I can toss a few in as well

tests/test_spyre_online.py
rafvasq: @joerunde In my PR #59, I moved this class to spyre_util.py, I wondered if it could be reused until we can make better use of vLLM's utils like Travis suggested.

joerunde: yup, just rebased this to make use of that. Thanks!

PR #54, opened 2025-03-26
[V1] Fix importings for schedule repackaging

### LINKED ISSUE #45: [V1] Update to the new scheduler interface
The V1 scheduler rework is still in progress, but the official scheduler interface has dropped. We don't actually need to implement against it, since our scheduler extends the stock v1 scheduler. However, everything was repackaged under `vllm.v1.core.sched`, so we'll need to update all of our imports and re-test.

This blocks upgrading past vllm v0.8.1

wallashss:
Fix #45 

I created a new file that we can import the needed classes from vllm and it solves depending the installed version (for backward compatibility). 

I also did some changes that are only related to TYPE_CHECKING

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -1,16 +1,22 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections import deque
-from typing import Deque
+from typing import TYPE_CHECKING, Deque
 
 from vllm.logger import init_logger
 from vllm.platforms import current_platform
 from vllm.sampling_params import SamplingParams
-from vllm.v1.core.scheduler import Scheduler
-from vllm.v1.core.scheduler_output import SchedulerOutput
 from vllm.v1.engine import EngineCoreOutput, EngineCoreOutputs, FinishReason
 from vllm.v1.outputs import ModelRunnerOutput
 from vllm.v1.request import Request, RequestStatus
+
+try:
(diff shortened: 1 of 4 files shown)


vllm_spyre/v1/core/scheduler.py
```diff
@@ -1,16 +1,21 @@
...
 from vllm.v1.outputs import ModelRunnerOutput
 from vllm.v1.request import Request, RequestStatus
 
+try:
+    from vllm.v1.core.sched.scheduler import Scheduler
+except ImportError:
+    from vllm.v1.core.scheduler import Scheduler
+
+if TYPE_CHECKING:
+    from vllm_spyre.v1.core.sched.output import SchedulerOutput
```
joerunde, line 18: You can add an
```
else:
    SchedulerOutput = None
```
here so that `SchedulerOutput` is always defined and then you can use it in your type hint instead of needing to use the string `"ShedulerOutput"`

joerunde: (ditto for the v1 model runner file)

wallashss:
HMMMM interesting!

> Personally I would slightly prefer following the if/else pattern on type checking so that we can use the real types in the hints, just because my VSCode happens to like that better

I'll gonna do that now. Wait a sec.

wallashss:
Done @joerunde

joerunde:
nice!

PR #59, opened 2025-03-27
Add tensor parallel tests for online server

### LINKED ISSUE #57: [tests] Add online tensor parallelism test
### vLLM online mode  tensor parallelism testing

Need for an online test supporting tensor parallelism in `vllm-spyre/tests`.

template: `vllm-spyre/tests/test_spyre_online.py`

rafvasq:
Adds an online TP case - closes #57.

### DIFF
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -33,8 +33,7 @@ class RemoteOpenAIServer:
                  env_dict: Optional[dict[str, str]] = None,
                  seed: Optional[int] = 0,
                  auto_port: bool = True,
-                 max_wait_seconds: Optional[float] = None,
-                 tensor_parallel_size: Optional[int] = 1) -> None:
+                 max_wait_seconds: Optional[float] = None) -> None:
         # NB: This implementation does not ensure that the model is downloaded
         # before booting the server, it should be used with models already
         # cached on disk
@@ -54,12 +53,6 @@ class RemoteOpenAIServer:
                                  f"when `seed={seed}`.")
 
             vllm_serve_args = vllm_serve_args + ["--seed", str(seed)]
-
-        if tensor_parallel_size > 1:
-            vllm_serve_args = vllm_serve_args + [
-                "--tensor-parallel-size",
-                str(tensor_parallel_size)
-            ]
 
         parser = FlexibleArgumentParser(
             description="vLLM's remote OpenAI server.")
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -1,16 +1,142 @@
 import math
 import os
-from typing import Any, Dict, List, Tuple, Union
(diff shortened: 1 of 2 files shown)


tests/test_spyre_online_multi.py
```diff
@@ -0,0 +1,57 @@
...
+
+@pytest.mark.parametrize("model", get_spyre_model_list())
+# (prompt_length/new_tokens/batch_size)
+@pytest.mark.parametrize("warmup_shape", [[
+    (64, 20, 4),
+]])
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+@pytest.mark.parametrize("tensor_parallel_size", [2])
+@pytest.mark.parametrize("vllm_version", ["V0", "V1"])
+def test_openai_serving(model, warmup_shape, backend, tensor_parallel_size,
```
joerunde, line 15: Nice!

Any thoughts on moving this over to `test_spyre_tensor_parallel.py`? I kinda think we should try to start organizing these files a bit instead of creating more of them. I can also move my basic online tests over to `test_spyre_basic.py`

tests/spyre_util.py
```diff
@@ -1,18 +1,144 @@
...
+
+    def __init__(self,
+                 model: str,
+                 vllm_serve_args: list[str],
+                 *,
+                 env_dict: Optional[dict[str, str]] = None,
+                 seed: Optional[int] = 0,
+                 auto_port: bool = True,
+                 max_wait_seconds: Optional[float] = None,
+                 tensor_parallel_size: Optional[int] = 1) -> None:
```
joerunde, line 37: Since this is only used to set args in `vllm_serve_args`, can we instead call this like
```
with RemoteOpenAIServer(
            model, ["--tensor-parallel-size", "1"], ...)
```
?

tests/test_spyre_online_multi.py
rafvasq: I think that's cleaner, I just wonder if in the future you could see a situation where online tests would have to be refactored/separated from offline. Up to you

rafvasq: Also, online tests are passing/exiting properly without --forked

tests/spyre_util.py
rafvasq: I updated the util and the test so that it would be called like you said but still through a pytest parameter.

tests/test_spyre_online_multi.py
joerunde: >I think that's cleaner, I just wonder if in the future you could see a situation where online tests would have to be refactored/separated from offline. Up to you

Yeah... that could end up being the case as well 🤷 
We can leave test organization until later

joerunde (approved):
🌶️

Issue #55, opened 2025-03-26
[V1] Regression on scheduler


wallashss:
We found a regression during the development of #54, an assert `spyre_model_runner.py` fails probably due to behavior change of vllm engine's scheduler. 

> Note: #54 is required to repro the issue due to broken imports caused by recent refactoring of vLLM  

Repro on:
```
commit 27df5199d99627e1eb101071c2155f888181bd64 (HEAD -> main, origin/main, origin/HEAD)
```

This simple offline script can consistently repro the issue.

``` py
from vllm import LLM, SamplingParams

# Define prompts and their corresponding sampling parameters
prompts = [
    "Hello, my name is",
    "The capital of France is",
    "The future of AI is"
]
sampling_params_list = [
    SamplingParams(seed=123, temperature=0.8, top_p=0.95),
    SamplingParams(seed=123, temperature=0.5, top_p=0.9),
    SamplingParams(seed=123, temperature=0.7, top_p=0.85)
]

model = "/models/llama-194m/"
llm = LLM(model=model, enforce_eager=False)

# Generate texts for each prompt with its sampling parameters
outputs = llm.generate(prompts, sampling_params_list)

# Print the outputs
for response in outputs:
    print(f"Prompt: (response.prompt!r), Generated text: {response.outputs[0].text!r}")

```

Outputs:

```
Prompt: (response.prompt!r), Generated text: " 5c1. I'm a teacher, and you teach me how to"
Prompt: (response.prompt!r), Generated text: ' Paris. It is located in the center of France and is the largest city in'
Prompt: (response.prompt!r), Generated text: ' in the hands of machine learning, which is the process of a computer learning to'
ERROR 03-26 20:20:17 [core.py:344] EngineCore hit an exception: Traceback (most recent call last):
ERROR 03-26 20:20:17 [core.py:344]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/v1/engine/core.py", line 337, in run_engine_core
ERROR 03-26 20:20:17 [core.py:344]     engine_core.run_busy_loop()
ERROR 03-26 20:20:17 [core.py:344]   File "/opt/vllm/lib64/python3.11/site-packages/vllm/v1/engine/core.py", line 371, in run_busy_loop
ERROR 03-26 20:20:17 [core.py:344]     outputs = step_fn()
ERROR 03-26 20:20:17 [core.py:344]               ^^^^^^^^^
[...]

PR #62, opened 2025-03-28
[V1] Fix assert failure when finishing a batch

wallashss:
This PR address #55 

What I observed using vLLM v0.8.0 vs the main of upstream: on upstram main there was an extra call for spyre_model_runner receiving `scheduler_output` with only `finished_request_ids` filled with the last request to clean up. However, for v0.8.0 this call did not happen, and looks like it was wrong because the worker did not have a chance to cleanup the data. If I understood correctly, This behavior was solved in this PR https://github.com/vllm-project/vllm/pull/14388, which was also the basis to write this PR to solve the issue.


joerunde:
Sweeet, thanks @wallashss!
Are we now fully working on v1 again on post-0.8.2 vllm? If so what do you think about updating the CI install to vllm@5f063a8 so that it tests your new logic?

wallashss:
> Are we now fully working on v1 again on post-0.8.2 vllm?

Probably, yes? I think we just have to make sure communicate the team that we are moving forward.

> If so what do you think about updating the CI install to vllm@5f063a8 (or later) so that it tests your new logic?

That means update `Dockerfile.spyre`, right?

joerunde:
> That means update Dockerfile.spyre, right?

Yeah, that way the CI running here has the latest v1 engine and can actually test out these changes

wallashss:
> Yeah, that way the CI running here has the latest v1 engine and can actually test out these changes

Thanks, I just wanted to make sure I seeing the right spot

joerunde:
@wallashss I think the `--depth 1` on the git clone there is causing it to not have that commit in history, and the subsequent `git fetch` is only fetching the tags

wallashss:
> @wallashss I think the `--depth 1` on the git clone there is causing it to not have that commit in history, and the subsequent `git fetch` is only fetching the tags

Yeah, I think it's good now.

joerunde:
nice, it looks like the test actually caught a problem!

wallashss:
Hey @joerunde, 

I think I need an opinion for the error that is failing.

Just to recap, the log from the test:


```
INFO 03-31 18:39:09 [spyre_model_runner.py:385] Padding request of length 1 tokens to 64 tokens.
ERROR 03-31 18:39:09 [async_llm.py:337] EngineCore output handler hit an error: 
ERROR 03-31 18:39:09 [async_llm.py:337] Traceback (most recent call last):
ERROR 03-31 18:39:09 [async_llm.py:337]   File "/usr/local/lib/python3.12/site-packages/vllm/v1/engine/async_llm.py", line 315, in _run_output_handler
ERROR 03-31 18:39:09 [async_llm.py:337]     processed_outputs = self.output_processor.process_outputs(
ERROR 03-31 18:39:09 [async_llm.py:337]                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 03-31 18:39:09 [async_llm.py:337]   File "/usr/local/lib/python3.12/site-packages/vllm/v1/engine/output_processor.py", line 318, in process_outputs
ERROR 03-31 18:39:09 [async_llm.py:337]     self._update_stats_from_output(req_state, engine_core_output,
ERROR 03-31 18:39:09 [async_llm.py:337]   File "/usr/local/lib/python3.12/site-packages/vllm/v1/engine/output_processor.py", line 382, in _update_stats_from_output
ERROR 03-31 18:39:09 [async_llm.py:337]     iteration_stats.update_from_output(engine_core_output,
ERROR 03-31 18:39:09 [async_llm.py:337]   File "/usr/local/lib/python3.12/site-packages/vllm/v1/metrics/stats.py", line 104, in update_from_output
ERROR 03-31 18:39:09 [async_llm.py:337]     assert num_new_generation_tokens > 0
ERROR 03-31 18:39:09 [async_llm.py:337]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 03-31 18:39:09 [async_llm.py:337] AssertionError
```

And it is caused by a [test](https://github.com/vllm-project/vllm-spyre/blob/c720b8fca41082aed730bf0dd6420813dfad56d7/tests/test_spyre_online.py#L67) with an interesting comment:

```
        # Short prompt under context length but requesting too many tokens for
        # the warmup shape should return an empty result
        completion = client.completions.create(model=model,
                                               prompt="Hello World!",
                                               max_tokens=25)
```


From this in vLLM code [line](https://github.com/vllm-project/vllm/blob/e6e3c55ef28f30bca855399419c61bb70af03db2/vllm/v1/core/sched/scheduler.py#L650),  I am pretty sure that this is the only place where EngineCoreOutput is instanced. And it is only created when there is a `new_token_ids`.

```py
            if new_token_ids:
                # Add EngineCoreOutput for this Request.
                outputs.append(
                    EngineCoreOutput(
                        request_id=req_id,
                        new_token_ids=new_token_ids,
                        finish_reason=request.get_finished_reason(),
                        new_logprobs=new_logprobs,
                        new_prompt_logprobs_tensors=prompt_logprobs_tensors,
                        stop_reason=request.stop_reason,
                        events=request.take_events()))
```

 However, in the specialization of the Scheduler in vllm-spyre plugin we have this [snippet](https://github.com/vllm-project/vllm-spyre/blob/c720b8fca41082aed730bf0dd6420813dfad56d7/vllm_spyre/v1/core/scheduler.py#L188)

```py
 for request in rejected_requests:
            queue.remove(request)
            reject_outputs.append(
                EngineCoreOutput(request.request_id,
                                 new_token_ids=[],
                                 finish_reason=FinishReason.ABORT,
                                 stop_reason="Request did not fit any warmup "
                                 "shape"))
            request.status = RequestStatus.FINISHED_ABORTED
            self._free_request(request)
            self.rejected_requests.remove(request.request_id)
```

What makes me think that we need to review this implementation, because it probably does not match with the engine design, and we may have issues like this in the future.  What do you think?

wallashss:
cc/ @tjohnson31415 I think we might have some insight from you as well. 😄

wallashss:
I opened this issue #68  to follow up this PR. I just added a workaround to pass the tests and warn users the limitation. Everything else should still works fine, but requests does not fit in warmup shapes will make vllm crash. I did not found a way from the plugin to disable the stats globally. I hope later we can get more updates from vllm upstream and make it proper working to remove these workarounds.

wallashss:
> @wallashss I had another thought we could try: What happens if we just return a dummy token in the output to satisy that assertion?

Already tested it... Unfortunately this [assert](https://github.com/vllm-project/vllm-spyre/blob/c720b8fca41082aed730bf0dd6420813dfad56d7/tests/test_spyre_online.py#L74) fails... So I think it is a matter of what decision we should take.

tjohnson31415:
Ah, well if it is just that test assertion, I think we should update the test; it could assert on the `finish_reason` and `stop_reason` instead. Eventually we will need a better way to reject these requests without the empty response anyways.

wallashss:
> Ah, well if it is just that test assertion, I think we should update the test; it could assert on the finish_reason and stop_reason instead. 

Ok... but this also means change the behavior of v0 as well, are you sure about that? My first look was not so good about that 😬 .

tests/spyre_util.py
```diff
@@ -163,10 +163,14 @@ def generate_spyre_vllm_output(model: str, prompts: List[str],
...
 
     results = []
+
     for req_output in vllm_outputs:
         result = {}
         result['text'] = req_output.outputs[0].text
-        result['token_ids'] = tuple(req_output.outputs[0].token_ids)
+        # TODO: Workaround for V1, if request does not fit in a warmup shape
+        # token_ids may be filled with -1.
+        token_ids = [t for t in req_output.outputs[0].token_ids if t >= 0]
```
joerunde, line 172: This makes me sad D: 
(But so does all the code in the scheduler that does this dummy scheduling anyway that I wrote)

Approved with sadness lol

joerunde (approved):
lgtm

tests/spyre_util.py
wallashss: Thank you with sadness 😅

Issue #61, opened 2025-03-27
Cleanly handle installing dependencies that require `torch`


joerunde:
We need to be able to specify our dependencies, like `vllm` and `fms-model-optimizer`, that each depend on `torch`.

Currently, for vllm we hack around this by installing it from source, using the provided `use_existing_torch.py` script to strip out torch from the set of dependencies. This is maybe okay for our own CI, but won't work for end user installs of the plugin, and doesn't scale to other dependencies that also require torch.

It should be possible to use `uv` to install the plugin and its dependencies while not overwriting the existing torch installation. See for example: https://github.com/astral-sh/uv/issues/9174#issuecomment-2482710850

Even if we have a tool-specific install process that requires `uv` to work, that would be way better than the current setup. This will help with easily installing and testing against multiple versions of vllm, for example.

PR #63, opened 2025-03-28
:construction_worker: use UV package manager

joerunde:
Fixes https://github.com/vllm-project/vllm-spyre/issues/61

This integrates our project with `uv`, so that we can cleanly reject torch upgrades while installing all dependencies. This also modernizes our `pyproject.toml` to contain the source of truth for our dependencies, and sorts them into groups.

The updates to the Dockerfile demonstrate this working, though I think the dockerfile may not be around for much longer.

This blast radius ended up a bit bigger than I originally intended, but...
- The old version of `yapf` relied on `lib2to3` which no longer exists in newer pythons, and has to be installed separately on the system. This broke the simpler `uv` setup flow on github actions, so It was easier to just upgrade. This caused some changes in formatting in the test files
- I wanted to remove the requirements files since the requirements are now all properly specified in the pyproject.toml file. The old `setup.py` file pretty much only existed to parse those, so it was removed as well. This required a couple mode edits to `pyproject.toml` to specify setuptools as the build backend and set the extra entrypoint up

We can now either use `uv` directly to install the plugin and manage our local dev environments, or have `uv` compile the dependencies back into requirements files to emulate the old pip-only workflow. I've updated the pr comment action to provide instructions for both options of installing the linting dependencies.


pyproject.toml
```diff
@@ -1,6 +1,30 @@
...
+# Never install torch, so that no dependencies can override it.
+# This requires that torch is installed separately in the target environment
+override-dependencies = [
+    "torch; sys_platform == 'never'",
+    "torchaudio; sys_platform == 'never'",
+    "torchvision; sys_platform == 'never'",
+]
+
+[tool.uv.sources]
+vllm = { git = "https://github.com/vllm-project/vllm", rev = "v0.8.0" }
```
joerunde, line 26: Specifying the source for vllm allows `uv` to run the install from source, specifying the `empty` target so that the gpu deps don't get installed

pyproject.toml
```diff
@@ -1,6 +1,46 @@
...
+[project]
+name = "vllm-spyre"
+description = "vLLM plugin for Spyre hardware support"
+readme = "README.md"
+license = {text = "Apache 2"}
+version = "0.0.0" # TODO dynamic versioning
+dependencies = [
+    "fms-model-optimizer>=0.2.0",
+    "ibm-fms==0.0.8",
+    "vllm",
```
joerunde, line 16: We can specify vllm and fms-model-optimizer, and don't have to worry about either of them overriding torch 👍

Dockerfile.spyre
```diff
@@ -15,21 +15,19 @@ RUN microdnf update -y && microdnf install -y \
 RUN ln -sf $(which python${PYTHON_VERSION}) /usr/bin/python && \
     ln -sf $(which pip${PYTHON_VERSION}) /usr/bin/pip
 
-# Download and install vllm ###########################################################
-RUN git clone --depth 1 https://github.com/vllm-project/vllm.git \
```
joerunde, line 19: no more installing vllm manually 😎😎😎

joerunde:
@tdoublep I put all the code linting checks together into one job, so if we want to merge this we can remove the requirement on the `codespell` and `yapf` checks that no longer exist.

.github/workflows/lint_code.yml
```diff
@@ -36,17 +36,24 @@ jobs:
...
+        # using `always()` here ensures all the checks run even if previous
+        # checks fail
+        if: always()
+        run: isort . --check-only
+      - name: run yapf
+        if: always()
+        run: yapf --diff --recursive .
+      - name: Spelling check with codespell
+        if: always()
+        run: codespell --toml pyproject.toml
```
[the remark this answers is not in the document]
joerunde: We could, though right now that's running on multiple Python versions. I assume that's to make sure that mypy will be happy with any version of Python that somebody wants to develop locally with. But I guess it's also not too hard to just run all these checks with every Python version as well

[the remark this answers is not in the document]
tdoublep: I don't think we really need to run it with every python version, this is just inherited from upstream checks.

tdoublep (approved):
LGTM

Issue #36, opened 2025-03-19
Request cancellation can cause the server to crash


maxdebayser:
When I canceled a running benchmark execution vLLM crashed due to a key error:

```
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-01f682abe6b641df8f23bace775565d5-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-070db6bbeb3649c09aec12fbd43b1ff1-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-0d97f8112a9e403cb315e3a01c33d071-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-c64a088e123e472d9a495798cc210176-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-080c61e593dc48d19a2cb461d56b00cd-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-5ca5c5e83bf541268821a5f43657944e-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-f07e67efe202432b9f793a9b81353826-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-145235b34e4f47a291fcf6f546ca2cab-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-d6b0b237be5442c18218ba6d2efe0606-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-acb33c6575a2450782ff9dbee50b7103-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-55d2c6afc05e48ec85c46c8679c613e2-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-be060f7d51b34b6caac84380e3d3a751-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-8fc6a43694834bb28ad7bb5b81459d7b-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-81d1cd38b370403d94ccc917b1d14894-0.
INFO 03-19 19:08:31 [engine.py:307] Aborted request cmpl-82a88c3b1aef4635b592adab75bad56a-0.
[...]

PR #64, opened 2025-03-28
[Bugfix][V0][V1] Fix crashes from cancelling requests

tjohnson31415:
Some fixes from testing handling of request cancellation:
- in V0, guard against a KeyError in `_req_ids2idx`
- in v1, specialize the Scheduler's `finish_requests()` to handle the holdback_queue

FIX https://github.com/vllm-project/vllm-spyre/issues/36

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections import deque
-from typing import TYPE_CHECKING, Deque, Iterable, Union
+from typing import TYPE_CHECKING, Deque
 
 from vllm.logger import init_logger
 from vllm.sampling_params import SamplingParams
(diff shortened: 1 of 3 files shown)


vllm_spyre/v1/core/scheduler.py
```diff
@@ -140,6 +140,43 @@ def get_num_unfinished_requests(self) -> int:
...
+                # Invalid request ID.
+                continue
+
+            if request.status == RequestStatus.RUNNING:
+                self.running.remove(request)
+                self.scheduled_req_ids.discard(request.request_id)
+            else:
+                # this try-except is the specialization for Spyre
+                try:
+                    self.holdback_queue.remove(request)
```
joerunde, line 173: Ah dang, this is unfortunate.

I think maybe we can fix this in a simpler way by removing `self.holdback_queue` as an instance attribute, and instead just make it a local variable during `self.schedule()`. After we schedule a new batch, we can take all the requests that we held back and put them back in `self.waiting`, and then we won't need to worry about breaking assumptions that the v1 scheduler has

joerunde: That would let us get rid of the override on `get_num_unfinished_requests` as well

tjohnson31415: Done!
I left it as an instance variable so that we don't remake the deque. It is also still used in `_handle_rejects`, though there probably can't be rejection during scheduling?

joerunde: Right, I think that it should be safe to remove it from usage in `_handle_rejects` as well since this should all be synchronous. The output processing that calls `_handle_rejects` can't be happening concurrently with scheduling a new forward pass

joerunde: I'll leave it up to you on merging now vs. pulling out of _handle_rejects as well. I think we'll be getting rid of this rejected request business soon anyway

PR #67, opened 2025-03-31
[Continuous Batching] Introducing new env variables

yannicks1:
### Introducing new env variables

Adding two new env variable determining the dimension of the KV cache on AIU Spyre: 
- `VLLM_SPYRE_MAX_BATCH_SIZE`
- `VLLM_SPYRE_MAX_CONTEXT_LENGTH`

Note: variables will be interacting with AIU Spyre compiler.


nikolaospapandreou (approved):
LGTM

PR #70, opened 2025-04-04
[CI] Test against vllm/main

### LINKED ISSUE #41: [CI] Test against vllm/main
The upstream vllm code moves _fast_, usually around 30 PRs merged per day. It's impossible for maintainers or contributors to fully understand the needs of all users and the implications that each change has on them. There will be frequent changes that affect the plugin, sometime in breaking ways, and it'll be up to us to catch them and remediate as soon as we can.

Over in opendatahub, we handled this by running our tests in multiple environments, one of which installed the latest vllm main commit: https://github.com/opendatahub-io/vllm-tgis-adapter/blob/main/.github/workflows/tests.yaml#L38C1-L39C1

This has some obvious drawbacks, as sometimes the code on main may not be stable, and when a breaking change is introduced that requires upstream fixing, the job will fail for quite a while. We worked around those failures by force-merging where necessary, though maybe here we could come up with a better system like a tag that can optionally skip the test job against `vllm/main` when we know that it's expected to fail and don't have a fix ready.

ckadner:
Changes:

- Install Python package on GH action runner instead of building Docker image
- Use uv instead of pip as package manager
- Use matrix to test against different build targets
- Use matrix strategy to run v0 and v1 tests in parallel

Resolves #41

---

**TODO:**

- [ ] AFTER MERGE: Update branch protection rules for `test-spyre` as merge requirement
  - Remove <code>test-spyre</code>
  - Add `Test/V0 (vLLM:0.8.3)`, `Test/V1 (vLLM:0.8.3)`
  - Leave off `Test/V0 (vLLM:main)`, `Test/V1 (vLLM:main)` as a required check

### DIFF
--- a/.github/workflows/test-spyre.yml
+++ b/.github/workflows/test-spyre.yml
@@ -1,4 +1,4 @@
-name: test-sypre
+name: Test
 
 on:
   # Don't use pull_request.paths filter since this workflow is required for
@@ -17,14 +17,9 @@ on:
   workflow_dispatch:
 
 env:
-  # force output to be colored for non-tty GHA runner shell
   FORCE_COLOR: "1"
-  # prefer index for torch cpu version and match pip's extra index policy
-  UV_EXTRA_INDEX_URL: "https://download.pytorch.org/whl/cpu"
-  UV_INDEX_STRATEGY: "unsafe-best-match"
-  # facilitate testing by building vLLM for CPU when needed
   VLLM_CPU_DISABLE_AVX512: "true"
-  VLLM_TARGET_DEVICE: "cpu"
+  VLLM_TARGET_DEVICE: "empty"
   VLLM_PLUGINS: "spyre"
   VLLM_SPYRE_TEST_MODEL_DIR: "${{ github.workspace }}/models"
   HF_HUB_CACHE: "${{ github.workspace }}/.cache/huggingface/hub"
@@ -34,7 +29,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test-spyre:
+  test:
     timeout-minutes: 20
     runs-on: ${{ matrix.os }}
     strategy:
@@ -47,8 +42,6 @@ jobs:
             repo: "git+https://github.com/vllm-project/vllm --tag v0.8.0"
           - name: "vLLM:main"
             repo: "git+https://github.com/vllm-project/vllm --branch main"
-          - name: "ODH:main"
-            repo: "git+https://github.com/opendatahub-io/vllm --branch main"
         test_suite:
           - name: "V0"
             tests: "V0 and eager"
@@ -56,11 +49,6 @@ jobs:
           - name: "V1"
             tests: "(V1- and eager) or test_sampling_metadata_in_input_batch"
             flags: "--timeout=300 --forked"
-        exclude:
-          - vllm_version: { name: "vLLM:main" }
-            test_suite:   { name: "V1" }
-          - vllm_version: { name: "ODH:main" }
-            test_suite:   { name: "V1" }
 
     name: "${{ matrix.test_suite.name }} (${{ matrix.vllm_version.name }})"
 
@@ -84,29 +72,31 @@ jobs:
           cache-dependency-glob: |
             pyproject.toml
 
-      - name: "Install vLLM"
-        env:
-          VLLM_TARGET_DEVICE: empty
+      - name: "Install vLLM with Spyre plugin"
         run: |
-          # Install markupsafe from PyPI, Torch CPU index only has wheels for Python 3.13
-          uv add markupsafe --index force_pypi_index=https://pypi.org/simple
           uv add ${{ matrix.vllm_version.repo }}
           uv venv .venv --system-site-packages
           source .venv/bin/activate
-          uv pip install -v -e .
+          uv pip install -v .
           uv sync --frozen --group dev
 
       - name: "Download models"
         run: |
           mkdir -p "${VLLM_SPYRE_TEST_MODEL_DIR}"
-          
-          python -c "from transformers import pipeline; pipeline(\"text-generation\", model=\"JackFram/llama-160m\")"
-          VARIANT=$(ls "${HF_HUB_CACHE}/models--JackFram--llama-160m/snapshots/")
-          ln -s "${HF_HUB_CACHE}/models--JackFram--llama-160m/snapshots/${VARIANT}" "${VLLM_SPYRE_TEST_MODEL_DIR}/llama-194m"
-          
-          python -c "from sentence_transformers import SentenceTransformer; SentenceTransformer(\"sentence-transformers/all-roberta-large-v1\")"
-          VARIANT=$(ls "${HF_HUB_CACHE}/models--sentence-transformers--all-roberta-large-v1/snapshots/")
-          ln -s "${HF_HUB_CACHE}/models--sentence-transformers--all-roberta-large-v1/snapshots/${VARIANT}" "${VLLM_SPYRE_TEST_MODEL_DIR}/all-roberta-large-v1"
+
+          download_jackfram_llama() {
+            python -c "from transformers import pipeline; pipeline('text-generation', model='JackFram/llama-160m')"
+            VARIANT=$(ls "${HF_HUB_CACHE}/models--JackFram--llama-160m/snapshots/")
+            ln -s "${HF_HUB_CACHE}/models--JackFram--llama-160m/snapshots/${VARIANT}" "${VLLM_SPYRE_TEST_MODEL_DIR}/llama-194m"
+          }
+          download_roberta_large() {
+            python -c "from sentence_transformers import SentenceTransformer; SentenceTransformer('sentence-transformers/all-roberta-large-v1')"
+            VARIANT=$(ls "${HF_HUB_CACHE}/models--sentence-transformers--all-roberta-large-v1/snapshots/")
+            ln -s "${HF_HUB_CACHE}/models--sentence-transformers--all-roberta-large-v1/snapshots/${VARIANT}" "${VLLM_SPYRE_TEST_MODEL_DIR}/all-roberta-large-v1"
+          }
+          download_jackfram_llama &
+          download_roberta_large &
+          wait
 
       - name: "Run tests"
         env:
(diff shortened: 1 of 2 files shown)


ckadner:
There is a potential for another speed-up by 10 seconds, if we cache the installed entire Python site packages directory after PyTorch was installed. Cache key would depend on the PyTorch version, Python version, and OS.

.github/workflows/test-spyre.yml
```diff
@@ -1,29 +1,119 @@
...
+      - "vllm_spyre/**/*.py"
+      - pyproject.toml
+      - .github/workflows/test-spyre.yml
+  workflow_dispatch:
+
+env:
+  # force output to be colored for non-tty GHA runner shell
+  FORCE_COLOR: "1"
+  # prefer index for torch cpu version and match pip's extra index policy
+  UV_EXTRA_INDEX_URL: "https://download.pytorch.org/whl/cpu"
```
joerunde, line 23: Is this required? I see in the torch install step below we use `pip` and set the index url directly

.github/workflows/test-spyre.yml
```diff
@@ -1,29 +1,119 @@
...
+      - name: "Install vLLM"
+        env:
+          VLLM_TARGET_DEVICE: empty
+        run: |
+          # Install markupsafe from PyPI, Torch CPU index only has wheels for Python 3.13
+          uv add markupsafe --index force_pypi_index=https://pypi.org/simple
+          uv add ${{ matrix.vllm_version.repo }}
+          uv venv .venv --system-site-packages
+          source .venv/bin/activate
+          uv pip install -v -e .
```
joerunde, line 96: Why the editable flag here? 🤔

.github/workflows/test-spyre.yml
```diff
@@ -1,29 +1,119 @@
...
+        test_suite:
+          - name: "V0"
+            tests: "V0 and eager"
+            flags: "--timeout=300"
+          - name: "V1"
+            tests: "(V1- and eager) or test_sampling_metadata_in_input_batch"
+            flags: "--timeout=300 --forked"
+        exclude:
+          - vllm_version: { name: "vLLM:main" }
+            test_suite:   { name: "V1" }
```
joerunde, line 61: Is there a way to have this run, but mark it as not required to pass instead? We need to be running the v1 tests against main so that we know what fails ahead of the next vllm release

.github/workflows/test-spyre.yml
```diff
@@ -1,29 +1,119 @@
...
+
+env:
+  # force output to be colored for non-tty GHA runner shell
+  FORCE_COLOR: "1"
+  # prefer index for torch cpu version and match pip's extra index policy
+  UV_EXTRA_INDEX_URL: "https://download.pytorch.org/whl/cpu"
+  UV_INDEX_STRATEGY: "unsafe-best-match"
+  # facilitate testing by building vLLM for CPU when needed
+  VLLM_CPU_DISABLE_AVX512: "true"
+  VLLM_TARGET_DEVICE: "cpu"
```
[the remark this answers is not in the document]
joerunde: Ah, yeah I see it's set to `empty` below. What are these settings for?

.github/workflows/test-spyre.yml
```diff
@@ -1,29 +1,119 @@
...
+        run: |
+          # Install markupsafe from PyPI, Torch CPU index only has wheels for Python 3.13
+          uv add markupsafe --index force_pypi_index=https://pypi.org/simple
+          uv add ${{ matrix.vllm_version.repo }}
+          uv venv .venv --system-site-packages
+          source .venv/bin/activate
+          uv pip install -v -e .
+          uv sync --frozen --group dev
+
+      - name: "Download models"
```
joerunde, line 99: Bonus points if we could cache these, but definitely not necessary for this PR

joerunde:
Thanks @ckadner!

I mainly just had a lot of little questions, overall this looks super great

.github/workflows/test-spyre.yml
ckadner: Minus points actually :-)

- the download time from GHA cache is about equal to the download time from HF using the Python processes 
  - restore from GHA cache
  - download from HF


- the two models take up about 1.8 GB of cache (10 GB limit)

GHA cache makes most sense for operations that take up a lot of compute time, not when time is spent on downloads.

.github/workflows/test-spyre.yml
ckadner: I had done that, so I can separate the "local" Spyre code from the rest of the pip-installed upstream packages and cache the entire Python site-packages folder without polluting/invalidating cache by any "local" code changes.

In this version of the code that can be removed.

.github/workflows/test-spyre.yml
ckadner: I can speed up the HF download times by a few seconds by running the two Python processes in "parallel":

```sh
      - name: "Download models"
        run: |
          mkdir -p "${VLLM_SPYRE_TEST_MODEL_DIR}"
          download_jackfram_llama() {
            python -c "from transformers import pipeline; pipeline('text-generation', model='JackFram/llama-160m')"
            VARIANT=$(ls "${HF_HUB_CACHE}/models--JackFram--llama-160m/snapshots/")
            ln -s "${HF_HUB_CACHE}/models--JackFram--llama-160m/snapshots/${VARIANT}" "${VLLM_SPYRE_TEST_MODEL_DIR}/llama-194m"
          }
          download_roberta_large() {
            python -c "from sentence_transformers import SentenceTransformer; SentenceTransformer('sentence-transformers/all-roberta-large-v1')"
            VARIANT=$(ls "${HF_HUB_CACHE}/models--sentence-transformers--all-roberta-large-v1/snapshots/")
            ln -s "${HF_HUB_CACHE}/models--sentence-transformers--all-roberta-large-v1/snapshots/${VARIANT}" "${VLLM_SPYRE_TEST_MODEL_DIR}/all-roberta-large-v1"
          }
          download_jackfram_llama &
          download_roberta_large &
          wait
```

.github/workflows/test-spyre.yml
```diff
@@ -1,29 +1,119 @@
...
+      matrix:
+        os: ["ubuntu-latest"]
+        python_version: ["3.12"]
+        vllm_version:
+          - name: "vLLM:v0.8.0"
+            repo: "git+https://github.com/vllm-project/vllm --tag v0.8.0"
+          - name: "vLLM:main"
+            repo: "git+https://github.com/vllm-project/vllm --branch main"
+          - name: "ODH:main"
+            repo: "git+https://github.com/opendatahub-io/vllm --branch main"
```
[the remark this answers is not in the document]
ckadner: Okay. I will remove it.

.github/workflows/test-spyre.yml
ckadner: Yes, we would have to update the repository settings to change the branch protection rules for `test-spyre` as merge requirement, similar to what I had outline in PR #46 ...

> 
> 
> - [ ] AFTER MERGE: Update branch protection rules for `test-spyre` as merge requirement
>   - <details><summary>Remove <code>test-spyre</code>:</summary>
>     
>     image
>     </details>
>   - <details><summary>Add <code>test-spyre-v0</code> and <code>test-spyre-v1</code>:</summary>
>     
>     image  
>     </details>

.github/workflows/test-spyre.yml
joerunde: Ah, nice!

I was thinking more along the lines of reliability rather than speed here, since the upstream vllm CI downloads a tons of models in parallel from HF and often flakes out when a download fails. But this test suite is still small enough that it's probably fine to keep pulling from HF for now. We can always switch to gha cache if it becomes a problem.

.github/workflows/test-spyre.yml
```diff
@@ -1,29 +1,119 @@
...
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "latest"
+          python-version: ${{ matrix.python_version }}
+          enable-cache: true
+          ignore-nothing-to-cache: true
+          cache-dependency-glob: |
+            pyproject.toml
+
+      - name: "Install vLLM"
```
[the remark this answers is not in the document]
ckadner: Yes, in earlier versions I had named this step `"Install vLLM with Spyre plugin"`

But I like shorter names so I had truncated it.

I will restore the longer more descriptive version.

.github/workflows/test-spyre.yml
joerunde: Ahh, super interesting!

Is that also why we're creating a new venv here, so that we don't pollute the cached system site packages?

.github/workflows/test-spyre.yml
joerunde: Ah, gotcha. In that case I'll leave off the v1 + main job as a required check

.github/workflows/test-spyre.yml
ckadner: I see your point :-) 

https://github.com/vllm-project/vllm-spyre/actions/runs/14342866903/job/40206277191?pr=70#step:6:34

```
huggingface_hub.errors.HfHubHTTPError: 403 Forbidden: None.
Cannot access content at: https://huggingface.co/JackFram/llama-160m/resolve/main/config.json.
Make sure your token has the correct permissions.
```

I will do the hub cache

joerunde: oh lol, that was fast!

And of course any comments about the limitations would be great so the next maintainer knows not to try to stick a 7GB model in here

ckadner: @joerunde -- took me a bit to get cache updates to work properly with immutable caches. I pushed another commit that should:
- only create cache blobs for one of the matrix jobs
- not creating cache blobs for PR branches
- updating cache blobs on push to main when new models get added or old ones removed

joerunde:
Looks like we need another main merge and to test against 0.8.3

ckadner:
> > Looks like we need another main merge and to test against 0.8.3
> 
> @joerunde -- Merged. Updated to 0.8.3.
> 
> Should I make the `0.8.x` matrix version dynamic, i.e. reading it from `pyproject.toml`?

I made a quick change to update the job matrix to test against the project `default` version of vLLM and `vLLM:main`. 

A prettier version of this workflow could have a pre-test job to pull the actual version out of the `pyproject.toml` and then properly set the job matrix name. But using `default` will be better as the repository branch protection rules can not have dynamic merge requirements.

PR #72, opened 2025-04-04
[Tests] Adds GPTQ tests

### LINKED ISSUE #56: [tests] Add GPTQ test
GPTQ inference testing

We are currently not testing any GPTQ model on AIU Spyre. 
Should add at least one test in `vllm-spyre/tests testing` the GPTQ pipeline.

rafvasq:
Adds GPTQ test toward to close #56.

- Increases the timeout for server startup 
- Uses `get_test_combinations` to integrate gptq testing with basic online testing since it only works on one backend


tests/test_spyre_online_gptq.py
```diff
@@ -0,0 +1,49 @@
...
+from tests.spyre_util import RemoteOpenAIServer, get_spyre_model_list
+
+
+@pytest.mark.parametrize("model", get_spyre_model_list(isGPTQ=True))
+@pytest.mark.parametrize("warmup_shape", [[
+    (64, 20, 4),
+]])
+@pytest.mark.parametrize("backend", ["sendnn_decoder"])
+@pytest.mark.parametrize("quantization", ["gptq"])
+@pytest.mark.parametrize("vllm_version", ["V0"])
```
yannicks1, line 12: (I might have missed that) Is GPTQ only working for V0? Maybe @dpatel-ops knows more here?

rafvasq: It works on both from what I've tested, updated

yannicks1 (approved):
Awesome changes, thanks! Much cleaner now. 
I see tests were passing before pulling in main again with the latest commit. Can be merged IMO

PR #78, opened 2025-04-08
[Tests] Refactor test suite

### LINKED ISSUE #49: Reorganize tests for V0 and V1
After the merge of #24  there are now tests that only make senses for V1 and the test setup does not consider that. 

We must update the actions of `test-spyre.yml` and also possibly repackage the tests as mentioned in this [comment](https://github.com/vllm-project/vllm-spyre/pull/24#discussion_r2012741189)

rafvasq:
Reorganizes tests to close #49 

- Organized tests into `e2e` or `worker` directories, current `worker` test doesn't require parameters like the rest
- Created and refactored online tests to use a fixture for openai server that can include server arguments like TP


tests/conftest.py
```diff
@@ -49,3 +50,42 @@ def cleanup_fixture(should_do_global_cleanup_after_test: bool):
...
+def remote_openai_server(request):
+    """ Fixture to set up a test server."""
+
+    params = request.node.callspec.params
+
+    # Extract parameters from the test function for the server
+    model = params['model']
+    warmup_shape = params['warmup_shape']
+    backend = params['backend']
+    vllm_version = params['vllm_version']
```
wallashss, line 65: In general I liked how the wrapping works here. However, what happens if the user forget to declare this params? Should we throw an exception or try to get a better message to inform the user about this?

.github/workflows/test-spyre.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           export MASTER_ADDR=localhost && \
           export DISTRIBUTED_STRATEGY_IGNORE_MODULES=WordEmbedding && \
           cd vllm-spyre && \
-          python -m pytest --timeout=300  tests -v -k "V0 and eager" && \
-          python -m pytest --forked --timeout=300  tests -v -k "(V1- and eager) or test_sampling_metadata_in_input_batch"
+          python -m pytest --timeout=300  tests/worker -v && \
```
wallashss, line 27: At first I was wondering if `test_spyre_input_batch` should run in the same batch as V1, because it is a test that only makes sense for v1 engine. But I think the e2e have a different purpose, so I guess it is nice how you separated. 

How should we evolve this folder structure? Besides the e2e tests, should we reflect the module structure? If so, I think the only one tests at worker should be at `v1/worker`.

wallashss:
In general LGTM, just some quick question before we move forward this PR. Good job!

PR #79, opened 2025-04-09
[CB ] e2e continuous batching tests

prashantgupta24:
Adding some continuous batching unit tests

cc @yannicks1 @maxdebayser 

Also fixes https://github.com/vllm-project/vllm-spyre/issues/151 - basically installs fms from custom branch before running the `cb` tests.

Once the fms custom branch is merged to main, we can delete the duplicated `test_batch_handling` test within `tests/e2e/test_spyre_basic.py` and have the test within `test_spyre_cb` test both CB and non-cb conditions. We can also remove the `cb` flag and have the test run with other `e2e` tests.


tests/test_spyre_cb.py
```diff
@@ -0,0 +1,153 @@
...
+        m.setenv("VLLM_SPYRE_MAX_BATCH_SIZE", str(2))
+        m.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+
+        # start the engine
+        engine_args = EngineArgs(
+            model=model,
+            enforce_eager=True,  # reduce test time
+        )
+
+        engine = V1LLMEngine.from_engine_args(engine_args)
```
prashantgupta24, line 116: might be easier to set up `engine_core` and loop through the `step` function for it, that way we can directly get the `waiting` and `running` queues. Will give it a shot next

tests/test_spyre_cb.py
```diff
@@ -0,0 +1,282 @@
...
+        assert len(engine_core.scheduler.waiting) == 0
+        assert len(engine_core.scheduler.running) == 2
+
+        _ = engine_core.step()
+        assert len(engine_core.scheduler.waiting) == 0
+        assert len(engine_core.scheduler.running) == 2
+
+        _ = engine_core.step()
+        assert len(engine_core.scheduler.waiting) == 0
+        assert len(engine_core.scheduler.running) == 2
```
prashantgupta24, line 282: Hmm, it seems like this goes on for more than I understand, request[0] should have hit the `stop_reason` already and ejected from `running` queue, but it stays here until it hits its `max_tokens` 🤔 Perhaps I'm understanding this wrong.

tests/test_spyre_cb.py
```diff
@@ -0,0 +1,282 @@
...
+                                          stop="1",
+                                          ignore_eos=True)
+
+        sampling_params2 = SamplingParams(max_tokens=max_tokens2,
+                                          temperature=0.0,
+                                          stop="1",
+                                          ignore_eos=True)
+
+        prompt1 = "7 6 5 4"
+        prompt2 = "10 9 8 7"
```
maxdebayser, line 103: Can you perhaps test with 3 prompts and a max batch size of 2 so that we can verify that the third prompt stays in the waiting state until on of the two others leaves the batch?

tests/test_spyre_cb.py
```diff
@@ -0,0 +1,282 @@
...
+        m.setenv("VLLM_SPYRE_MAX_BATCH_SIZE", str(2))
+        m.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+
+        # start the engine
+        engine_args = EngineArgs(
+            model=model,
+            enforce_eager=True,  # reduce test time
+        )
+
+        engine = V1LLMEngine.from_engine_args(engine_args)
```
maxdebayser, line 123: Just thinking aloud: if you set `envs.VLLM_ENABLE_V1_MULTIPROCESSING = False`, `from_engine_args()` should create an engine that has an `InprocClient` for the engine core and then the engine step will be in lockstep with the engine core step. So you could combine both tests if you want. Although perhaps it would make the combined test a little harder to read.

maxdebayser:
Nice, this is going in the right direction.

tests/test_spyre_cb.py
prashantgupta24: Hmm, yeah it's hard for me already to wrap my head around that :) Let me wrap up the basic tests first then I can try to take a look at this if that works

tests/test_spyre_cb.py
prashantgupta24: Done

tests/test_spyre_cb.py
```diff
@@ -0,0 +1,282 @@
...
+    # order given
+    prompts = [
+        "7 6 5 4",
+        "10 9 8 7",
+        "8 7 6 5",
+    ]
+
+    # Ensure that both:
+    # - The model doesn't crash
+    # - The output sequences are correct
```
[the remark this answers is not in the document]
prashantgupta24: Uh combining this test with the `test_spyre_basic` is proving to be tricky. There are a number of differences between the 2 tests:
1. Different backends (cb still doesn't support AIU)
1. CB doesn't support v0
1. `logprobs` within `SamplingParams` doesn't work with cb (still investigating)

tests/test_spyre_cb.py
prashantgupta24: " Stop strings require detokenizing the output. Detokenization is handled in the output processor that runs at the AsyncLLM and LLMEngine level
So, we stream tokens from EngineCore -> AsyncLlm. AsyncLLM checks for stop strings in the OutputProcessor class. If a stop string is detected, we stop generating and send an abort message back to EngineCore to terminate the request in the EngineCore " 

This is why the request wasn't stopping using the `stop` string

tests/test_spyre_cb.py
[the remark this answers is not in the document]
prashantgupta24: If we can remove `logprobs` from the SamplingParams for `tests/test_spyre_basic.py` (which I think is fine since we're not really doing anything with them), we can reuse the tests. But I'm also curious as to why it's not working with cb

prashantgupta24:
Interestingly enough, 6 prompts with 2 or 3 `max_num_seqs` immediately kills the test without any exception.

6 prompts with 4 or 5 `max_num_seqs` fails as seen before, while 1 and 6 pass as expected. 
```
test_spyre_cb.py::test_cb_handling[prompts5-1-eager-/models/llama-194m-2] Killed
```

test_spyre_cb.py::test_cb_handling[prompts5-1-eager-/models/llama-194m-4] FAILED                                                                                                                                                       [ 10%]
test_spyre_cb.py::test_cb_handling[prompts5-1-eager-/models/llama-194m-5] FAILED                                                                                                                                                       [ 20%]
test_spyre_cb.py::test_cb_handling[prompts5-1-eager-/models/llama-194m-6] PASSED                                                                                                                                                       [ 30%]
test_spyre_cb.py::test_cb_handling[prompts5-1-eager-/models/llama-194m-1] PASSED                                                                                                                                                       [ 40%]
test_spyre_cb.py::test_cb_handling[prompts5-1-eager-/models/llama-194m-3] Killed
```

**Comment by prashantgupta24** [comment]
Spent almost the whole day today unsuccessfully trying to set up a local working env, ugh! Hard to do on an M1. Unfortunately debugging the lines one by one gives me the correct result for some reason.

**Comment by prashantgupta24** [comment]
Alright all tests pass when run against the new CB code with fms branch [paged_attn_mock](https://github.com/foundation-model-stack/foundation-model-stack/tree/paged_attn_mock)
```
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(1)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(1)] PASSED                                                                                                                      [  6%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(1)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(2)] PASSED                                                                                                                      [ 12%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(1)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(3)] PASSED                                                                                                                      [ 18%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(1)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(4)] PASSED                                                                                                                      [ 25%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(2)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(1)] PASSED                                                                                                                      [ 31%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(2)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(2)] PASSED                                                                                                                      [ 37%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(2)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(3)] PASSED                                                                                                                      [ 43%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(2)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(4)] PASSED                                                                                                                      [ 50%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(3)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(1)] PASSED                                                                                                                      [ 56%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(3)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(2)] PASSED                                                                                                                      [ 62%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(3)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(3)] PASSED                                                                                                                      [ 68%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(3)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(4)] PASSED                                                                                                                      [ 75%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(4)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(1)] PASSED                                                                                                                      [ 81%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(4)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(2)] PASSED                                                                                                                      [ 87%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(4)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(3)] PASSED                                                                                                                      [ 93%]
tests/e2e/test_spyre_cb.py::test_cb_handling[num_prompts(4)-v1-use_cb(1)-eager-llama-194m-max_num_seqs(4)] PASSED                                                                                                                      [100%]
```

File: tests/e2e/test_spyre_basic.py
Comment (prashantgupta24) [inline tests/e2e/test_spyre_basic.py:1]: This test has been moved to `test_spyre_cb` since it essentially tests the same things but both with CB and static.

**Comment by prashantgupta24** [comment]
> What is the status of this PR? Do we have blockers? These tests would be very helpful for the development of CB.

The tests on this PR work well they just need the custom branch from FMS

**Comment by wallashss** [comment]
> The tests on this PR work well they just need the custom branch from FMS

Yeah, you are right... These tests passing will break the others because of the divergence of FMS version. I had this problem before. 

I was wondering if we could merge this as-is and just skip them on CI for now, and we activate later again once we solve the FMS issue. I see value to have this test merged ASAP and I think we will be able to iterate the CB feature faster. We could make a follow up issue for this. What do you think?

Also cc @joerunde

**Comment by prashantgupta24** [comment]
> > The tests on this PR work well they just need the custom branch from FMS
> 
> Yeah, you are right... These tests passing will break the others because of the divergence of FMS version. I had this problem before.
> 
> I was wondering if we could merge this as-is and just skip them on CI for now, and we activate later again once we solve the FMS issue. I see value to have this test merged ASAP and I think we will be able to iterate the CB feature faster. We could make a follow up issue for this. What do you think?
> 
> Also cc @joerunde

Yeah I think we talked about this in one of the stand-ups. Worth trying it out I guess? Another option is to actually install the custom FMS branch in the CI when we actually run this test, although I'm not sure how tricky that will be. Honestly the tests are well contained so it's easy to copy them over whenever you want to test it out.

**Comment by wallashss** [comment]
> Another option is to actually install the custom FMS branch in the CI when we actually run this test, although I'm not sure how tricky that will be

Good idea! Could be, pretty easy !

```
pip install git+https://github.com/foundation-model-stack/foundation-model-stack.git@paged_attn_mock
```

I'm okay with moving forward like this.

**Comment by prashantgupta24** [comment]
> > Another option is to actually install the custom FMS branch in the CI when we actually run this test, although I'm not sure how tricky that will be
> 
> Good idea! Could be, pretty easy !
> 
> ```
> pip install git+https://github.com/foundation-model-stack/foundation-model-stack.git@paged_attn_mock
> ```
> 
> I'm okay with moving forward like this.

Yeah, sounds good! Probably better to open a separate story for that. Let me do that now.

**Comment by prashantgupta24** [comment]
Alright this PR installs the custom FMS branch! Let me fix up the tests next week then maybe we can merge it

File: tests/e2e/test_spyre_cb.py [hunk 6]
```diff
@@ -0,0 +1,238 @@
...
+    prompts: list[str],
+    # vllm_version: str,
+    monkeypatch: pytest.MonkeyPatch,
+):
+    """Test that the spyre worker correctly handles
+    continuous batches of requests that
+    finish after different numbers of forward passes"""
+
+    if cb and len(prompts) < max_num_seqs:
+        pytest.skip("Skipping since this doesn't really test CB")
```
prashantgupta24, line 58: Let me know if that doesn't make sense, I believe it could be an overkill to test these scenarios

.github/workflows/test-spyre.yml
```diff
@@ -55,6 +55,9 @@ jobs:
           - name: "utils"
             markers: "utils"
             flags: "--timeout=300"
+          - name: "cb"
+            markers: "cb"
+            flags: "--timeout=300 --forked"
```
wallashss, line 60: Do we know if there's order guaranteed here? I had experienced errors when switching back to the static batching with the fms branch installed.

prashantgupta24: It's an isolated job on a different runner so hopefully order shouldn't matter?

tests/e2e/test_spyre_cb.py
wallashss: IMHO It does make sense to remove the if and let it run. I found already cases that the data is not properly cleaned because it was expecting the engine has new requests process them. It might be interesting to find those cases if we do not properly set the state of system consistent. 

In fact, I was thinking about to duplicate most of the tests by turn on CB and expect the correct behavior of the system. However, due to the limitation of the branches of fms it will not be not so easy. I think we should consider this later.

prashantgupta24: > I was thinking about to duplicate most of the tests by turn on CB and expect the correct behavior of the system

Interesting point! Yeah we can definitely see about that later. For now I can revert the `if` statement

.github/workflows/test-spyre.yml
wallashss: > It's an isolated job on a different runner so hopefully order shouldn't matter?

If it is isolated env, then it's fine, the order does not matter. Tks.

wallashss (approved):
LGTM! Thanks a lot for this @prashantgupta24

PR #80, opened 2025-04-10
[draft v2] Add warmup performance metric gathering

pmur:
This adds support for capturing warmup metrics on each rank, and writing the results to the filesystem. Both are controlled via the following envvars:

	VLLM_SPYRE_PERF_METRIC_LOGGING_ENABLE
		0 (default) - no logging is performed
		1 - logging is captured to a rank specific file
                    in the directory VLLM_SPYRE_PERF_METRIC_LOGGING_DIR

	VLLM_SPYRE_PERF_METRIC_LOGGING_DIR
		/tmp (default)
		This is a path to an existing directory to write a rank
		specific file which captures warmup metrics in a csv
                like format. If the file exists, it is overwritten.

For #58


pmur:
@joerunde @tdoublep , second spin at this. Not quite as nice as prometheus, but still tenable.

RajalakshmiSR:
watching

Daniel-Schenker:
Commenting to receive updates.

vllm_spyre/perf_metrics.py
```diff
@@ -0,0 +1,59 @@
...
+class SpyrePerfMetricFileLogger(SpyrePerfMetricLoggerBase):
+    """ A per-rank file logging object """
+
+    def __init__(self, rank: int):
+        super().__init__(rank)
+        self.time_fmt = "%m-%d %H:%M:%S"
+        self.log_path = os.path.join(envs.VLLM_SPYRE_PERF_METRIC_LOGGING_DIR,
+                                     f"perf_log_rank_{str(rank)}.txt")
+        # Note, this file is closed during object deletion.
+        self.file = open(self.log_path, "w", buffering=1)  # noqa: SIM115
```
joerunde, line 39: Since this isn't something that we're writing a whole lot of info to (ie we're not putting one line in here per processed request) I think it's probably okay to open as needed with `"a"` mode to append each time, instead of keeping it open

joerunde:
@Daniel-Schenker I went ahead and made some small changes since today was @pmur's last day here (congrats and have fun at Red Hat!)

I tested manually, seems to work fine. I'll merge once CI passes

joerunde (approved):

PR #81, opened 2025-04-10
:bug: actually install everything

joerunde:
Our tests are passing because they're importing the source directly, the actual installed `vllm_spyre` package is empty 🙃

### DIFF
--- a/.github/workflows/test-spyre.yml
+++ b/.github/workflows/test-spyre.yml
@@ -81,8 +81,11 @@ jobs:
         run: |
           uv venv .venv --system-site-packages
           source .venv/bin/activate
+
+          # Syncs both the runtime and dev deps, based on the lockfile contents
+          uv sync --frozen
+          # Builds and installs the vllm_spyre wheel into .venv
           uv pip install -v .
-          uv sync --frozen --group dev
 
       - name: "Restore HF models cache"
         uses: actions/cache/restore@v4
--- a/.github/workflows/test-spyre.yml
+++ b/.github/workflows/test-spyre.yml
@@ -129,6 +129,7 @@ jobs:
           MASTER_ADDR: localhost
           DISTRIBUTED_STRATEGY_IGNORE_MODULES: WordEmbedding
         run: |
+          rm -fr vllm_spyre
           source .venv/bin/activate
           uv run pytest ${{ matrix.test_suite.flags }} \
             tests -v -k "${{ matrix.test_suite.tests }}"
--- a/.github/workflows/test-spyre.yml
+++ b/.github/workflows/test-spyre.yml
@@ -132,7 +132,9 @@ jobs:
           MASTER_ADDR: localhost
           DISTRIBUTED_STRATEGY_IGNORE_MODULES: WordEmbedding
         run: |
+          # Delete the source code so we can ensure we're testing the installed
+          # wheel
           rm -fr vllm_spyre
           source .venv/bin/activate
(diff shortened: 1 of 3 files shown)


joerunde:
Also the lockfile hadn't been updated so we were still testing vllm v0.8.0

.github/workflows/test-spyre.yml
```diff
@@ -129,6 +132,9 @@ jobs:
           MASTER_ADDR: localhost
           DISTRIBUTED_STRATEGY_IGNORE_MODULES: WordEmbedding
         run: |
+          # Delete the source code so we can ensure we're testing the installed
+          # wheel
+          rm -fr vllm_spyre
           source .venv/bin/activate
-          uv run pytest ${{ matrix.test_suite.flags }} \
+          python3 -m pytest ${{ matrix.test_suite.flags }} \
```
joerunde, line 139: `uv run` will re-sync by default which uninstalls the `vllm_spyre` wheel :(

We could do a `--no-sync` here, though I think the explicit `source .venv/bin/activate / python -m ...` is more familiar

.github/workflows/test-spyre.yml
```diff
@@ -81,8 +81,11 @@ jobs:
         run: |
           uv venv .venv --system-site-packages
           source .venv/bin/activate
+
+          # Syncs both the runtime and dev deps, based on the lockfile contents
+          uv sync --frozen
+          # Builds and installs the vllm_spyre wheel into .venv
```
joerunde, line 87: `uv sync` will uninstall the installed wheel and instead point back to the local source, so the pip install needs to come last here

uv.lock
```diff
@@ -3299,7 +3344,7 @@ dependencies = [
 [[package]]
 name = "vllm-spyre"
 version = "0.0.0"
-source = { virtual = "." }
+source = { editable = "." }
```
ckadner, line 3347: do you know how this was `virtual` before?

.github/workflows/test-spyre.yml
ckadner: 👍🏻

.github/workflows/test-spyre.yml
ckadner: makes sense, though I needed to read your comments 3 times and scroll through earlier build logs to grasp what's happening 🫤

pyproject.toml
```diff
@@ -21,7 +21,7 @@ spyre = "vllm_spyre:register"
 
 [tool.setuptools.packages.find]
 where = ["."]  # list of folders that contain the packages (["."] by default)
-include = ["vllm_spyre"]  # package names should match these glob patterns (["*"] by default)
+include = ["vllm_spyre*"]  # package names should match these glob patterns (["*"] by default)
```
ckadner, line 24: what are we gaining from the trailing asterisk?

ckadner (approved):
Looks good. Just 2 questions -- mostly for my learning ;-)

pyproject.toml
joerunde: This was the broken part- it was only pulling in files directly in the `vllm_spyre` directory, and nothing nested underneath

uv.lock
joerunde: 🤔  Nope!
I still have a lot more to learn here

.github/workflows/test-spyre.yml
joerunde: me too!

Issue #74, opened 2025-04-07
Implement the new warmup for continuous batching


joerunde:
Followup for continuous batching: Implement the warmup

yannicks1:
will be merged with PR #82

Issue #73, opened 2025-04-07
Use real fms api instead of mock one


joerunde:
Followup for continuous batching

Use the real implementation of the model with paged attention instead of the wrapper

yannicks1:
initial implementation working on branch [dev-updated-fms-api ](https://github.com/vllm-project/vllm-spyre/tree/dev-updated-fms-api). Needs to be consumed with fms branch [paged_attn_mock](https://github.com/foundation-model-stack/foundation-model-stack/tree/paged_attn_mock)

yannicks1:
closed as part of #82

PR #82, opened 2025-04-11
Paged attention/ new fms API 

yannicks1:
### Introducing paging along prompt dimension and adhering to new fms API

closing  issues 
- #73 
- #74 
- #99 
- #97 
- #98

Note: needs to be consumed with fms branch [paged_attn_mock](https://github.com/foundation-model-stack/foundation-model-stack/tree/paged_attn_mock) for continuous batching (`VLLM_SPYRE_USE_CB` = 1), and fms `main` branch for static batching (`VLLM_SPYRE_USE_CB` = 0). Of course, once `paged_attn_mock` is merged to fms `main`, both will consume fms `main`.


vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -241,6 +242,105 @@ def load_model(self):
...
+                total_num_scheduled_tokens=prompt_len,
+                scheduled_spec_decode_tokens={},
+                scheduled_encoder_inputs={},
+                num_common_prefix_blocks=0,
+                finished_req_ids=set(),
+                free_encoder_input_ids=[],
+                structured_output_request_ids={},
+                grammar_bitmask=None,
+            )
+            logger.info("Warmup prefil %d/2...", i + 1)
```
joerunde, line 296:
```suggestion
            logger.info("Warmup prefil %d/%d...", i + 1, batch_size)
```

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -241,6 +242,105 @@ def load_model(self):
...
+            self.execute_model(scheduler_output)
+
+        # one decode iteration across both sequences
```
joerunde, line 299: Nice, do we no longer need to do a full sequence length's worth of decode iterations?

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -241,6 +242,105 @@ def load_model(self):
...
+        ]
+
+        scheduler_output = SchedulerOutput(
+            scheduled_new_reqs=[],
+            scheduled_cached_reqs=cached_requests,
+            num_scheduled_tokens={
+                'warmup-0': 1,
+                'warmup-1': 1
+            },
+            total_num_scheduled_tokens=2,
```
joerunde, line 320: Should these both be parameterized on batch size? 
```
num_scheduled_tokens={f"warmup-{i}": 1 for i in range(batch_size)},
total_num_scheduled_tokens=batch_size
```

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -241,6 +242,105 @@ def load_model(self):
...
+        )
+        logger.info("Warmup decode 1/1...")
+        self.execute_model(scheduler_output)
+
+        # free blocks
+        for req in dummy_requests:
+            logger.debug("Freeing request id: %s", req.req_id)
+            for freed_block in self.model_runner.req_ids2blocks[req.req_id]:
+                self.model_runner.free_blocks.append(freed_block)
+            del self.model_runner.req_ids2blocks[req.req_id]
```
joerunde, line 337: maybe we could schedule another pass where all the requests are marked as finished so that the model runner can use its own cleanup logic

vllm_spyre/v1/worker/spyre_worker.py
yannicks1: No, according to the compiler team this is enough now:)

vllm_spyre/v1/worker/spyre_worker.py
yannicks1: good point, did that.

vllm_spyre/v1/worker/spyre_worker.py
yannicks1: we actually used this exact approach but then changed it to the above as it is much less lines of code and does not require and additional code in the model runner. see [here](https://github.com/vllm-project/vllm-spyre/pull/83#discussion_r2039553645)

joerunde: anyway- we can definitely fix that later if it becomes an issue

joerunde (approved):
Haven't wrapped my head around all the right-padding stuff, bet let's get this into main so we don't have to keep another long-lived dev branch around

sducouedic (commented):
I am not sure if it is best to do it in this PR as it is not related to CB, but can we merge this into one line? `logits, self.past_key_value_states = output`. [LINK](https://github.com/vllm-project/vllm-spyre/blob/d1677c9038c939131ff056f058a1c4508cbae392/vllm_spyre/model_executor/model_loader/spyre.py#L387-L388).

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -272,11 +289,23 @@ def __init__(
             print(f"[SpyreCausalLM] model type {self.config.model_type} "
                   f"not supported in ContinuousBatchingFmsModel")
```
sducouedic, line 291: Should we raise an Error here?

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -272,11 +289,23 @@ def __init__(
...
-        # (layers)x(k,v)x[max_batch, num_kv_heads, max_model_len, head_dim]
-        self.fms_kv_cache: list[tuple[torch.Tensor, torch.Tensor]] = [
-            (torch.empty((max_batch, num_kv_heads, max_model_len, head_dim)),
-             torch.empty((max_batch, num_kv_heads, max_model_len, head_dim)))
-            for i in range(num_layers)
+        BLOCK_SIZE = 64
+        NUM_BLOCKS = max_batch * max_model_len // BLOCK_SIZE  # 64
+
+        # List[layers] of Tuple[k,v] of
+        # Tensor[NUM_BLOCKS, BLOCK_SIZE, num_kv_heads, max_model_len, head_dim]
```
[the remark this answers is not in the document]
yannicks1: it's used indirectly via `NUM_BLOCKS`.

tjohnson31415:
@yannicks1 I created a PR into this branch to fix a crash I found:
https://github.com/vllm-project/vllm-spyre/pull/112

vllm_spyre/v1/worker/spyre_worker.py
yannicks1: I do agree, but let's deal with that in a separate PR. We are very close to merging this one:)

vllm_spyre/v1/core/scheduler.py
```diff
@@ -268,7 +268,7 @@ def schedule(self) -> "SchedulerOutput":
             self.last_running = self.total_running
             self.running = []
             logger.debug(
-                "Scheduling a prompt step of %d requests, holding back %d "
+                "Scheduling a prefil step of %d requests, holding back %d "
```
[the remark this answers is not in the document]
yannicks1: Thanks for pointing this out, it was actually not the only occasion where I wrote prefill wrong :)

vllm_spyre/model_executor/model_loader/spyre.py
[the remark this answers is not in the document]
sducouedic: ah okay, it looked like these were used to describe the tensor shape, because of the order of the values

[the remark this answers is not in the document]
sducouedic: sorry, I don't mean to block this PR, but what does this line of comment mean exactly if it is not the shape of the tensors right below?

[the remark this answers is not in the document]
yannicks1: Ahh, only now I see: it is about the comment, not the actual code. Good catch, updated it [here](https://github.com/vllm-project/vllm-spyre/pull/82/commits/8af639ce27239bc8edd8b392ae027186683ceec8).

sducouedic (approved):
LGTM

PR #87, opened 2025-04-15
[tests] Fix tests against vllm/main

rafvasq:
- Never install `triton` to fix tests against vllm/main (e.g. [error](https://github.com/vllm-project/vllm-spyre/actions/runs/14474659595/job/40597229196#step:10:77)) specifically caused by version `3.2.0` 
  - Could also be fixed by updating to torch `2.6.0+cpu` to match vllm

- Use `cpu` as device name to work around changes introduced in https://github.com/vllm-project/vllm/pull/16464/files


vllm_spyre/platform.py
```diff
@@ -19,13 +19,13 @@
...
     _enum = PlatformEnum.OOT
-    device_name: str = "spyre"
+    device_name: str = "cpu"
     device_type: str = "cpu"
     supported_quantization: list[str] = ["gptq"]
 
     @classmethod
     def get_device_name(cls, device_id: int = 0) -> str:
-        return "spyre"
+        return "cpu"
```
joerunde, line 28: I guess technically returning `spyre` here is still safe since the offending vllm code uses `current_platform.device_name` directly.

I think this method is meant for returning the human readable device name, based on this code snippet:
```
        try:
            from vllm.platforms import current_platform
            device_name = current_platform.get_device_name().lower()
        except Exception:
            # This is only used to set default_max_num_batched_tokens
            device_name = "no-device"

        if "h100" in device_name or "h200" in device_name:
```

Which is for sure a bit confusing but if that's the case then I think we should keep `spyre` here for the humans

vllm_spyre/platform.py
```diff
@@ -19,13 +19,13 @@
 
 class SpyrePlatform(Platform):
     _enum = PlatformEnum.OOT
-    device_name: str = "spyre"
+    device_name: str = "cpu"
```
joerunde, line 22: Oh let's also put a comment about this hack here too for future readers

rafvasq: Done!

joerunde (approved):
nice

PR #90, opened 2025-04-16
Add `requires-python` and update type hints

joerunde:
This PR adds `requires-python` to the project table in our pyproject.toml. This tells installers what versions of python we support, which is good practice to include when publishing a package. Currently python 3.9 is the oldest version of python that has official support, it will reach end of life this October: https://devguide.python.org/versions/. This plugin shouldn't have any code that specifically prohibits installing with future python versions, so this sets the valid python version range to 3.9+.

As of python 3.9, type hinting with primitive container types (list, dict, tuple) became supported and type hinting with the types from the `typing` package was deprecated. This causes `ruff` to now fail, as we won't support any versions of python that require the older style type hints. So this PR also updates all the type hints, this was done with `ruff check . --fix --unsafe-fixes`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,7 @@ dependencies = [
     "ibm-fms==0.0.8",
     "vllm",
 ]
+requires-python = ">=3.9"
 
 [project.entry-points."vllm.platform_plugins"]
 spyre = "vllm_spyre:register"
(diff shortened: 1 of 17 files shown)


tjohnson31415 (approved):
Only typing changes + requires-python. LGTM!

PR #95, opened 2025-04-17
Right padding of prompts

yannicks1:
### Right padding of prompts

Ensures that prefills (batch size 1) are always of length N*64 (right padding) while the actual prompt tokens still align with the tkv of the current decoding batch (left padding). This a requirement from the compiler.


tdoublep (approved):
LGTM - great work

PR #102, opened 2025-04-21
♻️ use pathlib for constructing model path

prashantgupta24:
This is more pythonic + it allows the user to select an empty dir for `models`


joerunde (approved):
lpgtm!

PR #103, opened 2025-04-22
[CB] Fix Tensor Parallelism Error

### LINKED ISSUE #99: [CB] Fix tensor parallelism error for TP>1 with the new FMS API
With the new FMS API for continuous batching, inference fails when TP is greater than 1.

sducouedic:
This fixes tensor parallelism error when using the new FMS API by adapting the tensor shape of `self.past_key_value_states` using the built-in function `get_num_kv_heads` of vLLM. 
* Closes #99


yannicks1 (approved):
LGTM. Tested for a couple of models on CPU with TP 2 and 4 and all combinations worked.

PR #104, opened 2025-04-22
:bug: Fix static batching with multiple shapes

### LINKED ISSUE #20: [V1] Enable multiple warmup shapes
Currently the V1 integration will fail when multiple warmup sizes are set with a graph compilation failure. The failure looks very similar to the failure that occurs if you accidentally enable `torch.inference_mode` somewhere. An easy test to trigger this is https://github.com/vllm-project/vllm-spyre/blob/main/tests/test_spyre_warmup_shapes.py

Maybe related to https://github.com/vllm-project/vllm-spyre/issues/12

joerunde:
Closes #20 

This pulls over some code from https://github.com/foundation-model-stack/aiu-fms-testing-utils/pull/13 that properly marks tensors as static/dynamic for prefill and decode when doing static batching. I've tested on a spyre card and confirmed that I can compile multiple warmup shapes (with batch sizes > 1) and run batches of requests

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -105,6 +105,9 @@ class SpyreModelRunner(ModelRunnerBase[ModelInputForSpyre]):
         # Lazy initialization: after load_model.
         self.model: nn.Module
 
+        # Flag to be turned off after warmup is complete
+        self.warmup_mode = True
+
     def get_model(self) -> nn.Module:
         return self.model
 
@@ -210,6 +213,10 @@ class SpyreModelRunner(ModelRunnerBase[ModelInputForSpyre]):
                                       dtype=torch.float16,
                                       use_mla=False)
         return {"foo": attn_spec}
+
+    def complete_warmup(self):
+        """Turn off warmup mode once the warmup is complete"""
+        self.warmup_mode = False
 
 
 class StaticBatchingSpyreModelRunner(SpyreModelRunner):
@@ -539,6 +546,10 @@ class StaticBatchingSpyreModelRunner(SpyreModelRunner):
         """Yoinked from 
         https://github.com/foundation-model-stack/aiu-fms-testing-utils/pull/13
         """
+        if not self.warmup_mode:
+            # Only mark tensors when we're warming up and compiling the graphs
+            return
+
         # To produce like graphs during pre-fill, we mark the prefill
         # batch x seq as static, but relax this for decode for the seq
         if model_input.is_prompt:
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -423,6 +423,7 @@ class StaticBatchingSpyreModelRunner(SpyreModelRunner):
         self._update_states(scheduler_output)
 
         model_input = self.prepare_model_input(scheduler_output)
+        self._mark_input_tensors(model_input)
 
         # TODO(Wallas): I think it would be better move the indices as argument
         # of the forward rather than set as an attribute of the model. I'm not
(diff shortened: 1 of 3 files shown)


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -534,6 +535,32 @@ def _get_padded_batch_size(self, new_requests: list[NewRequestData]):
...
+            # we always want prefill to be static to produce same-like graph
+            torch._dynamo.mark_static(model_input.input_tokens, 0)
+            torch._dynamo.mark_static(model_input.input_tokens, 1)
+            torch._dynamo.mark_static(model_input.input_masks, 0)
+            torch._dynamo.mark_static(model_input.input_masks, 1)
+            torch._dynamo.mark_static(model_input.input_masks, 2)
+            torch._dynamo.mark_static(model_input.input_positions, 0)
+            torch._dynamo.mark_static(model_input.input_positions, 1)
+        else:
+            # we always want the decode to be dynamic on sequence
```
tjohnson31415, line 554: The FMS code has a check on `if torch._dynamo.config.dynamic_shapes:` before using `mark_dynamic`.  It looks like PyTorch will raise an exception if using `mark_dyamic` with `dynamic_shapes == False` ([REF](https://github.com/pytorch/pytorch/blob/66a76516bfc341b2b55bb2056d2faa9c2de46d69/torch/_dynamo/decorators.py#L186-L187)).

That might be fine, but wanted to point it out.

tjohnson31415 (approved):
There is an additional `if` check in FMS, but otherwise the code looks as yoinked. If it is ok to force dynamic shapes, then LGTM.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -423,6 +423,7 @@ def execute_model(
         self._update_states(scheduler_output)
 
         model_input = self.prepare_model_input(scheduler_output)
+        self._mark_input_tensors(model_input)
```
tjohnson31415, line 426: It looks like we'll be calling this for every model execution, but I think the dynamic/static settings are only for configuring the `torch.compile` for the model. It would be better if `_mark_input_tensors` only happened for the warmup.

vllm_spyre/v1/worker/spyre_model_runner.py
joerunde: Ah yeah, it was my understanding that we're always using dynamic shapes, and I didn't see the CB code making that check either. I can read back on some issue comments to see if that assumption is true 🤔

vllm_spyre/v1/worker/spyre_model_runner.py
joerunde: Ah, gotcha. I was also following the CB code block here and doing it always, I can try to pipe in some flag to see if we're in warmup mode to not do it on every forward pass and see what happens.

I'm not 100% sure though if this is supposed to be only at warmup time or if it is supposed to be on every forward pass

joerunde:
Ah, I also had to remove this test from the `V0` suite because I didn't backport the fix to the v0 model runner. I'm not gonna do that since we're focusing on V1.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -534,6 +542,36 @@ def _get_padded_batch_size(self, new_requests: list[NewRequestData]):
...
+            torch._dynamo.mark_static(model_input.input_tokens, 1)
+            torch._dynamo.mark_static(model_input.input_masks, 0)
+            torch._dynamo.mark_static(model_input.input_masks, 1)
+            torch._dynamo.mark_static(model_input.input_masks, 2)
+            torch._dynamo.mark_static(model_input.input_positions, 0)
+            torch._dynamo.mark_static(model_input.input_positions, 1)
+        else:
+            # we always want the decode to be dynamic on sequence
+            torch._dynamo.mark_dynamic(model_input.input_tokens, 1)
+            torch._dynamo.mark_dynamic(model_input.input_masks, 1)
```
yannicks1, line 567: aren't these dimension 1 all the time? I thought the first dimension of the mask is the only dynamic thing here as we only feed the new token (KV caching the previous tokens)... @joerunde

joerunde: 🤔 hmmm, good point. I had definitely copy-pasted without thinking too much about these. Let me investigate a little bit

Issue #71, opened 2025-04-04
[V1] remove ModelInputForSpyre class


yannicks1:
### [V1] remove ModelInputForSpyre class

There is no `ModelRunnerInput` classes (inheriting from `ModelRunnerInputBase`) in upstream V1. Thus suggesting to remove `ModelInputForSpyre` as this was merely used for V0 to V1 migration.

Issue #76, opened 2025-04-07
Cleaner integration of continuous batching


joerunde:
Followup for continuous batching work

Integrate better with vLLM data structures (e.g. InputBatch)

PR #107, opened 2025-04-23
Refactor ModelInputForSpyre dataclass

prashantgupta24:
Fix https://github.com/vllm-project/vllm-spyre/issues/71


prashantgupta24:
I wish there was a simple way to revert all the lint fixes that happened to this file

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -40,60 +32,33 @@
 
 logger = init_logger(__name__)
 
-TModelInputForSpyre = TypeVar('TModelInputForSpyre',
-                              bound="ModelInputForSpyre")
-
-
-@dataclass(frozen=True)
-class ModelInputForSpyre(ModelRunnerInputBase):
```
joerunde, line 48: Should we keep this as a slimmer dataclass with just the things we need in it? That way we can at least keep one container as the return from `prepare_model_input` instead of returning 5 things:
```
        input_tokens, input_positions, input_masks, \
            sampling_metadata, is_prompt = (
            self.prepare_model_input(scheduler_output))
``` 

We can nix the unused `virtual_engine` and the `pooling_metadata` since this won't run embeddings, along with the broadcasting functions that were used in v0

joerunde:
Just one queston on whether we should keep a container here anyway- otherwise thanks for hunting down all the unused code bits here!

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24: Yeah I thought about that, the only reason I went against that decision was because of naming conventions - calling it something like `ModelInput` might confuse folks. Perhaps a unique disambiguous name would be helpful

prashantgupta24: Gave it a shot, let me know if it looks better!

wallashss (approved):
Thanks @prashantgupta24 for this!

LGTM, if @joerunde does not have any comments, I think it's safe to merge and I can keep working on this code.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -393,16 +371,16 @@ def prepare_model_input(
                     self.input_batch.soft_remove_request(req_id)
                 self.input_batch.refresh_sampling_metadata()
 
-            (input_tokens, input_positions, input_masks) = \
-                self._prepare_decode(scheduler_output.scheduled_cached_reqs)
-
-        sampling_metadata = self.input_batch.sampling_metadata
+            (input_tokens, input_positions,
+             input_masks) = self._prepare_decode(
```
joerunde, line 375: If we wanted to push the refactor just a tiny bit further, both `_prepare_prompt` and `_prepare_decode` can now return a valid `ModelForwardInputs`

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -775,14 +730,15 @@ def execute_model(
         # Sample the next token.
         output: SamplerOutput = self.model.sample(
             logits=logits,
-            sampling_metadata=model_input.sampling_metadata,
+            # TODO: Uncomment once Wallas is done with the work
+            # sampling_metadata=self.input_batch.sampling_metadata,
```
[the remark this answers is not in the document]
joerunde: it's #76 if you wanna link an issue, but I think this is fine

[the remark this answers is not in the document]
joerunde: Does this still work without passing `sampling_metadata` at all?

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24: `_prepare_prompt` returns `seq_lens` as well, not sure if we want to add that as a field within the `ModelForwardInputs`

prashantgupta24: Although I don't see `seq_lens` being used anywhere in CB, probably safe to delete it?

joerunde: probably! I assume that's a holdover from the V0 times

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -40,12 +33,9 @@
 
 logger = init_logger(__name__)
 
-TModelInputForSpyre = TypeVar('TModelInputForSpyre',
-                              bound="ModelInputForSpyre")
-
 
-@dataclass(frozen=True)
-class ModelInputForSpyre(ModelRunnerInputBase):
+@dataclass()
```
prashantgupta24, line 37: `frozen` can be brought back once `sampling_metadata` is removed

vllm_spyre/v1/worker/spyre_model_runner.py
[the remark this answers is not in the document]
prashantgupta24: Nope, had to bring it back

joerunde (approved):
lpgtm!

PR #109, opened 2025-04-24
:fire: remove v0 from static batching test

joerunde:
We don't have v0 support for multiple warmup shapes on spyre cards, so this removes v0 from the static batching tests


wallashss (approved):
Perfect! Thanks!

PR #110, opened 2025-04-24
:zap: consolidate tensor marking for static batch runner

joerunde:
Two small changes:
- @yannicks1 is right that `model_input.input_masks, 1` is always gonna be size 1, so we don't have to mark it dynamic in the decode passes on warmup
- Marking dimension 2 of each `past_key_value_states` tensor as dynamic only needs to be done at warmup time, we don't need to do it on every forward pass

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -564,13 +564,14 @@ class StaticBatchingSpyreModelRunner(SpyreModelRunner):
         else:
             # we always want the decode to be dynamic on sequence
             torch._dynamo.mark_dynamic(model_input.input_tokens, 1)
-            torch._dynamo.mark_dynamic(model_input.input_masks, 1)
             torch._dynamo.mark_dynamic(model_input.input_masks, 2)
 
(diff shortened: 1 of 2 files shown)


joerunde:
Running tests for this manually on a spyre card now to ensure this doesn't break

tjohnson31415 (approved):
Code changes LGTM.

joerunde:
Internal tests look good as well 😎

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -564,13 +564,14 @@ def _mark_input_tensors(self, model_input: ModelInputForSpyre) -> None:
         else:
             # we always want the decode to be dynamic on sequence
             torch._dynamo.mark_dynamic(model_input.input_tokens, 1)
```
yannicks1, line 566: Ain't that dimension also 1 all the time? My previous [comment](https://github.com/vllm-project/vllm-spyre/pull/104#discussion_r2058423925) was actually referring to both lines:)

joerunde: Oh lol

PR #111, opened 2025-04-25
[V1] Optimizing the static batching scheduler 

yannicks1:
### Optimizing the static batching scheduler in V1

Introducing a skip queue to fill up batches more optimally. 

Idea: move past request that do not fit the current batch warmup shapes. 
Previously the batch was scheduled when reaching a request that cannot be part of the current batch. Unused slots in a batch were simply wasted/padded. With this change we try to still fill those slots with subsequent request. Note that priority is preserved: e.g. the skipped requests will come first in the next batch.


joerunde (approved):
LGTM, thanks!

PR #112, opened 2025-04-25
fix/refactor: remove last_running and total_running

tjohnson31415:
These changes from testing request cancellation with the continuous batching branch. In some cases, the cancellation would result in a crash:
```
ERROR 04-25 16:39:23 [core.py:387] EngineCore hit an exception: Traceback (most recent call last):
ERROR 04-25 16:39:23 [core.py:387]   File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/engine/core.py", line 380, in run_engine_core
ERROR 04-25 16:39:23 [core.py:387]     engine_core.run_busy_loop()
ERROR 04-25 16:39:23 [core.py:387]   File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/engine/core.py", line 400, in run_busy_loop
ERROR 04-25 16:39:23 [core.py:387]     self._process_input_queue()
ERROR 04-25 16:39:23 [core.py:387]   File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/engine/core.py", line 425, in _process_input_queue
ERROR 04-25 16:39:23 [core.py:387]     self._handle_client_request(*req)
ERROR 04-25 16:39:23 [core.py:387]   File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/engine/core.py", line 443, in _handle_client_request
ERROR 04-25 16:39:23 [core.py:387]     self.abort_requests(request)
ERROR 04-25 16:39:23 [core.py:387]   File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/engine/core.py", line 193, in abort_requests
ERROR 04-25 16:39:23 [core.py:387]     self.scheduler.finish_requests(request_ids,
ERROR 04-25 16:39:23 [core.py:387]   File "/home/senuser/my-vllm/lib64/python3.11/site-packages/vllm/v1/core/sched/scheduler.py", line 724, in finish_requests
ERROR 04-25 16:39:23 [core.py:387]     self.running.remove(request)
ERROR 04-25 16:39:23 [core.py:387] ValueError: list.remove(x): x not in list
ERROR 04-25 16:39:23 [core.py:387] 
CRITICAL 04-25 16:39:23 [core_client.py:359] Got fatal signal from worker processes, shutting down. See stack trace above for root cause issue.
Killed
```

This is due to state being moved to `self.last_running` and `self.total_running` causing `self.running` to be empty even when requests are Running.
 
This PR makes some changes similar to what we did with the holdback_queue where state is hidden from the base scheduler but restored after the call to `schedule()`. This way, we can avoid specializing `finish_requests`

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -231,9 +231,6 @@ class ContinuousBatchingSpyreScheduler(SpyreScheduler):
     def __init__(self, *args, **kwargs) -> None:
         # Initialize SpyreScheduler
         super().__init__(*args, **kwargs)
-        # running queue of last decoding step
-        self.last_running: list[Request] = []
-        self.total_running: list[Request] = []
 
     def add_request(self, request: Request) -> None:
         """This override rejects requests that exceed max context length"""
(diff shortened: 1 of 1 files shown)


yannicks1:
thanks, I will put @nikolaospapandreou in the loop here as he is the original author of the continuous batching scheduler:)

yannicks1 (approved):
very elegant implementation! tested it with a couple of examples and all works fine!

PR #113, opened 2025-04-25
:fire: remove dynamic mark on input_tokens[1]

joerunde:
Following up on https://github.com/vllm-project/vllm-spyre/pull/110#discussion_r2059737772 because sometimes my reading comprehension is less than stellar


joerunde:
Tests pass on spyre cards ✅

tjohnson31415 (approved):
Maybe I should actually test something instead of just rubber stamping all of these changes....

joerunde:
> Maybe I should actually test something instead of just rubber stamping all of these changes.... but not this time, LGTM

🙈

Issue #94, opened 2025-04-16
Dig into config issues when setting vllm options like --max-num-seqs


joerunde:
With the continuous batching support, we noticed that the `VllmConfig` objects that are created are different in the main process and the worker processes: https://github.com/vllm-project/vllm-spyre/pull/66#discussion_r2034075804

This caused us to introduce new environment variables instead of using the existing vllm args, because we needed to ensure that the batch size and context length were accurate everywhere. We should look into this and understand why there are differences in the config objects, and fix any bugs upstream so that we can accept the usual vllm args.

prashantgupta24:
Sus behavior - 
```
WARNING 04-24 20:03:37 [config.py:3799] Current vLLM config is not set.
```

This line re-triggers `VllmConfig()` which defaults all values. Will keep investigating...

PR #114, opened 2025-04-25
[CB] 🔥 remove cb env vars

prashantgupta24:
Fix https://github.com/vllm-project/vllm-spyre/issues/94


prashantgupta24:
The only place left to remove the env vars are from `validate_request` within `platform.py`

vllm_spyre/platform.py
```diff
@@ -77,26 +78,28 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+        #         # the max batch size respecting AIU Spyre KV cache size
+        #         scheduler_config.max_num_seqs =\
+        #             envs_spyre.VLLM_SPYRE_MAX_BATCH_SIZE
+        #         # ToDo: this function check_and_update_config is called twice:
+        #         # 1st time scheduler_config.max_num_seqs is what user sets
+        #         # 2nd time scheduler_config.max_num_seqs is 128
+        #     else:
+        #         # The v0 scheduler will run out of
+        #           blocks if this is overridden
+        #         scheduler_config.max_num_seqs = max_batch_size
```
prashantgupta24, line 102: not 100% sure we need this?
```
        #         scheduler_config.max_num_seqs = max_batch_size
```

vllm_spyre/platform.py
```diff
@@ -133,6 +136,13 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 scheduler_config.max_num_seqs, cache_config.block_size,
                 cache_config.num_gpu_blocks_override)
 
+    @classmethod
+    def use_all_gather(cls) -> bool:
+        """
+        Whether to use allgather in LogitsProcessor to gather the logits.
+        """
+        return True
+
```
prashantgupta24, line 145: This was the reason that there were two calls to `check_and_update_config` in `platform.py ` - since this function did not exist, the `interface.py`'s `use_all_gather` was being called which defaulted all the config values

prashantgupta24: Do we need to add a condition here like in `interaface.py`?

wallashss: I guess not, IIUC, when we use V1, it should use `use_all_gather`

vllm_spyre/platform.py
joerunde: For the static batching use case the max batch size is still set with `VLLM_SPYRE_WARMUP_BATCH_SIZES`, so we override it to make sure the scheduling config is consistent. (e.g. setting both `--max-num-seqs=1` and `VLLM_SPYRE_WARMUP_BATCH_SIZES=2,4,6` wouldn't work out too well.)

The model and cache configs are also updated in a similar fashion below

joerunde:
@prashantgupta24 There's an if check in here for `if cache_config and model_config:`, previously I had seen that on some invocations of `check_and_update_config` those were `None`. Can you check if that's still the cace after fixing this issue with `use_all_gather`?

Also cc @yannicks1, I think we may want to actually leave the environment variables in place for a bit as it sounds like they're currently used by the torch stack as well

examples/offline_inference_spyre_cb_test.py
```diff
@@ -16,9 +16,6 @@
 os.environ['VLLM_SPYRE_USE_CB'] = '1'
 os.environ['VLLM_USE_V1'] = '1'
 
-os.environ['VLLM_SPYRE_MAX_CONTEXT_LENGTH'] = '2048'
-os.environ['VLLM_SPYRE_MAX_BATCH_SIZE'] = str(max_num_seqs)
```
prashantgupta24, line 20: Do we want to leave these in since pytorch implementation might need them?

vllm_spyre/platform.py
```diff
@@ -43,6 +44,7 @@ def is_async_output_supported(cls, enforce_eager: Optional[bool]) -> bool:
 
     @classmethod
     def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
+        cls._config = vllm_config
```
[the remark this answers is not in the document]
prashantgupta24: Should I change it at the API layer itself? I can imagine validating requests might need config params in the future

examples/offline_inference_spyre_cb_test.py
yannicks1: yes, let's keep setting them here for the time being just to be on the safe side. cc @joerunde

joerunde:
> @joerunde I see both not none when I put print statements, so that should solve it

@prashantgupta24 would you mind removing the checks on cache and model configs then, and migrating the logic from that section to where it logically belongs above?

vllm_spyre/platform.py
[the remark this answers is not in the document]
joerunde: I think `get_current_vllm_config` will only work post-engine-bootup, and `check_and_update_config` is called during engine creation. I'm curious, does it also not work if you call `get_current_vllm_config` inside of `validate_request`?

I'm fine with keeping a class cache around like this though

prashantgupta24:
> I think we should go one of two directions with this:
> 
> 1. Delete the environment variables from `envs.py` completely, and inside `check_and_update_config` set them back in `os.environ` to the values that are in vllm_config. That way the single source of truth is the CLI args, and we control setting the environment up for `torch_sendnn` to work properly.
> 2. Keep the environment variables in `envs.py`, and override the `vllm_config` with their values inside `check_and_update_config`. That way the environment variables remain as the source of truth, but the rest of the code can still reference the config object instead of the environment variables.
> 
> I'd prefer going with option (1) so that our continuous batching is configured the same way as upstream vllm when running on gpus.

Yep option 1 is where I was going but then you mentioned that torch might still need them so I thought I can't do that. 

> Control setting the environment up for `torch_sendnn` to work properly

I'm not sure I fully understand how we would do that, can you elaborate?

vllm_spyre/platform.py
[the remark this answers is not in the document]
prashantgupta24: > does it also not work if you call `get_current_vllm_config` inside of `validate_request`?

Nope :(

prashantgupta24:
> > @joerunde I see both not none when I put print statements, so that should solve it
> 
> @prashantgupta24 would you mind removing the checks on cache and model configs then, and migrating the logic from that section to where it logically belongs above?

Let me clean up the whole logic block then, we use `v1` and `cb` far too many times within the function

joerunde:
> Yep option 1 is where I was going but then you mentioned that torch might still need them so I thought I can't do that.

> > Control setting the environment up for torch_sendnn to work properly

> I'm not sure I fully understand how we would do that, can you elaborate? Is it by set them back in os.environ to the values that are in vllm_config? Basically torch just needs the 2 env vars? Can it not read vllm_config?

well, there's no api currently to pass a `vllm_config` down to `torch_sendnn`, my simplified view of things is that it's basically just a compiler backend for `torch`- it operates on tensors and model graphs. My understanding is that it needs these two environment variables to be set so that it knows what shapes it needs to compile.

So something like this would set those environment variables for torch to use, without requiring the end user to do it:
```
def check_and_update_config(cls, vllm_config):
    os.environ["VLLM_SPYRE_MAX_CONTEXT_LENGTH"] = vllm_config.model_config.max_model_len
    os.environ["VLLM_SPYRE_MAX_BATCH_SIZE"] = vllm_config.scheduler_config.max_num_seqs
```
but, we'd have to make sure that those get set for all vllm processes so it may not be _quite_ that simple

examples/offline_inference_spyre_cb_test.py
prashantgupta24: Removing them from here since we set them in the `check_and_update_config` function now

vllm_spyre/platform.py
```diff
@@ -168,10 +172,9 @@ def get_warmup_shapes(cls, scheduler_config) -> tuple[dict[str, int], ...]:
                     "The lists in VLLM_SPYRE_WARMUP_PROMPT_LENS and "
                     "VLLM_SPYRE_WARMUP_NEW_TOKENS must have equal length")
 
-        if not envs_spyre.VLLM_SPYRE_USE_CB:
```
prashantgupta24, line 171: duplicated check, the function is called before checking `if not envs_spyre.VLLM_SPYRE_USE_CB`

yannicks1:
Yes, option (1) is the way to go here! these env variables have to be set for the AIU compiler team, but should not be consumed in vllm-spyre code. Please have a look at [this](https://github.com/vllm-project/vllm-spyre/pull/136) PR, we have agreed on another name. Thus also please remove it from `offline_inference_spyre_cb_test.py`

prashantgupta24:
> Yes, option (1) is the way to go here! these env variables have to be set for the AIU compiler team, but should not be consumed in vllm-spyre code. Please have a look at [this](https://github.com/vllm-project/vllm-spyre/pull/136) PR, we have agreed on another name. Thus also please remove it from `offline_inference_spyre_cb_test.py` @joerunde @prashantgupta24

Done!

prashantgupta24:
> sorry I accidentally updated your branch @prashantgupta24 , thought I was on another PR... :/

No problem!

vllm_spyre/platform.py
```diff
@@ -43,97 +45,97 @@ def is_async_output_supported(cls, enforce_eager: Optional[bool]) -> bool:
...
 
         if parallel_config.worker_cls == "auto":
-            if envs.VLLM_USE_V1:
-                parallel_config.worker_cls = \
-                    "vllm_spyre.v1.worker.spyre_worker.SpyreWorker"
-            else:
-                parallel_config.worker_cls = \
-                    "vllm_spyre.worker.spyre_worker.SpyreWorker"
+            parallel_config.worker_cls = (
+                f'vllm_spyre{".v1" if envs.VLLM_USE_V1 else ""}'\
```
yannicks1, line 62: nice hack:)

vllm_spyre/platform.py
```diff
@@ -43,97 +45,97 @@ def is_async_output_supported(cls, enforce_eager: Optional[bool]) -> bool:
...
-                    "vllm_spyre.v1.core.scheduler.ContinuousBatchingSpyreScheduler"
-            else:
-                scheduler_config.scheduler_cls = \
-                    "vllm_spyre.v1.core.scheduler.StaticBatchingSpyreScheduler"
-        else:
-            scheduler_config.scheduler_cls = \
-                "vllm_spyre.core.scheduler.SpyreScheduler"
-
-        if not envs_spyre.VLLM_SPYRE_USE_CB:
+        # continuous batching related checks
```
yannicks1, line 65: a bit confusing here as the lines after `# continuous batching related checks` are the static batching checks

vllm_spyre/platform.py
```diff
@@ -43,97 +45,97 @@ def is_async_output_supported(cls, enforce_eager: Optional[bool]) -> bool:
...
+            cache_config.num_gpu_blocks_override,
+        )
+
+        # set env vars for torch_sendnn to consume
+        os.environ["VLLM_DT_MAX_CONTEXT_LEN"] = str(
+            vllm_config.model_config.max_model_len)
+        os.environ["VLLM_DT_MAX_BATCH_SIZE"] = str(
+            vllm_config.scheduler_config.max_num_seqs)
 
-        if cache_config and model_config:
```
yannicks1, line 109: this check is removed, is this save?

yannicks1 (changes requested):
Also, the code currently breaks here: https://github.com/vllm-project/vllm-spyre/blob/c0070a3d4b6bd90f6629c0f7b78211c7c87cc7c3/vllm_spyre/v1/core/scheduler.py#L213

vllm_spyre/platform.py
prashantgupta24: Added the checks 👍

prashantgupta24:
Seems like something's messed up on vllm/main, looking 👀

yannicks1 (approved):
LGTM, great work!

PR #122, opened 2025-04-28
:construction_worker: add test pypi workflow

joerunde:
I went ahead and setup a pending publisher for this on test.pypi.org

### DIFF
--- a/.github/workflows/build-and-publish.yaml
+++ b/.github/workflows/build-and-publish.yaml
@@ -1,0 +1,32 @@
+name: Build and Publish vllm-spyre Library
+
+on:
+    release:
+      types: [published]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


joerunde:
hmm, seems to be a problem lining up the publisher info on test.pypi.org :(

Will look more tomorrow!

joerunde:
just kidding, the test.pypi.org url has to also be set in the publish command, we're live! 

https://test.pypi.org/project/vllm-spyre/

I'll promote to regular pypi tomorrow. We should probably hook up `setuptools_scm` to make sure we don't flub any versions though

joerunde:
The pending publisher has been set up on pypi.org so this should be authenticated to push

rafvasq (approved):
It all looks good/straightforward to me

joerunde:
alright, 🤞!

Issue #121, opened 2025-04-28
Fix current main branch failures


joerunde:
See https://github.com/vllm-project/vllm-spyre/actions/runs/14711516759/job/41284841211

V0 is failing with:

```
>                   mm_processor_kwargs=seq_group.mm_processor_kwargs,
                    prompt_adapter_request=seq_group.prompt_adapter_request,
                )
E               AttributeError: 'SequenceGroup' object has no attribute 'mm_processor_kwargs'

.venv/lib/python3.12/site-packages/vllm_spyre/core/scheduler.py:1270: AttributeError
```

V1 is failing with:
```
ERROR 04-28 15:28:17 [core.py:396]   File "/home/runner/work/vllm-spyre/vllm-spyre/.venv/lib/python3.12/site-packages/vllm_spyre/v1/worker/spyre_worker.py", line 80, in compile_or_warm_up_model
ERROR 04-28 15:28:17 [core.py:396]     self._warmup_spyre_fixed_size(prompt_len, num_decode_tokens,
ERROR 04-28 15:28:17 [core.py:396]   File "/home/runner/work/vllm-spyre/vllm-spyre/.venv/lib/python3.12/site-packages/vllm_spyre/v1/worker/spyre_worker.py", line 286, in _warmup_spyre_fixed_size
ERROR 04-28 15:28:17 [core.py:396]     NewRequestData(
ERROR 04-28 15:28:17 [core.py:396] TypeError: NewRequestData.__init__() got an unexpected keyword argument 'prompt'
```

Issue #118, opened 2025-04-27
[Deprecation]: Remove legacy/input mapper from V0


DarkLight1337:
Apply vllm-project/vllm#15686 to the model runner.

These methods should not be used anymore:

```
MultiModalRegistry.has_processor
MultiModalRegistry.create_input_mapper
MultiModalRegistry.init_mm_limits_per_prompt
```

I plan to remove those methods from upstream by v0.10.0

joerunde:
Fixed by #123

PR #123, opened 2025-04-28
Fixes to get things working with upstream main

tjohnson31415:
Updates to get vllm-spyre working with vLLM `main` after some recent changes.

For V1, the optional `prompt` field was removed from some ouptut classes in vLLM. Since these were optional anyways (and just used in our warmup requests), there isn't any backwards compatibility issue.
https://github.com/vllm-project/vllm/pull/17214/files#diff-cafd89ce8a698a56acb24ada62831cbc7a980782f78a52d1742ba238031f296cL25

For V0, mulitmodal input processing has gone through an overhaul, the part that matters here is the removal of `mm_processor_kwargs` from `SequenceGroup`. This field was also optional before it was removed.
https://github.com/vllm-project/vllm/pull/15686/files#diff-0e64e07127ff20fc01be18cc95c0f58e0ff818e0dbabd5e8d2e07ca27943a7e4L1599

FIX https://github.com/vllm-project/vllm-spyre/issues/121 https://github.com/vllm-project/vllm-spyre/issues/118

### DIFF
--- a/vllm_spyre/v1/compat/output.py
+++ b/vllm_spyre/v1/compat/output.py
@@ -3,12 +3,15 @@ import inspect
 from dataclasses import dataclass
 from typing import ClassVar
 
+# yapf conflicts with ruff for this block
+# yapf: disable
 try:
     # vllm v0.8.2+
     from vllm.v1.core.sched.output import (
         NewRequestData as UpstreamNewRequestData)
 except ImportError:
     from vllm.v1.core.scheduler import NewRequestData as UpstreamNewRequestData
+# yapf: enable
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


joerunde (approved):
lpgtm!

tjohnson31415:
Whoops, not backwards compatible... Will fix

vllm_spyre/v1/compat/output.py
```diff
@@ -0,0 +1,31 @@
...
+        # The prompt field was removed in https://github.com/vllm-project/vllm/pull/17214/files#diff-cafd89ce8a698a56acb24ada62831cbc7a980782f78a52d1742ba238031f296cL25
+        # NB: this assumes kwargs are used at the call site
+        if self._legacy_signature:
+            kwargs['prompt'] = None
+
+        super().__init__(*args, **kwargs)
+
+
+if 'prompt' in inspect.signature(UpstreamNewRequestData.__init__).parameters:
+    NewRequestData._legacy_signature = True
```
joerunde, line 31: hah, I would've just wrapped the invocation in a try/except TypeError. This is slick

PR #124, opened 2025-04-29
:memo: Add release docs

joerunde:
This PR adds a RELEASING.md file with quick instructions on how to make a release. I didn't want to throw it in the main README since that's exposed on pypi.

This also deletes the workflow that automated the github release creation, because the pypi deploy job can't trigger on a release that's created by a github action without setting up alternative auth for it.

The `workflows/scripts` dir is also now completely unused, so I deleted it. 🔥🔥🔥


wallashss (approved):
LPGTM!

wallashss:
PS: Thanks for the clean up!

PR #126, opened 2025-04-29
[CB] Continuous batching support on spyre input batch

wallashss:
This PR extends the `spyre_input_batch` that will make it compatible with continuous batching.

Recap:

The original `gpu_input_batch` has some optimizations that minimizes copies of memory to the device, and it manipulates requests by adding/removing and making their data in a dense tensors.

Our `spyre_input_batch` was based on the `gpu_input_batch`. However, we currently cannot manipulated tensors that are allocated in a spyre card, therefore all tensors are in CPU. For that reason, we removed all the optimizations related of transfer data between devices. Another key difference is the layout of the memory in this batch: In static batching, finishing requests are just masked out, and not removed until the batch is fully done and it is reset. This masked out is done by `soft_remove_request`. There are some special management based on this mask to make sampling parameters in a dense fashion. 

> NOTE: see [get_unpadded_output_indices](https://github.com/vllm-project/vllm-spyre/blob/cb38f1bca4bf271bbb9022afc40fa1e0c8b3b8bb/vllm_spyre/v1/worker/spyre_input_batch.py#L438) for a explanation of how currently requests are structured for spyre considering a mask and padding.

Change:
To be able to support continuous batching, we keep the soft delete approach. But now it is possible to add new request in free slots of finished requests from previous step. Then, we can keep the rest of the implementation as-is.


vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -174,23 +174,40 @@ def req_ids(self) -> list[str]:
...
-        else:
-            self._req_ids[req_index] = req_id
-            self.req_output_token_ids[req_index] = request.output_token_ids
+        self._req_ids[req_index] = req_id
+
+        # NOTE: differently from gpu input batch, self.req_output_token_ids
+        # is not synced with self._req_ids, it should use
+        # self.model_indices_mask to resolve its index considering masked
+        # out requests.
+        shifted_index = self.model_indices_mask[:req_index].sum().item()
```
joerunde, line 208: should this call `self.req_idx_to_dense_index`?

wallashss: Updated it!

vllm_spyre/v1/worker/spyre_input_batch.py
prashantgupta24, line 1: `Since we do not yet support continuous batch` -> this line exists in this file. Do you think this line will change now? Or is there something still left for us to fully support it?

wallashss: You're right! I updated the comment. Please check it out.

tests/v1/worker/test_spyre_input_batch.py
```diff
@@ -18,9 +18,8 @@
 MAX_NUM_PROMPT_TOKENS = 64
 
 
-def _remove_requests(
-        input_batch: InputBatch, batch_size: int,
-        reqs: list[CachedRequestState]) -> tuple[set[str], list[int]]:
+def _remove_requests(input_batch: InputBatch, batch_size: int,
+                     reqs: list[CachedRequestState]) -> set[str]:
```
prashantgupta24, line 22: The function doc needs to be changed, it still says this function returns a `tuple`

tests/v1/worker/test_spyre_input_batch.py
```diff
@@ -69,11 +66,12 @@ def _construct_expected_sampling_metadata(
                                          dtype=torch.bool,
                                          device=device)
 
-    index_in_input_batch = 0
+    # index_in_input_batch = 0
```
prashantgupta24, line 69: remove commented line?
```suggestion
```

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -174,23 +177,39 @@ def req_ids(self) -> list[str]:
...
+        available_indices = self.req_indices_mask.logical_not().nonzero()
+        available_indices_list = available_indices.squeeze(dim=-1).tolist()
+        return available_indices_list[0] if available_indices_list else None
+
+    def req_idx_to_dense_index(self, req_index):
+        return self.req_indices_mask[:req_index].sum().item()
+
+    def req_id_to_dense_index(self, req_id):
+        req_index = self.req_id_to_index[req_id]
+        return self.req_idx_to_dense_index(req_index)
```
prashantgupta24, line 190: It would really help if we can add some good detailed function docs to these functions, ideally with a very basic example to demonstrate what exactly it's returning (if applicable)

prashantgupta24: Also adding the return types from these functions would be great too

tests/v1/worker/test_spyre_input_batch.py
```diff
@@ -196,67 +233,38 @@ def test_sampling_metadata_in_input_batch(batch_size: int):
...
-        assert torch.allclose(
-            expected_sampling_metadata.allowed_token_ids_mask,
-            sampling_metadata.allowed_token_ids_mask)
-    assert expected_sampling_metadata.bad_words_token_ids == \
-        sampling_metadata.bad_words_token_ids
+    sampling_metadata = input_batch._make_sampling_metadata()
+
+    # Create expected output.
+    expected_sampling_metadata = _construct_expected_sampling_metadata(
+        reqs, req_ids_retained, input_batch, device=torch.device(device))
```
prashantgupta24, line 268: thinking out loud - is it worth moving these 2 metadata calls within `compare_results`? Since you do it twice, it could result in some less duplicated code within the test

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -448,11 +471,11 @@ def get_unpadded_output_indices(self) -> dict[str, int]:
         This will output: {"B": 0, "C": 1}
         """
 
-        indices = self.model_indices_mask.nonzero().squeeze(dim=-1).tolist()
+        indices = self.req_indices_mask.nonzero().squeeze(dim=-1).tolist()
```
prashantgupta24, line 474: can we change the docstring for this function to reflect `self.req_indices_mask` as well?

vllm_spyre/v1/worker/spyre_input_batch.py
prashantgupta24, line 1: There are still 3 occurances of `model_indices_mask` in this file within docstrings, should convert them to `req_indices_mask` right

tests/v1/worker/test_spyre_input_batch.py
```diff
@@ -196,67 +233,38 @@ def test_sampling_metadata_in_input_batch(batch_size: int):
...
-    )
-    assert torch.allclose(
-        expected_sampling_metadata.presence_penalties,
-        sampling_metadata.presence_penalties,
-    )
-    assert torch.allclose(
-        expected_sampling_metadata.repetition_penalties,
-        sampling_metadata.repetition_penalties,
-    )
+    # Add more requests
```
prashantgupta24, line 256: worth adding a check here to see if `req_ids_to_remove` is 0. If that's the case we should not proceed since it's basically not doing anything new and just asserting the same stuff one more time?

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -312,21 +330,26 @@ def soft_remove_request(self, req_id: str):
...
         
-        TODO: in the future we shall remove/change this method to actual clean 
-        up all the data 
+        For the continuous batching, the removed request indices can be 
+        overwritten by new requests
         '''
         if req_id not in self.req_id_to_index:
             return
-        req_index = self.req_id_to_index[req_id]
+        # req_index = self.req_id_to_index[req_id]
```
prashantgupta24, line 338: remove commented line?
```suggestion
```

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -312,21 +330,26 @@ def soft_remove_request(self, req_id: str):
...
+        req_index = self.req_id_to_index.pop(req_id)
+
+        if req_index is None:
+            return None
```
prashantgupta24, line 342: Since we're not really returning anything
```suggestion
            return
```

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -312,21 +330,26 @@ def soft_remove_request(self, req_id: str):
...
-        TODO: in the future we shall remove/change this method to actual clean 
-        up all the data 
+        For the continuous batching, the removed request indices can be 
+        overwritten by new requests
         '''
         if req_id not in self.req_id_to_index:
             return
-        req_index = self.req_id_to_index[req_id]
+        # req_index = self.req_id_to_index[req_id]
+        req_index = self.req_id_to_index.pop(req_id)
```
prashantgupta24, line 339: Can the `req_index` be `None` if it exists?

wallashss: I guess not. BTW, I did an update in this part, There was a redundancy in the checking.

tests/v1/worker/test_spyre_input_batch.py
wallashss: I'm not sure. If I do this I will have a third function with two responsibilities, considering that I will not merge `compare_results` in that. So, I don't feel we will have gains. I guess right now it has an okay readability.

tests/v1/worker/test_spyre_input_batch.py
wallashss: I checked this already in the past. This only happens when `batch_size=1`, which may sometimes not remove the request. I guess add an extra if to skip this case has little gain with a little impact in code readability. If you think it really worth to add, I can do, but IMO it's okay like this.

wallashss:
Thanks a lot @prashantgupta24 for the review! I think that I addressed or answered to all your comments. Please let me know if I missed something. 

I also would appreciate if you could resolve the comments that I have addressed.

tests/v1/worker/test_spyre_input_batch.py
prashantgupta24: `num_reqs_to_remove` is a random int including and between 0 and `batch_size` though right? So it _could_ happen for other batch_sizes too, although the probability is less.

prashantgupta24 (approved):
Changes look good, thanks Wallas!

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -174,23 +177,81 @@ def req_ids(self) -> list[str]:
...
+        '''
+        This data structure has 3 types of references for data:
+        
+        - [request id | req_id] : str -> An id of the request, is passed as 
+        input in `add_request`.
+        - [request index | req_index | req_idx] : int -> The index of the data
+        in this batch. This index is aligned with `req_indices_mask` which can
+        deactivate indices in the batch. In static batching, the finished 
+        requests are only deactivated, and the data is not reorganized until
+        the bash is fully processed. On the other hand, in continuous batching,
```
prashantgupta24, line 190:
```suggestion
        the batch is fully processed. On the other hand, in continuous batching, 
```

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -174,23 +177,81 @@ def req_ids(self) -> list[str]:
...
+    def req_id_to_dense_index(self, req_id) -> int:
+        '''
+        This data structure has 3 types of references for data:
+        
+        - [request id | req_id] : str -> An id of the request, is passed as 
+        input in `add_request`.
+        - [request index | req_index | req_idx] : int -> The index of the data
+        in this batch. This index is aligned with `req_indices_mask` which can
+        deactivate indices in the batch. In static batching, the finished 
+        requests are only deactivated, and the data is not reorganized until
```
prashantgupta24, line 189:
```suggestion
        requests are only deactivated and the data is not reorganized until
```

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -312,21 +372,24 @@ def soft_remove_request(self, req_id: str):
...
-        TODO: in the future we shall remove/change this method to actual clean 
-        up all the data 
+        For the continuous batching, the removed request indices can be 
+        overwritten by new requests
         '''
-        if req_id not in self.req_id_to_index:
+
+        req_index = self.req_id_to_index.pop(req_id, None)
+
+        if req_index is None:
```
prashantgupta24, line 381: If it doesn't exist then it will throw a `key error` though

prashantgupta24: Oh my bad, did not check the default return. Makes sense now

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -312,21 +372,24 @@ def soft_remove_request(self, req_id: str):
...
         needed to keep the map of finished padded requests.
         
-        TODO: in the future we shall remove/change this method to actual clean 
-        up all the data 
+        For the continuous batching, the removed request indices can be 
+        overwritten by new requests
         '''
-        if req_id not in self.req_id_to_index:
+
+        req_index = self.req_id_to_index.pop(req_id, None)
```
prashantgupta24, line 379: The docstring for this function says
```
This method however, does not change the self.req_id_to_index
```

but this line is changing it right?

prashantgupta24: Also is removing requests from `req_id_to_index` still considered a `soft remove`?

wallashss: Yeah, I think it is safe to promote it to `remove_request`. 

> I thought masking out the request was the soft remove part, but maybe I understood it incorrectly

Yes, but after this round of changes, the batch is functional for the basic creation and delete of requests. Before, it was not possible, and most of the work was done by clearing the batch.

prashantgupta24 (approved):
LGTM, one small nit, great work!

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -40,13 +40,15 @@ class InputBatch:
...
+    the indices stored in `req_indices_mask`. Sometimes we need to check it
     to get the correct index of a request see `get_unpadded_output_indices`. 
     
+    For static batching, the correct usage of this class consists in add 
+    requests and clear the whole batch before process more requests. 
     
-    Since we do not yet support continuous batch, the correct usage of this
-    class consists in add requests and clear the whole batch before process
-    more requests. 
+    For continuous batching, when requests are "soft removed", it opens
```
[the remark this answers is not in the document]
wallashss: Sure! Missed that, thank you!

Issue #119, opened 2025-04-28
[CB] clean up request metadata for last submitted request


yannicks1:
### clean up request metadata for last submitted request

I have noticed that the request metadata, namely `self.req_ids2blocks`, ` self.req_ids2left_pads` and `self.free_blocks` are not freed up/ updated for the last submitted request. This should be solved to avoid running out of blocks when using continuous batching in server mode.

Note: code currently on branch `dev-updated-fms-api`

wallashss:
BTW:
>  I believe a possible solution to the above could also solve https://github.com/vllm-project/vllm-spyre/pull/82#discussion_r2044889419 in the same go.

I changed that in #127, I went back to your previous solution which is also what vllm does. vLLM has a case where you can call execute model with an "empty" scheduler output with just the request ids to finish. In this case we can properly clean up the data in the model runner side.

wallashss:
I did some testing, and I got some surprises:

- The online serving is doing fine. All data is cleaned when all requests finish.
- The offline does not. I debugged and checked that there are still request metadata (self.req_ids2blocks,  self.req_ids2left_pads and self.free_blocks). What is more surprising is that vllm upstream also has this problem in offline mode. I guess it is the way how the engine work wrapped by the llm class. In a nutshell it keeps running steps until the scheduler has pending requests, but in the "last step" it only puts in the queue to be released in a next step that is never executed. The model runner does not have a chance to release the metadata from the last request.

It is not good to have pending data to clean, but I think it is not so bad since it only in offline mode. if one send calls another `llm.generate` it will release in the next step and everything should keeping working in a fresh state. I think the solution is work with upstream to report/fix the issue and then update the vllm dependency to get the fix.

prashantgupta24:
cc @sducouedic this would be great to add a test for!

wallashss:
UPDATE:

I talked with @nikolaospapandreou and we observed the issue of the missing cleaning that I talked early only happens if we disable vllm multiprocessing with `VLLM_ENABLE_V1_MULTIPROCESSING=0`.

So I guess everything is fine, I suggest that we can close now this issue. Even though there's a potential issue without multiprocessing, I think it is an edge case that we should not worry about it for now.

yannicks1:
okay, thanks for the investigation! Do you agree @nikolaospapandreou that we can close this issue?

PR #127, opened 2025-04-29
[CB] Update spyre model runner for new spyre input batch 

wallashss:
This PR primarily integrate the model runner with the new spyre input batch for continuous batching.

Summary:
- Removed some hardcoded code to related to sampling parameters. Now, it should work fine requests with sampling parameters with CB.
- Update the logic in warm up to properly clean the data of the model runner, so it can have a consistent state to receive new requests.
- Now we have two input batches: one for the decoding and other for the prefill. Currently, we can only prefill one request at a time, and the batch is reset for every new prefill. This is similar to how the static batching works. This solution also ease to split the requests, because we only have to swap the batch to process without the need to activate/deactivate requests that are in decoding step.


wallashss:
This should be follow up for https://github.com/vllm-project/vllm-spyre/pull/126

wallashss:
Currently, it pass in the tests of https://github.com/vllm-project/vllm-spyre/pull/79

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -836,31 +890,8 @@ def prepare_model_input(
...
-            presence_penalties=dummy_tensors(0.1),
-            repetition_penalties=dummy_tensors(0.1),
-            output_token_ids=[[] for _ in range(num_reqs)],
-            min_tokens={},
-            logit_bias=[None for _ in range(num_reqs)],
-            allowed_token_ids_mask=None,
-            bad_words_token_ids=None,
-        )
+
+        model_inputs.sampling_metadata = self.input_batch.sampling_metadata
```
prashantgupta24, line 894: instead of adding it to `model_inputs`, we can use this line directly
```
self.input_batch.sampling_metadata
```

within the `execute_model`

```
            logits=logits,
            sampling_metadata=self.input_batch.sampling_metadata,
        )
```

This will also allow us to keep `model_inputs` as a `frozen` dataclass, which is how it was earlier.

wallashss: Sure, good idea!

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -208,6 +220,56 @@ def complete_warmup(self):
...
+                req_index,
+                start_token_index:end_token_index] = req_data.new_token_ids
+
+        if not update_for_cb:
+            return
+
+        # Continuous batching stuff
+        removed_req_indices: list[int] = []
+        for req_id in scheduler_output.finished_req_ids:
+            req_index = self.input_batch.soft_remove_request(req_id)
```
prashantgupta24, line 269: `soft_remove_request` isn't returning anything is it?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -132,7 +144,7 @@ def _prepare_pad_input_ids(
             seq_len = input_ids_i.size(0)
             if max_len > seq_len:
                 logger.info(
-                    "Left padding request of length %d tokens to %d tokens.",
+                    "Left padding request ofla length %d tokens to %d tokens.",
```
prashantgupta24, line 147: typo?
```suggestion
                    "Left padding request of length %d tokens to %d tokens.",
```

wallashss: Worse: The cat typed this and committed. I'll revert it 😄

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -874,6 +912,7 @@ def execute_model(
 
         t0 = time.time()
 
+        self._update_states(scheduler_output)
```
wallashss, line 915: NB: I changed a little bit the behavior there. `self._update_states` now free the allocated pages for finishing requests. This order important, so we can finish requests without the need to decode or prefill. For instance, check the next snippet, and a related PR https://github.com/vllm-project/vllm/pull/14388 that introduced this state at vllm. 

I'm not 100% sure if there are other side effects, would be nice If anyone could confirm that. 

@prashantgupta24 @yannicks1 @nikolaospapandreou 
Next snippet:
```
if not scheduler_output.total_num_scheduled_tokens:
            # Return empty ModelRunnerOuptut if there's no work to do.
            return CBSpyreModelRunnerOutput(
                req_ids=[],
                req_id_to_index={},
                sampled_token_ids=[],
                spec_token_ids=None,
                logprobs=None,
                prompt_logprobs_dict={},
                tkv=0,
            )
```

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24, line 1: Reminder to make the `ModelForwardInputs` dataclass `frozen` again and remove `sampling_metadata`

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -602,6 +635,12 @@ def _prepare_prompt(
 
         # Internal state is managed here.
         slot_mapping = []
+
+        # TODO: we are deactivating all, because we
+        # only encode or prefill at time.
+        # self.input_batch.deactivate_all_requests()
+        self.prefill_batch.clear_requests()
```
prashantgupta24, line 642: Nit: the docstring for `clear_requests` says
```
Clear the batch, mostly used by static batching
```
If we plan on using it from CB, let's change the docstring

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -602,6 +635,12 @@ def _prepare_prompt(
...
+
```
prashantgupta24, line 643: Shouldn't we also do this here?
```suggestion
self.input_batch.clear_requests()

```

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -602,6 +635,12 @@ def _prepare_prompt(
 
         # Internal state is managed here.
         slot_mapping = []
+
+        # TODO: we are deactivating all, because we
+        # only encode or prefill at time.
+        # self.input_batch.deactivate_all_requests()
```
prashantgupta24, line 641: Shouldn't we do this instead here?

```suggestion
       self.input_batch.clear_requests()
```

wallashss: This method `deactivate_all_requests` does not exist anymore. It was something that tried, but did not work as expected. This approach was later replaced by the prefill_batch.

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24: Should we also clear requests from input batch here?

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -346,13 +346,23 @@ def _warmup_spyre_dynamic_size(self, special_token_ids):
...
+            num_common_prefix_blocks=0,
+            # The requests to be removed
+            finished_req_ids=set([r.req_id for r in dummy_requests]),
+            free_encoder_input_ids=[],
+            structured_output_request_ids={},
+            grammar_bitmask=None,
+        )
+        self.execute_model(scheduler_output)
 
         self.model_runner.tkv = 0  # type: ignore[union-attr]
```
wallashss, line 368: @yannicks1, considering that the previous `execute_model` should correctly clean up the request of the model runner do we still need to reset the tkv?

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: > Should we also clear requests from input batch here?

There is trick logic going on here. I'll update the PR description soon. But in a nutshell, our model runner does one of the following: prefill or decode, like the vlllm V0 engine. On continuous batching We should never clear the input_batch, it must be "manually" cleaned by finishing the requests. On the other hand, the prefill decode (which currently only support batch_size=1) we can always reset it before a new prefill like in the same way we do with static batching.

Check this code snippet that execute this logic:

```py
def prepare_model_input(...):

        # NOTE: We assume that all sequences in the group are all prompts or
        # all decodes.
        # Also assuming that new sequences are prefills
        is_prompt = len(scheduler_output.scheduled_new_reqs) > 0

        # Prepare input tensors.
        if is_prompt:
            # Assert no running requests
            assert len(scheduler_output.scheduled_cached_reqs) == 0

            return self._prepare_prompt(scheduler_output.scheduled_new_reqs)
        else:
            [....]
            return self._prepare_decode(scheduler_output.scheduled_cached_reqs)
```

prashantgupta24: Hmm. Let me wrap my head around that. Right now, using this code, I get an error during the second prefill though
```
vllm_spyre/v1/worker/spyre_model_runner.py:926: in execute_model
vllm_spyre/v1/worker/spyre_model_runner.py:890: in prepare_model_input
    self._prepare_prompt(scheduler_output.scheduled_new_reqs)
vllm_spyre/v1/worker/spyre_model_runner.py:679: in _prepare_prompt
    self.input_batch.add_request(req_state)

    def add_request(
        self,
        request: "CachedRequestState",
        req_index: Optional[int] = None,
    ) -> None:
        if req_index is None:
            req_index = self.get_available_index()
>       assert req_index is not None
E       AssertionError
```

prashantgupta24: This is when I try to run the continuous batching tests that I have in the other PR

prashantgupta24: Adding the input batch clear request somehow solved the issue but maybe I understood it wrong

wallashss: Do you mean the tests of #79 ?

I've been testing with them. But maybe I missed something.

prashantgupta24: Yep. Try it out?

prashantgupta24: Let's add a warning in case the user sets batch size to 1 with CB?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -582,6 +579,37 @@ def __init__(
...
+        super()._update_states(scheduler_output)
+
+        # Continuous batching stuff
+        for req_id in scheduler_output.finished_req_ids:
+            if req_id in self.req_ids2blocks:
+                logger.debug("Freeing request id: %s", req_id)
+                for freed_block in self.req_ids2blocks[req_id]:
+                    self.free_blocks.append(freed_block)
+                del self.req_ids2blocks[req_id]
+                del self.req_ids2left_pads[req_id]
```
prashantgupta24, line 608: should we also delete completed request from `self.requests?`

wallashss: Good catch! Yes. Fixed it. Thanks!

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -582,6 +584,38 @@ def __init__(
...
+
+        for req_id in scheduler_output.finished_req_ids:
+            del self.requests[req_id]
+            self.input_batch.remove_request(req_id)
```
prashantgupta24, line 617: can we not move this within the `for` loop above?

tests/e2e/test_spyre_cb.py
```diff
@@ -108,15 +108,12 @@ def test_cb_with_steps(model: str, backend: str, cb: int,
...
-                                          add_special_tokens=False)[1]] \
-        if "llama-194m" in model \
-        else tokenizer.encode(stop_token,
-                                          add_special_tokens=False)
+        # There's an issue with JackFram/llama-160m tokenizer it returns 2
+        # tokens for "1", the first is quotes, the second is the actual number
+        # 1. So, we always get the last token id which also works for the case
+        # the tokenizer return a single token id.
+        stop_token_ids = tokenizer.encode(stop_token,
+                                          add_special_tokens=False)[-1:]
```
prashantgupta24, line 116: good idea!

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: Yes, I kinda of missed that. Because I am planning to change this block management very soon.

I merged the loops, makes sense.

prashantgupta24 (approved):
lgtm!

vllm_spyre/v1/worker/spyre_worker.py
yannicks1: > considering that the previous execute_model should correctly clean up the request of the model runner 

does it reset the tkv too? @wallashss

vllm_spyre/v1/worker/spyre_model_runner.py
yannicks1: > I changed a little bit the behavior there. self._update_states now free the allocated pages for finishing requests. This order important, so we can finish requests without the need to decode or prefill.

Does this solve [this](https://github.com/vllm-project/vllm-spyre/issues/119) issue ?

vllm_spyre/v1/worker/spyre_worker.py
wallashss: Currently, no. I kept the reset by the worker. I should be fine.

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: I think now it is solved!

PR #129, opened 2025-04-30
[CB] Scheduler constraints

### LINKED ISSUE #96: [CB] Add scheduler constraints
### Continuous Batching: scheduler constraints

We need to incorporate several constraints into the `ContinuousBatchingSpyreScheduler`:
- Ensure that the prompt length of an inserted prompt is smaller than the current decode tkv. 
- Only schedule prompts when their number of requested output tokens is less than the maximal context length - the current tkv
- Deal with requests where no max number of new tokens is specified.

sducouedic:
To match all the scheduling constraints, the scheduler needs to know the current tkv value of the model runner. 

The changes in this PR allow to pass this tkv value from the spyre model runner back to the scheduler through the `ModelRunnerOutput` instance returned by the `execute_model` function of the model runner class (as designed by vllm). To do so, an instance of the new class `CBSpyreModelRunnerOutput` inheriting from `ModelRunnerOutput` is return with the additional `tkv` attribute set. 

NB: this has been tested with tensor parallelism as well.

Closes #96


vllm_spyre/v1/core/scheduler.py
```diff
@@ -321,7 +338,11 @@ def schedule(self) -> "SchedulerOutput":
...
-                self.max_num_running_reqs and\
-                len(self.waiting) < max_prompt_batch_size
+        max_context_len = envs_spyre.VLLM_SPYRE_MAX_CONTEXT_LENGTH
+
+        return len(self.running)+len(self.waiting) == 0 or\
+            (len(self.running)+len(self.waiting)\
+                < self.max_num_running_reqs and\
+            len(self.waiting) < max_prompt_batch_size and\
+            self.holdback_queue[0].num_prompt_tokens <= self.tkv and\
+            self.holdback_queue[0].max_tokens <= (max_context_len - self.tkv))
```
joerunde, line 348: I think it'd be a bit easier to understand this if we passed `self.holdback_queue.peek()` to this method so that we have
```
def can_schedule(self, request) -> bool:
    ...
    request.num_prompt_tokens <= self.tkv and \
    request.max_tokens <= (max_context_len - self.tkv)
```

yannicks1, line 348: could we maybe make this conditional check more readable for the user? 
I am thinking of something like this: 
```
        start_new_batch = len(self.running) + len(self.waiting) == 0
        # check that there is space in the current decode batch
        cond1 = len(self.running) + len(
            self.waiting) < self.max_num_running_reqs
        # check that there is space in the prefill batch
        cond2 = len(self.waiting) < max_prompt_batch_size
        # check that the prompt length does not exceed the current tkv
        cond3 = self.holdback_queue[0].num_prompt_tokens <= self.tkv
        # check that the number of requested tokens can be served
        cond4 = self.holdback_queue[0].max_tokens <= (max_context_len -
                                                      self.tkv)
        return start_new_batch or (cond1 and cond2 and cond3 and cond4) 
```

vllm_spyre/v1/core/scheduler.py
```diff
@@ -140,9 +144,26 @@ class ContinuousBatchingSpyreScheduler(SpyreScheduler):
...
+        scheduler_output: SchedulerOutput,
+        model_runner_output: ModelRunnerOutput,
+    ) -> EngineCoreOutputs:
+        # Need an instance of CBSpyreModelRunnerOutput which holds the tkv value
+        assert isinstance(
+            model_runner_output, CBSpyreModelRunnerOutput
+        ), "Expecting an instance of CBSpyreModelRunnerOutput"
+        "when doing continuous batching."
+        self.tkv = model_runner_output.tkv
+        return super(SpyreScheduler,
```
joerunde, line 163: Now that the continuous batching scheduler doesn't inherit from the static batching scheduler, I think we can just user `super()` everywhere

yannicks1: We still have the `SpyreScheduler` base class in between `Static/ContinuousBatchingSpyreScheduler` and the V1 `Scheduler`. Thus `super(SpyreScheduler, self)` is needed here.

vllm_spyre/v1/core/scheduler.py
```diff
@@ -235,9 +236,25 @@ class ContinuousBatchingSpyreScheduler(SpyreScheduler):
     running: list[Request]
 
     def __init__(self, *args, **kwargs) -> None:
+        self.tkv = 0
```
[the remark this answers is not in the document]
yannicks1: as @sducouedic is on vacation, I addressed this [here](https://github.com/vllm-project/vllm-spyre/pull/129/commits/529871202889d05a895c98d6c5ffc9e4f161c3c4)

vllm_spyre/v1/core/scheduler.py
yannicks1: did that [here](https://github.com/vllm-project/vllm-spyre/pull/129/commits/40eb201d5efe5e6dfc7826c1cada93c16227a269). Had to use `self.holdback_queue[0]` instead of `self.holdback_queue.peek()` as deque does not have peek()...

nikolaospapandreou (approved):
LGTM

PR #131, opened 2025-04-30
[CB] strip repeated left padding on batch level

### LINKED ISSUE #120: [CB] strip repeated left padding on batch level
### strip repeated left padding on batch level

As correctly pointed out by [Josh](https://github.com/JRosenkranz), we should check for repeated left padding on batch level and remove it. 

Problem: with our monotonically increasing, homogeneous `tkv` (compiler requirement) we might end up with a lot of left padding for all our running sequences and therefore block request from being processed due to a limited token budget. 

Fix: We should check for the maximum common number of left padded blocks among the running sequences and strip them by updating all the request metadata, freeing the blocks and decreasing `tkv` accordingly.

yannicks1:
### [CB] strip repeated left padding on batch level

changes:
- remove fully padded blocks in a batch to save computation and decrease tkv by updating request metadata and tensors accordingly
- introducing an env variable `VLLM_SPYRE_RM_PADDED_BLOCKS` to enable/disable this feature (this switch was requested by the compiler team) (default disabled)
- optimization: using deque() instead of python list provides O(1) popping from the left instead of O(n). Don't expect a huge performance boost, but we want to make sure to use the optimal data structure for maintaining the block ids. 

closes #120


joerunde:
Do I understand correctly that this would trigger when the oldest request in a batch (ie that started at the lowest tkv) finishes, so that all remaining requests start with padded blocks?

Generally LGTM, would like some tests but I guess we're in the wild west here with continuous batching until hardware support catches up

joerunde (approved):
LGTM, for transparency I haven't tried running this code

yannicks1:
> Do I understand correctly that this would trigger when the oldest request in a batch (ie that started at the lowest tkv) finishes, so that all remaining requests start with padded blocks?

exactly!

PR #132, opened 2025-04-30
[Draft] Bump vLLM and remove scheduler rejection logic

joerunde:
We should be good to simplify our schedulers now that 0.8.5 is out


vllm_spyre/platform.py
```diff
@@ -218,16 +218,21 @@ def validate_request(
...
-            if (prompt_len + max_tokens
+            # context length. This needs to take the padded prompt length
+            # into account.
+
+            # ceil division to pad to next block boundary
+            n = prompt_len
+            d = 64  # hardcoded AIU Spyre block size
+            prompt_padding_len = ((n + d - 1) // d) * d
+            if (prompt_padding_len + max_tokens
                     > envs_spyre.VLLM_SPYRE_MAX_CONTEXT_LENGTH):
```
prashantgupta24, line 229: bit of a tangential question but how does the `validate_request` get access to vllm configs? Thinking out loud when we want to remove these global env variables

prashantgupta24: possibly add a `cls` config variable that gets assigned the `vllm_config` at `check_and_update_config`? 🤔

pyproject.toml
```diff
@@ -13,7 +13,7 @@ license = {text = "Apache 2"}
 dependencies = [
     "fms-model-optimizer>=0.2.0",
     "ibm-fms==0.0.8",
-    "vllm",
+    "vllm>=0.8.5",
```
wallashss, line 16: Question: set vllm dependency like won't make it install its dependencies like cuda related libs?

vllm_spyre/platform.py
joerunde: hmmmmmm

`vllm.get_config()`?

pyproject.toml
joerunde: well, it won't if you use `uv sync` because we've set the overrides to never install torch.

What I don't currently know is if there's a clean way to get `uv pip install vllm-spyre` to have that same behavior when installing from a wheel, like from pypi

vllm_spyre/platform.py
prashantgupta24: uhhh vllm/config.py already imports vllm.platforms :( :(

prashantgupta24 (approved):
lgtm

PR #137, opened 2025-05-02
:bug: Fix new token minimum requirement error message

gkumbhat:
### Changes

If we set `VLLM_SPYRE_WARMUP_NEW_TOKENS=2`, currently we get error message:
```
AssertionError: VLLM_SPYRE_WARMUP_NEW_TOKENS must be at least 2 (spyre requirement).
```

This is a big confusing, since "atleast 2" would mean, `2` is allowed value, but actually minimum requirement is `3` as per the `if` condition there.

This PR fixes the error message.


rafvasq (approved):
LGTM!

PR #139, opened 2025-05-02
fix: Add validation and test for prompt len % 64

### LINKED ISSUE #133: Validate static batching prompt lengths
The prompt lengths provided in `VLLM_SPYRE_WARMUP_PROMPT_LENS` must all be multiples of 64, because there is a hardcoded constraint that all prompts must be padded to multiples of 64.

We should raise at startup when validating the config if any provided prompt lengths are not multiples of 64

rafvasq:
- Adds validation step for prompt length
- Adds a test for ^ catching invalid length

Closes #133


joerunde (approved):
lgtm!

PR #144, opened 2025-05-07
Fix test pypi publications

joerunde:
Currently the publications on main to test pypi are broken, because the job doesn't checkout the main branch history to calculate the correct semantic version to push based on the last tag. This fixes that, so we should have green deployments to test pypi again, and for any failures we can investigate the actual packaging failure.

Here's a test run of the changes: https://github.com/vllm-project/vllm-spyre/actions/runs/14892371588/job/41827331149


wallashss (approved):
Nice!

PR #149, opened 2025-05-08
Add options to torch.compile

tdoublep:
We need to pass additional options to torch.compile to ensure that symbols get added to the graph.


sducouedic:
Does this only apply to decoder models or should we also modify `SpyreEmbeddingModelRunner` to remove the "default" compile mode and pass the sendnn_dynamic option?

yannicks1:
@sducouedic legitimate question. `SpyreCausalLM` does not get used in the embedding model path. V0 embedding model still runs, V1 is not implemented yet. I don't think these changes have to be applied to the embedding models as specific to CB/paged attention compiler support, but I might be wrong...

joerunde:
bot:test
TEST_FILE=tests/e2e/test_spyre_basic.py

PR #153, opened 2025-05-12
🐛 set default value for tkv

prashantgupta24:
since upstream PR introduced a default param in `ModelRunnerOutput`

https://github.com/vllm-project/vllm/pull/17751/files#diff-267e8bc1629d4ae978c28eea04881c9934449a35c1f74c6214636f213bde8ccdR104-R105


joerunde (approved):
Ah, seems a bit unfortunate since this should always have to be set by the caller :(

But not a huge deal, we have (or will soon have) tests to cover it

Issue #75, opened 2025-04-07
Add tests for continuous batching


joerunde:
Followup for the continuous batching work

prashantgupta24:
Closing this for now, creating individual issues for leftover tasks.

PR #156, opened 2025-05-13
[CB] test scheduler tkv

sducouedic:
This PR enables testing of the scheduler step-by-step in a parametrized fashion where the ground truth values (e.g. number of running, number of waiting, tkv value) can be passed to the function through a list of dictionary.

So far only prompts sequences that align well with the block boundaries has been tested, but the idea is to enable testing for different situations, by just passing the list of reference attributes for relevant steps. 

For https://github.com/vllm-project/vllm-spyre/issues/75


prashantgupta24:
@sducouedic thanks for starting work on this!

tests/e2e/test_spyre_cb.py
```diff
@@ -87,149 +86,134 @@ def test_cb_handling(
...
-        assert request_outputs[1].request_id == "1"  # req 1 is decoding
-        assert request_outputs[1].finished  # request 1 is done
-        assert request_outputs[1].outputs[0].text == " 3 2 "
 
-        assert len(engine_core.scheduler.waiting) == 1
-        assert len(engine_core.scheduler.running) == 1
+    # set env vars
+    monkeypatch.setenv("VLLM_SPYRE_USE_CB", "1")
+    monkeypatch.setenv("VLLM_USE_V1", "1")
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
```
prashantgupta24, line 101: Interesting, I don't really have a strong opinion on how `monkeypatch` should be used. We have it used as is and with `context` in spyre so doesn't really break convention. IMO I like the `context` use just because I know visually that once the context scope is over the env settings will definitely be removed.

tests/e2e/test_spyre_cb.py
```diff
@@ -87,149 +86,134 @@ def test_cb_handling(
...
+
+    # add first request
+    engine.add_request("1", prompt1, sampling_params)
+
+    assert len(engine_core.scheduler.waiting) == 1
+    assert len(engine_core.scheduler.running) == 0
+
+    request_outputs = engine.step()
+    assert len(request_outputs) == 1  # only 1 request
+    assert request_outputs[0].request_id == "1"  # req 1 is decoding (prefill step)
```
prashantgupta24, line 137: thanks for the extra comments! Definitely helpful

tests/e2e/test_spyre_cb.py
prashantgupta24, line 1: Re: the tokenizer, I had it using `stop` strings instead of `stop_tokens`. But in the earlier implementation of `engine core`'s `step` function, it was not being used correctly, that's why I had to revert. But glad that it works now!

prashantgupta24: This change is now merged to `main`. Thanks!

tests/e2e/test_spyre_cb.py
```diff
@@ -87,149 +86,134 @@ def test_cb_handling(
 @pytest.mark.parametrize("model", get_spyre_model_list())
 @pytest.mark.parametrize(
     "backend", [pytest.param("eager", marks=pytest.mark.cpu, id="eager")])
-@pytest.mark.parametrize("cb",
-                         [pytest.param(1, marks=pytest.mark.cb, id="cb")])
 # @pytest.mark.parametrize("vllm_version",
 #                          [pytest.param("V1", marks=pytest.mark.v1, id="v1")])
-def test_cb_with_steps(model: str, backend: str, cb: int,
+def test_cb_with_steps(model: str, backend: str,
```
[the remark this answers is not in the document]
sducouedic: Ok I see, I didn't know that! could we then simply mark the function instead of going through parametrization?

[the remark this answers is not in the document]
prashantgupta24: yeah I guess? Since it's temporary I didn't really think too much about it, but feel free to do so!

tests/e2e/test_spyre_cb.py
sducouedic: I guess it is really about personal preference here. I personally don't like to have extra indentation if not strictly unnecessary, especially with the formatter enforcing line length of 80. When using the monkeypath context here, and with my extra comment "(prefill step)", the corrected output of the formatter is a bit ugly, with lines that get wrapped. If that's okay we can remove the context here and I will change it everywhere else in a separate PR as you suggested?

prashantgupta24: Oh interesting! I did not think about that. Yeah sure that works!

prashantgupta24:
I also thought about what else we can assert in these tests. We have some constraints given that we only have the engine core variables to work with. I guess one value is the scheduler's tkv, which I'm assuming you also thought since you named the pr with that value?

prashantgupta24:
I also saw in another PR that you closed where you had a loop which went through the engine core's step function and that reduced the number of lines in this test by a huge margin. Any reason you closed that? If we can make that work that will help reduce the duplicate step functions we have here

sducouedic:
I closed my other PR because I noticed this morning your changes, we were kind of doing the same work in parallel. I wasn't aware that you were also writing tests for CB and didn't want to have another test file testing the same continuous batching behaviour. I am refactoring the function your are mentioning in this branch now and will push and get your comments and suggestions :)

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +83,175 @@ def test_cb_handling(
...
+    request_outputs = engine.step()
+    assert len(request_outputs) == 2  # both requests decoding now
+    assert request_outputs[0].request_id == "2"  # req 2 is decoding
+    assert request_outputs[1].request_id == "1"  # req 1 is decoding
+
+    assert len(engine_core.scheduler.waiting) == 0
+    assert len(engine_core.scheduler.running) == 2
+
+    # add the third request
+    # but since VLLM_SPYRE_MAX_BATCH_SIZE=2
```
prashantgupta24, line 181: just realized we don't set this env var anymore
```suggestion
    # but since max_num_seqs=2
```

sducouedic: good catch

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +91,412 @@ def test_cb_handling(
...
+        "tkv": 127,
+        "num_waiting": 1,
+        "num_running": 2,
+        "request_outputs": ["1", "0"]
+    # }, {
+    #     "step": 66,  # Sequence 0 finishes at step 66 (2 prefills + 64 decodes)
+    #     "tkv": 128,
+    #     "num_waiting": 1,
+    #     "num_running": 1,
+    #     "request_outputs": ["1"]  # TODO this fails?? Why do we observe ["1", "0"]?
```
prashantgupta24, line 330: I think that's because when a request finishes, it is marked `finished` at the last step. So here, in step 66 so it still shows up in `request_outputs` as `finished` and then gets cleared in the next step. `request_outputs[1].finished` should be `True` in this case

prashantgupta24: That should happen for request 1 as well though when it finishes 🤔

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +91,412 @@ def test_cb_handling(
...
+        mm_hashes=None,
+        mm_placeholders=None,
+        sampling_params=sampling_params,
+        eos_token_id=None,
+        arrival_time=0,
+        lora_request=None,
+    )
+
+
+def get_params_test_blocks_borders_aligned_prompts():
```
prashantgupta24, line 284: btw I really like this way of testing!

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +91,412 @@ def test_cb_handling(
...
+        "num_waiting": 0,
+        "num_running": 2,
+        "request_outputs": ["2"]
+    }, {
+        "step": 68,  # Decode sequences 1 and 2
+        "tkv": 129,
+        "num_waiting": 0,
+        "num_running": 2,
+        "request_outputs": ["2", "1"]
+    }, {
```
prashantgupta24, line 343: yeah after checking it out I see `seq 1` actually finishes at step 69 instead of 71
```
E               AssertionError: Step 69, request outputs
E               assert ['2', '1'] == ['2']
```

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +91,412 @@ def test_cb_handling(
...
+                                         temperature=0.0,
+                                         ignore_eos=True)
+        request = create_random_request(request_id=i,
+                                        num_tokens=prompt_length,
+                                        sampling_params=sampling_params)
+        requests.append((add_step, request))
+
+    # Run steps, until last step is reached (provided through parameter)
+    request_outputs = []
+    for step in range(last_step):
```
prashantgupta24, line 464: We can get the last step from `checked_steps` directly right? I don't see any other place it's used.

Also we want to loop till one more than the last step to account for the last step.

```suggestion
    for step in range(checked_steps[-1]['step'] + 1):
```

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +91,412 @@ def test_cb_handling(
...
+        "tkv": 133,
+        "num_waiting": 0,
+        "num_running": 1,
+        "request_outputs": ["2"]
+    }, {
+        # Sequence 2 finishes at step 73 (start step + 1 prefill + 7 decodes)
+        "step": 73,  
+        "tkv": 134,
+        "num_waiting": 0,
+        "num_running": 1,
```
prashantgupta24, line 361: This step wasn't running since the loop only looped till 73 (excluding it). I have a suggestion below on how to fix it instead of using `last_step` explicitly
```suggestion
        "num_running": 0,
```

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +91,412 @@ def test_cb_handling(
...
+
+    # Assert there is no more running or waiting request after last step
+    assert len(scheduler.waiting) == 0, \
+        "Last step done but still there are waiting requests"
+    assert len(scheduler.running) == 0, \
+        "Last step done but still there are running requests"
+
+    # Tkv in scheduler is cleared one step later
+    _ = engine_core.step().outputs
+    assert scheduler.tkv == 0, "Tkv not cleared two steps after last step"
```
prashantgupta24, line 502: Honestly we can even include a step 74 that checks for everything cleared
```
        {
            "step": 74,
            "tkv": 0,
            "num_waiting": 0,
            "num_running": 0,
            "request_outputs": [],
        },
```

this can eliminate all these lines, although it does take a hit on readability. Will leave it up to you

tests/e2e/test_spyre_cb.py
prashantgupta24: Worth exploring if we can attach `request_outputs[n].finished` as a boolean check for each step somehow

tests/e2e/test_spyre_cb.py
sducouedic: good catch! trying to figure out why 69

tests/e2e/test_spyre_cb.py
sducouedic: good idea, maybe test also which ids are running and waiting instead of the length. This will help to decorticate the steps

tests/e2e/test_spyre_cb.py
sducouedic: You're right, I'll change that

tests/e2e/test_spyre_cb.py
sducouedic: mmh
maybe the looper can have no idea of how it should finishes, neither about the initialization and just test everything given the elements in `checked_steps`. That's probably the simplest.

tests/e2e/test_spyre_cb.py
sducouedic: Ok I have it now. But should definitely test more values like `request_outputs[n].finished` as you suggested

tests/e2e/test_spyre_cb.py
sducouedic: done

prashantgupta24:
Looking great! Not sure if it should be part of this test itself, but we can add a request with a prompt_length > tkv to see how it behaves (basically it will be left in waiting until the tkv value crosses over its prompt length, IIUC)

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +91,415 @@ def test_cb_handling(
...
+        request_id=str(request_id),
+        prompt_token_ids=[request_id] * num_tokens,
+        mm_inputs=None,
+        mm_hashes=None,
+        mm_placeholders=None,
+        sampling_params=sampling_params,
+        eos_token_id=None,
+        arrival_time=0,
+        lora_request=None,
+    )
```
prashantgupta24, line 281: we need to add `cache_salt=None` as part of `EngineCoreRequest` to make the `vllm:main` check pass

```suggestion
def create_random_request(
        request_id: int, num_tokens: int,
        sampling_params: SamplingParams) -> EngineCoreRequest:
    return EngineCoreRequest(
        request_id=str(request_id),
        prompt_token_ids=[request_id] * num_tokens,
        mm_inputs=None,
        mm_hashes=None,
        mm_placeholders=None,
        sampling_params=sampling_params,
        eos_token_id=None,
        arrival_time=0,
        lora_request=None,
        cache_salt=None
    )
```

prashantgupta24:
Oh hey you probably knew this but I finally figured out a way to get hold of the `model runner` through the `engine_core`:
```
runner_obj = engine_core.model_executor.driver_worker.worker.model_runner
```
this is an instance of `ContinuousBatchingSpyreModelRunner`, so we can access all the internal variables like `req_ids2blocks, req_ids2left_pads, tkv, free_blocks` etc! Should be fun to assert stuff with them

tests/e2e/test_spyre_cb.py
sducouedic: The cb test now fail with the default vllm version. We need to update it.

sducouedic:
No I didn't know that! I was looking for that the other day, this will be useful if we want to test the padding at some point.

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +93,555 @@ def test_cb_handling(
...
+        request = create_random_request(request_id=i,
+                                        num_tokens=prompt_length,
+                                        sampling_params=sampling_params)
+        requests.append((add_step, request))
+
+    # if set, in-between steps are added as normal decode steps
+    if auto_check_basic_decode_steps:
+        checked_steps = augment_checked_steps(checked_steps)
+
+    # Run steps, until last step is reached (provided through parameter)
```
prashantgupta24, line 618: `(provided through parameter)` -> is this a leftover comment from earlier implementation where the `last_step`  was given explicitly?

tests/e2e/test_spyre_cb.py
prashantgupta24: oh whoops yeah good point

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +93,555 @@ def test_cb_handling(
...
+            "waiting": [],
+            "running": [],
+            "request_outputs": []
+        }
+    ]
+
+    return (seqs_max_tokens, prompts_lengths, steps_add_reqs, checked_steps)
+
+
+def get_params_test_blocks_borders_misaligned_prompts():
```
prashantgupta24, line 399: I'm trying to understand what you mean by `misaligned_prompts` here - feels relatively similar to `aligned_prompts`? I thought this would test the scenario in which there is a small request and a very large request. That could test the scenario where the scheduler pauses to schedule the large request until the tkv crosses its `prompt_length`

prashantgupta24: Ah from what I understand you wanted to test if request 2 finishes before request 0? So basically a misalignment in the sequence of finished requests. In that case could the test read `misaligned_finish_sequence` or something to understand it better?

tests/e2e/test_spyre_cb.py
sducouedic: ah yes, I should remove that

tests/e2e/test_spyre_cb.py
sducouedic: No this tests something else. So what happens is that CB is working on blocks of length 64 fashion, and when a sequences finishes and another starts, if the value at that time is not a multiple of 64, then this involves some different logic than when the tkv is 64 (some right padding). In the first test the second sequences get scheduled when `tkv=128`, so no right-padding is involved, and in the second test the second sequence gets scheduled when `tkv=120` so right-padding to 128 is required. It is true that first Sequence 2 finishes before Sequence 1, it's kind of bonus. Maybe it misleading, and I should make them finish in order?

About the other scenarios to test and including your suggestion, should we merge that PR first, and then this way people will be free to add some more tests?

sducouedic: I added some docstring to explain. It was not clear before, thanks for pointing this

prashantgupta24: Ah okay. Let me wrap my head around that . 

And yes, let's merge this PR before trying to include the other edge cases. I'll review it once after lunch

sducouedic: Btw, now I am wondering if I we also compare the final output with some ground truth (e.g. using HuggingFace), like you did in your step-by-step test at the end you still checks the correctess of the output. My first opinion was that it was already tested in your first end-to-end `test_cb_handling` test so need for checking. But given a second thought, even if the tkv and all scheduling is correct, nothing guarantees that some edge case produces a wrong output. Maybe this can be added later in another PR.

prashantgupta24: Yeah hopefully some amount of output correction is done by the basic `test_cb_handling`, but we can definitely add more edge cases to it

prashantgupta24 (approved):
LGTM. Great work!

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +92,389 @@ def test_cb_handling(
...
+            "step": 69,
+            "tkv": 130,
+            "waiting": [],
+            "running": [],
+            "request_outputs": ["1"],
+            "finished_requests": ["1"]
+        },
+        {
+            # Tkv should be cleared one step later
+            "step": 70,  # Decode sequence 2
```
prashantgupta24, line 319: nothing's decoding at this point right?
```suggestion
            "step": 70,
```

sducouedic:
Thanks for your detailed review @prashantgupta24

prashantgupta24:
> Thanks for your detailed review @prashantgupta24

You're welcome!

tests/e2e/test_spyre_cb.py
```diff
@@ -83,177 +92,389 @@ def test_cb_handling(
...
+    ])
+def test_scheduler_cb_steps_tkv(model: str,
+                                backend: str,
+                                monkeypatch: pytest.MonkeyPatch,
+                                max_num_seqs: int,
+                                seqs_max_tokens: list[int],
+                                prompts_lengths: list[int],
+                                steps_add_reqs: list[int],
+                                checked_steps: list[dict[str, Any]],
+                                auto_check_basic_decode_steps: bool = True):
```
yannicks1, line 378: Is `auto_check_basic_decode_steps` ever `False` or does it always default to `True`? If so, can we remove it or do you want to parametrize it and check both cases?

sducouedic: Thanks for reviewing Yannick. It is True for the existing tests so far, but when True in imposes constraints on the `checked_steps` (every prefill + first decode after the prefill needs to be explicitely added to `checked_steps`). I was thinking this could be annoying for the future tests.

PR #159, opened 2025-05-15
[Docs] Add overview and examples

rafvasq:
- Adds a starter `user_guide/spyre_overview` doc with links to recent blogs and video
- Adds `getting_started/examples` for code under `/examples`via script from vLLM add them automatically
- Small reorg of examples into folders
- Adds [myst_url_schemes](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#customising-external-url-resolution) to make github links resolve nicely (also ported from vLLM repo)

This serves at least as a placeholder for now until more Spyre content is made available.

PR #161, opened 2025-05-15
📝 docs for local development on CPU

prashantgupta24:
Add documentation on how to run the `eager` (CPU) examples + tests on `arm64` and `x86_64`.

See it live - https://vllm-spyre--161.org.readthedocs.build/en/161/getting_started/local_development.html

~Will wait for https://github.com/vllm-project/vllm-spyre/pull/159 to merge first~


docs/source/getting_started/local_development.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Run CB example
+
+Note: You will have to change the `model-name` in the example file before running it.
+
+```sh
+HF_HUB_OFFLINE=1 python examples/offline_inference_spyre_cb.py
+```
+
+## Debugging using debugpy
```
rafvasq, line 144: I'm not sure if including debugpy here is necessary, since it's more of a general topic and not spyre. It could be linked to (https://code.visualstudio.com/docs/python/debugging) in another debugging doc we end up having but I don't think it's needed here.

docs/source/getting_started/local_development.md
```diff
@@ -0,0 +1,171 @@
...
+   uv sync --frozen
+   ```
+  
+   or also with lint:
+  
+   ```sh
+   uv sync --frozen --group lint
+   ```
+  
+   **Note: `--group dev` is enabled by default**
```
rafvasq, line 49: You could use nice-looking ["admonitions"](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#admonition-types) for notes with more context like this one.

docs/source/getting_started/local_development.md
prashantgupta24: Yeah might be too technical, although it's a good thing to have a reference for somewhere. Let me think about where to add it

prashantgupta24:
> Overall, I think that this doc could get split up a bit. 🤔
> 
> * For example, the installation part could be part of the general [Installation doc](https://vllm-spyre.readthedocs.io/en/latest/installation.html) but denoting that it's different way to install (for local dev on mac).
> * Running the tests could be a section in the `Contributing` guide. See https://docs.vllm.ai/en/stable/contributing/overview.html#developing for inspiration.
> * Running the examples are a bit straightforward and I'm not really sure they require extra documentation except maybe in a future quickstart doc or in the Contributing guide too?

Thanks for the review Rafael! 

1. Yeah I was avoiding touching the `installation` file until you made some changes. I also don't understand the difference between installing `spyre` via `uv sync` (like in this PR) vs using `uv pip install -e` (like in `installation.md`). As far as I can tell, both are installing a local editable version of `spyre` in the venv.
2. For the tests I agree they could be in the contributing section. Then we could eventually add other variations of running the tests, for example, on the CI or on a machine with an actual spyre card.
3. The thing with examples is that there is some caveat of running the examples on a local machine (like setting the HF hub offline variable) so that's why I had it in this section. Possibly I can move that in the quickstart doc and add a note

docs/source/getting_started/installation.md
```diff
@@ -5,12 +5,52 @@ installation of the plugin and its dependencies. `uv` provides advanced
...
+1. Activate it
+  
+   ```sh
+   source .venv/bin/activate
+   ```
+
+1. Install `vllm-spyre` locally with dev (and optionally lint) dependencies
+  
+   ```sh
+   uv sync --frozen --active
```
joerunde, line 36: `VLLM_TARGET_DEVICE=empty` needs to be set before building the project, right?

docs/source/getting_started/installation.md
```diff
@@ -5,12 +5,52 @@ installation of the plugin and its dependencies. `uv` provides advanced
...
+   ```
+  
+   or also with lint:
+  
+   ```sh
+   uv sync --frozen --group lint
```
joerunde, line 42: I think `--inexact` can also be used here so that it doesn't cull out other things like `xgrammar` and `torch`

docs/source/getting_started/installation.md
prashantgupta24: I have never set that variable but it still works for me though.

docs/source/getting_started/installation.md
prashantgupta24: Yeah I thought about it, since it's a fresh environment I was assuming that there is nothing installed yet. But I guess it's safe to add it though

prashantgupta24: Although I think this makes sense for the lint PR too? 

https://github.com/vllm-project/vllm-spyre/pull/175/files What do you think?

docs/source/getting_started/installation.md
prashantgupta24: Can you try it out?

joerunde (approved):
let's get it 👍

PR #162, opened 2025-05-16
Support both paged and non-paged attention

yannicks1:
### Final FMS API supporting paged and non-paged attention

This final Fms API will support both the static and continuous batching APIs. Currently consuming [this](https://github.com/foundation-model-stack/foundation-model-stack/tree/paged_attn_attn_kwargs_impl) fms branch. Only minimal changes required on vLLM Spyre side. 

Note: Do not merge until in fms main.

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -578,7 +578,13 @@ class ContinuousBatchingSpyreModelRunner(SpyreModelRunner):
         self.tkv: int = 0
         # set self.free_blocks to the minimal value of 4 required for warmup
         # is reset to the value returned by the Spyre compiler after warmup
-        self._set_free_blocks(num_blocks=4)
+        # self._set_free_blocks(num_blocks=4)
+        # for the time being we set this to num_blocks consistent with the
+        # cache dimension of ContinuousBatchingFmsModel.past_key_value_states
+        num_blocks = (vllm_config.scheduler_config.max_num_seqs *
+                      vllm_config.model_config.max_model_len //
+                      self.BLOCK_SIZE)
+        self._set_free_blocks(num_blocks=num_blocks)
         self.dummy_req_ids2blocks: list[int] = []
 
         # TODO: Remove this once we can prefill and decode


.github/workflows/test.yml
prashantgupta24, line 1: Adding it here just has an FYI - we should remove the CB custom flag altogether on this branch

prashantgupta24 (approved):

joerunde:
bot:test
TEST_FILE=tests/e2e/test_spyre_online.py

prashantgupta24:
new test incoming - `test_cb_max_tokens` - We will need to uncomment the v1 marker on this test.

tests/e2e/test_spyre_cb.py
prashantgupta24, line 1: new test incoming - `test_cb_max_tokens` - We will need to uncomment the v1 marker on this test.

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -89,8 +90,10 @@ def compile_or_warm_up_model(self) -> None:
             logger.info(
                 "Warming up for prompt length %d, decoding %d tokens with "
                 "batch size %d", prompt_len, num_decode_tokens, batch_size)
-            self._warmup_spyre_fixed_size(prompt_len, num_decode_tokens,
-                                          self.restricted_tokens, batch_size)
+            with _maybe_warmup_context():
```
sducouedic, line 93: maybe the `with _maybe_warmup_context()` can used one level up before the for loop, instead of importing warmup_mode from torch_sendnn for each warmup shape. Maybe not that important since we don't warmup for many shapes usually.

sducouedic (approved):
LGTM

vllm_spyre/v1/worker/spyre_worker.py
joerunde: I think it needs to be done on each shape here so that the context closes and runs `update_lazyhandle()` after each shape, right?

sducouedic: aah ok you are probably right, I don't really understand what is happening with that context to be honest.
btw, I was looking for `update_lazyhandle` in the code to understand, and couldn't find it, even with a general "find" in the directory. When did that disappear?

tjohnson31415: The call to `update_lazyhandle` is now hidden in the `warmup_mode` context provided by `torch_sendnn` such that it gets called when the context exits. Use of `warmup_mode` was part of the change to remove the deprecated `sendnn_decoder` backend in https://github.com/vllm-project/vllm-spyre/pull/186.

So you'd have to look in the `torch_sendnn` code, but the `warmup_mode` context is basically just:
```
class warmup_mode:
    def __init__(self, enabled: bool = True):
        global _warmup_mode
        self._old_mode = _warmup_mode
        _warmup_mode = enabled

    def __enter__(self):
        return None

    def __exit__(self, exc_type, exc_val, exc_tb):
        global _warmup_mode
        update_lazyhandle()
        _warmup_mode = self._old_mode
 ```

File: vllm_spyre/v1/worker/spyre_worker.py [hunk 2]
```diff
@@ -63,7 +63,8 @@ def compile_or_warm_up_model(self) -> None:
         """Prepare model for execution through compilation/warmup."""
         # TO DO: implement warmup for continuous batching
```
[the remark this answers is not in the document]
yannicks1: no, should be fine, I will remove it

joerunde:
Merging now that we have driver changes to support this 🎉

yannicks1:
nice to have that finally in main !

PR #163, opened 2025-05-16
:fire: remove supported models

joerunde:
Aligning with testing on spyre hardware. We'll update this soon with more details of which models have been verified to work


rafvasq (approved):
🫡

PR #164, opened 2025-05-16
✨ add doc lint fix

prashantgupta24:
Adds a `fix` option for markdown linter. Works okay in most cases. Docs for it - https://pymarkdown.readthedocs.io/en/latest/user-guide/#fix-mode-failure-correction

Note: changes to the `uv.lock` file are big because uv introduced a big change here - https://github.com/astral-sh/uv/pull/13176 . This changes all occurrences of `upload_time` to `upload-time` . Apart from that, the only actual change should be the version update for the doc linter.


rafvasq (approved):
LGTM! I think it'll help while we're at the very start of adding docs.

PR #170, opened 2025-05-16
[Docs] 🐛 fix lint command

prashantgupta24:
`uv sync` will remove any existing packages that you might have in your local venv and replace them with the lint dependencies.

`uv pip install --group lint` will install the extra `lint` dependencies without removing anything that exists.


rafvasq (dismissed):
Nice! I think you could update the contributing doc too:

https://github.com/vllm-project/vllm-spyre/blob/main/docs/source/contributing/overview.md?plain=1#L14

joerunde (approved):
sure!

PR #171, opened 2025-05-19
[CI/CD] mark current tests that are failing when compile cache is enabled

sducouedic:
This PR mark the current tests that are currently failing when compile cache is enabled. It will allow to run the CI/CD pipeline with compile cache, while filtering out the tests that fail. 

Goes along with [this PR](https://github.ibm.com/ai-foundation/aiu-tests/pull/257)

### DIFF
--- a/tests/e2e/test_spyre_warmup_shapes.py
+++ b/tests/e2e/test_spyre_warmup_shapes.py
@@ -3,6 +3,8 @@
 Run `python -m pytest tests/test_spyre_warmup_shapes.py`.
 """
 
+import os
+
 import pytest
 from spyre_util import (compare_results, generate_hf_output,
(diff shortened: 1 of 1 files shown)


joerunde:
Thanks @sducouedic!

I just learned you can actually have pytest check a condition in the mark, so I updated this to skip if caching is enabled. That way we don't have to add more changes to our test suite when invoking the tests 🎉 !

I'll check this with our PR bot and then we can get this in for way speedier tests 🐎

joerunde:
Ah shoot, required manual testing because the pr bot can't enable caching 🤦 

But seems to work!

joerunde (approved):

sducouedic:
Nice looks great thanks

PR #172, opened 2025-05-19
[CB] refactoring spyre model runner

wallashss:
This PR does some refactoring primarily on spyre_model_runner. This changes tries to reduce code deduplication between static batching and continuous batching. However, the intention of this work will not be complete until a next PR has as goal remove kv cache manager from the spyre model runner.

Summary of changes:

- Reduce code deduplication in spyre model runner, some methods are common in `SpyreMoldeRunner` class, while `StaticBatchingSpyreModelRunner` and  `ContinuousBatchingSpyreModelRunner` override few of them to do their specific logic
- Changed `ContinuousBatchingFmsModel` class to get the attention metadata via forward context, and changed the model runner to pass to use the `with set_forward_context` to pass the attention metadata. This is the way vLLM does to support multiple attention backends [[REF](https://github.com/vllm-project/vllm/pull/10558)] 
- Moved the left pads to the CachedRequestState. 
- Bugfix: The `execute_model` in CB model runner was inconsistent with the data of input batch when it outputs the resul in `CBSpyreModelRunnerOutput`. Changed it with prepare_prompt to use the data of input batch.
- Misc: few renamed variables, more comments, and TODOs


joerunde:
We need to remove the `xfail` on the batch handling test here and ensure it passes

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24, line 1: So something in this PR makes the batch_handling work for continuous batching! Nice job 🥳 That means we can enable the batch handling test for CB, right?

prashantgupta24:
```
tests/e2e/test_spyre_basic.py::test_batch_handling[cb-eager-llama-194m] XPASS [ 27%]
```

prashantgupta24:
> We need to remove the `xfail` on the batch handling test here and ensure it passes

jinx

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -942,114 +997,49 @@ def execute_model(
...
-            is_prompt=model_input.is_prompt,
-            current_tkv_mask=model_input.current_tkv_mask,
-            left_padded_prompt_mask=model_input.left_padded_prompt_mask,
-            block_table=model_input.block_table,
-            slot_mapping=model_input.slot_mapping)
-
-        # Only perform sampling in the driver worker.
-        if not self.is_driver_worker:
+        if not output:
+            # TODO: Review this! see `base_execute_model`
```
prashantgupta24, line 1003: Earlier we used to return `output` as is and were not checking the empty condition. Was that a bug or can we still do that? (will investigate myself)

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -942,114 +997,49 @@ def execute_model(
...
-            torch._dynamo.mark_dynamic(model_input.current_tkv_mask, 0)
-            torch._dynamo.mark_dynamic(model_input.left_padded_prompt_mask, 0)
-
-            # sequence
-            torch._dynamo.mark_static(model_input.input_tokens, 1)  # always 1
-            torch._dynamo.mark_dynamic(model_input.block_table, 1)
-            torch._dynamo.mark_static(model_input.slot_mapping, 1)  # always 1
-            torch._dynamo.mark_static(model_input.input_positions,
-                                      1)  # always 1
+        output = self.base_execute_model(scheduler_output, **kwargs)
```
prashantgupta24, line 1000: Maybe I still haven't understood it fully, but could we not call `super().execute_model` here and rename the function in `class SpyreModelRunner` to `execute_model` ?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -251,6 +266,102 @@ def _update_states(self, scheduler_output: SchedulerOutput):
...
+        self.update_states(scheduler_output)
+
+        if not scheduler_output.total_num_scheduled_tokens:
+            # Return empty ModelRunnerOuptut if there's no work to do.
+            return EMPTY_MODEL_RUNNER_OUTPUT
+
+        model_input = self.prepare_model_input(scheduler_output)
+
+        # Execute the model
+        attn_metadata = self.build_attn_metadata(model_input)
```
prashantgupta24, line 317: Uhh, maybe I didn't understand it, but in the function `build_attn_metadata`, we don't really do anything with the `model_input`. So why are we passing it here?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -215,7 +217,20 @@ def complete_warmup(self):
         """Turn off warmup mode once the warmup is complete"""
         self.warmup_mode = False
 
-    def _update_states(self, scheduler_output: SchedulerOutput):
+    def build_attn_metadata(self,
+                            _: ModelForwardInputs) -> SpyreAttentionMetadata:
+
+        # TODO: probably sooner we will need
+        attn_metadata = SpyreAttentionMetadata()
+        return attn_metadata
```
prashantgupta24, line 225: Do we need to have this function defined in the superclass as well if we only need to use it for continuous batching?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -914,26 +959,36 @@ def pad_input_ids(
...
+            self, model_input: ModelForwardInputs) -> SpyreAttentionMetadata:
+
+        # TODO: probably we can remove some fields of the model input and
+        # update only the SpyreAttentionMetadata
+        attn_metadata = SpyreAttentionMetadata(
+            slot_mapping=model_input.slot_mapping,
+            current_tkv_mask=model_input.current_tkv_mask,
+            left_padded_prompt_mask=model_input.left_padded_prompt_mask,
+            block_table=model_input.block_table)
+        return attn_metadata
```
prashantgupta24, line 972: nit:
```suggestion
        return SpyreAttentionMetadata(
            slot_mapping=model_input.slot_mapping,
            current_tkv_mask=model_input.current_tkv_mask,
            left_padded_prompt_mask=model_input.left_padded_prompt_mask,
            block_table=model_input.block_table)
```

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24: Oh, okay, I have no idea about that. Let me read up on it.

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24: Yep! You also remove the fixture above the test

wallashss: Thanks! I think that I saw in an earlier version where it was marker and did not recognize the code now. I'll do it.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -827,33 +869,36 @@ def _prepare_decode(
...
+        offset = n_padded_blocks * self.block_size
 
         if offset > 0:
             logger.debug("Number of removed blocks due to left padding: %d",
                          n_padded_blocks)
 
-            for req_id in self.req_ids2left_pads:
-                self.req_ids2left_pads[req_id] -= offset
+            for req in requests:
+                req.left_padding -= offset
```
prashantgupta24, line 891: This is great! I wonder why we weren't doing this earlier instead of creating `req_ids2left_pads`

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24: Same question for the below two functions as well

prashantgupta24: Oh, now I get it. Static batching uses the default implementation defined in the base class. 🤦

wallashss:
@yannicks1 I'd appreciate if you could give me some feedback here. Thanks!

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -251,6 +266,104 @@ def _update_states(self, scheduler_output: SchedulerOutput):
...
+            logprobs=(output.logprobs_tensors.tolists()
+                      if output.logprobs_tensors else None),
+            prompt_logprobs_dict={
+                req_id: None
+                for req_id in self.input_batch.req_id_to_index
+            },  # TODO: take a decision regarding prompt logprobs
+            **extra_kwargs,
+        )
+
+        return model_output
```
prashantgupta24, line 365: nit
```suggestion
        return ModelRunnerOutput(
            req_ids=list(req_id_to_index.keys()),
            req_id_to_index=req_id_to_index,
            sampled_token_ids=output.sampled_token_ids.tolist(),
            spec_token_ids=None,
            logprobs=(output.logprobs_tensors.tolists()
                      if output.logprobs_tensors else None),
            prompt_logprobs_dict={
                req_id: None
                for req_id in self.input_batch.req_id_to_index
            },  # TODO: take a decision regarding prompt logprobs
            **extra_kwargs,
        )
```

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -229,7 +245,7 @@ def _update_states(self, scheduler_output: SchedulerOutput):
 
             # Update the cached states.
             num_computed_tokens = req_data.num_computed_tokens
-            req_state.num_computed_tokens = num_computed_tokens
+            # req_state.num_computed_tokens = num_computed_tokens
```
joerunde, line 248: Is this update no longer required? Should we delete the line?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -270,7 +383,7 @@ def __init__(
         self.spyre_warmup_shapes = SpyrePlatform.get_warmup_shapes(
             self.scheduler_config)
 
-    def _prepare_prompt(
+    def prepare_prompt(
```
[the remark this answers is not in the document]
wallashss: Moved back to set `_` prefix. Actually, for me these methods are _protected_, and I was not sure if I should use the same convention for private method.

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: Almost forgot this. Good catch. I commented this line because I fixed the type annotation of self.requests:

```
CachedRequestData -> CachedRequestState
```

So, we do not have this property since the beginning of this implementation it was there by coincidence because of the wrong type. Then, yes we can delete this line.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -215,7 +217,20 @@ def complete_warmup(self):
...
 
-    def _update_states(self, scheduler_output: SchedulerOutput):
+    def build_attn_metadata(self,
+                            _: ModelForwardInputs) -> SpyreAttentionMetadata:
+
+        # TODO: probably sooner we will need
+        attn_metadata = SpyreAttentionMetadata()
+        return attn_metadata
+
+    def get_sampling_metadata(self, model_input: ModelForwardInputs):
```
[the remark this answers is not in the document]
joerunde: Also this should have a type hint
```
def get_sampling_metadata(self) -> SamplingMetadata:
```

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -360,12 +478,19 @@ def _prepare_decode(
...
-        return ModelForwardInputs(
+        model_input = ModelForwardInputs(
             input_tokens=input_tokens,
             input_positions=self._position_ids,
             input_masks=self._mask,
             is_prompt=False,
         )
+        self._mark_input_tensors(model_input)
+
+        # TODO: Added here temporarily until we can remove dummy token
```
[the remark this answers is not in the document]
wallashss: Sorry, for that. I updated the comment. In CB we update this indices in different ways, most of cases we could only get from the input batch as we do in static batch. However the input_batch is not awareness of the dummy token, so currently it will not work for batch_size=1. The idea is, once we can do that, we could move this line to the execute_model in the base class just once and remove this updates of the specialized classes.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -607,23 +629,23 @@ def __init__(
             vocab_size=vllm_config.model_config.get_vocab_size(),
         )
 
+        self.decode_batch = self.input_batch
```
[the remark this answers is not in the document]
wallashss: Thanks for pointing out, this was an idea that I dropped and forgot to remove this.

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: Yes, I am aware, but my intention is to expand the scheduler as we already do to set this logic there. Currently, we are somehow hacking/bypassing the kv cache manager to compute something that we do not use. So, I believe we could at least replace some useless computation with something useful and try to keep the model runner clean. Once vLLM + compiler support better our case, we can remove the hacking code.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -942,114 +989,46 @@ def execute_model(
...
-                      if output.logprobs_tensors else None),
-            prompt_logprobs_dict={req_id: None
-                                  for req_id in req_ids
-                                  },  # TODO(wallas?): prompt logprobs too
-            tkv=self.tkv,
-            **extra_kwargs,
-        )
-        return model_output
+
+def mark_input_tensors_for_cb(model_input: ModelForwardInputs) -> None:
```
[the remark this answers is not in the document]
wallashss: I'm sorry, but I didn't understand what you mean. Is that a suggestion or just a question?

I removed this from the class because of the comment in #198. There isn't any usage of self there, so it's kinda make sense to not be there. But for static batching, self is needed and it is used to mark some model tensors as static/dynamic.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -751,27 +777,36 @@ def _prepare_decode(
                                         dtype=torch.bool,
                                         device="cpu")
 
-        for cached_request in cached_requests:
+        assert len(self.input_batch.req_id_to_index) == len(cached_requests)
+        # TODO(wallas): I think we can do better here, without sorting and
+        # creating a intermediary dictionary
+        cached_reqs_map = {c.req_id: c for c in cached_requests}
+        req_ids = self.input_batch.sorted_requests_ids
+
```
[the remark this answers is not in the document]
wallashss: Thanks for the suggestion @yannicks1 . 

However `self.input_batch.sorted_requests_ids` is property and it does a sort. So, in a loop it is not a good idea, we will get O(N^2).

But the context here: this is the change that fix a bug caused by an inconsistency of request id between the input batch and the model runner output. The challenge is that we need the information of cached_requests (which is a list) and we need to build input_tokens, input_positions, slot_mapping based on the indices calculated by the input batch. That's the why I sort the indices of requests, and build the map to access the cached_request that contains updated block ids to get it during the loop.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -914,26 +950,34 @@ def pad_input_ids(
...
+        # TODO: probably we can remove some fields of the model input and
+        # update only the SpyreAttentionMetadata
+        return SpyreAttentionMetadata(
+            slot_mapping=model_input.slot_mapping,
+            current_tkv_mask=model_input.current_tkv_mask,
+            left_padded_prompt_mask=model_input.left_padded_prompt_mask,
+            block_table=model_input.block_table)
+
+    def get_sampling_metadata(self, is_prefill: bool) -> SamplingMetadata:
```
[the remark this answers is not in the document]
wallashss: I really thought about that... see this comment
https://github.com/vllm-project/vllm-spyre/pull/172#discussion_r2169551075

But I dropped the idea because I was worried of the 'hiding' be a source of bugs/undefined behaviors. On one hand, this would be a cleaner implementation for now, but on the other hand I feel that would be a trap latter if we don't take an extra care of this implementation. That's the why I decided to implement something more explicit, to let others be aware of this behavior. I always think twice when I add something that will be a side effect in the flow of data. 

> BTW: The first time I wrote the prefill batch, I was already concerned if this would be a good idea or not, but it was simplest and fastest solution at that moment.

However, I had other idea that would make more changes but I think we would get a good balance between clean implementation + explicit behavior. I think that we could get good results if we change the `execute_model` to something like that:

```py
# Model input will have more properties

(...)

# Sample the next token.
    logits=logits,
    sampling_metadata=model_input.sampling_metadata, # CHANGE HERE
)

(...)
req_id_to_index = model_input.req_id_to_index # CHANGE HERE

model_output = ModelRunnerOutput(
            req_ids=list(req_id_to_index.keys()),
            req_id_to_index=req_id_to_index,
            sampled_token_ids=output.sampled_token_ids.tolist(),
            spec_token_ids=None,
            logprobs=(output.logprobs_tensors.tolists()
                      if output.logprobs_tensors else None),
            prompt_logprobs_dict={
                req_id: None
                for req_id in self.input_batch.req_id_to_index
            },  # TODO: take a decision regarding prompt logprobs
            **extra_kwargs,
        )
```

So, we add these extra fields to model_input (I guess we would rename at least the class name which is `ModelForwardInputs`, I would like suggestions 😄 ) which are filled by `_prepare_prompt` and `_prepare_decode` and each specialized class will know how to proper do it. 

WDYT?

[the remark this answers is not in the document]
tjohnson31415: I was wondering about that comment on `self.decode_batch`!

Yeah, good points. This code is pretty hard to follow already... I also didn't notice that `update_states` interacts with `self.input_batch`, so my proposal would require us re-setting `input_batch = self.decode_batch` at the end of each execution...

I do like the idea of making things more explicit instead (in a follow up PR). Not sure what to call the "blob of things we need during execute_model" 😅

vllm_spyre/v1/worker/spyre_model_runner.py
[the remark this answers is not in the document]
yannicks1: oh, just a renaming. 
we have a function doing the same thing in the static batching model runner. we call it `_mark_input_tensors` there. we could use the same naming convention here e.g. `_mark_input_tensors` instead of `mark_input_tensors_for_cb` . Of course self does not need to be there...

wallashss:
Thanks everyone for the reviews! Anyone else have more comments? 

Can I get another stamp? 😄

tjohnson31415 (approved):

vllm_spyre/v1/worker/spyre_model_runner.py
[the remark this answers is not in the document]
yannicks1: @wallashss , I don't see where the O(N^2) is coming from? 
My suggestion was merely to write it a bit nicer e.g. 
```
req_id_map = {req_id: idx for idx, req_id in enumerate(self.input_batch.sorted_requests_ids)}
cached_requests = sorted(cached_requests, key=lambda req: req_id_map[req.req_id])

for cached_request in cached_requests:
    # do the same stuff
```

over all time complexity of sorting (twice, once in the first line and once in the second line) should still be `O(N log(N))`, not? Is the same as your approach (and should not be any bottleneck considering N is the batch size)...

PR #173, opened 2025-05-19
Add ppc64le inference script

Daniel-Schenker:
Adding a script that is used on ppc64le for running online and offline inference with vllm.

### DIFF
--- a/examples/spyre_inference_ppc64le.py
+++ b/examples/spyre_inference_ppc64le.py
@@ -1,0 +1,279 @@
+#!/usr/bin/env python3
+from openai import OpenAI, APIConnectionError
+
+import glob
+import os
+import pathlib
+import asyncio
+import argparse
+import gc
+import time
+import threading
+import numpy
+import signal
+import io
+import subprocess
+import sys
+import atexit
+
+# Bump if you make changes to the log file.
+VERSION = 1
+
+# Log the interesting blips to logbuf, and tee to stdout
+def log(string):
+    global logbuf
+    print(string)
+    print(string,file=logbuf)
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-d', '--debug', action='store_true')
+parser.add_argument('-l', '--online', action='store_true')
+parser.add_argument('-r', '--remote', action='store_true')
+parser.add_argument('-m', '--model', default="/models/granite-3.0-8b-instruct-r241014a/")
+parser.add_argument('-t', '--maxtoken', type=int, default=64)
+parser.add_argument('-n', '--promptlen', type=int, default=256)
+parser.add_argument('-i', '--prompts', default="/home/senuser/aiu-fms-testing-utils/tests/resources/prompts/granite/input")
+parser.add_argument('-p', '--port', type=int, default=8000)
+parser.add_argument('-o', '--output', default="./")
+parser.add_argument('-b', '--batch_size', type=int, default=1)
+parser.add_argument('-k', '--keepoutput', action='store_true')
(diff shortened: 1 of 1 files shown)


examples/spyre_inference_ppc64le.py
```diff
@@ -0,0 +1,323 @@
...
+import signal
+import subprocess
+import sys
+import threading
+import time
+
+import numpy
+from openai import APIConnectionError, OpenAI
+
+# Bump if you make changes to the log file.
```
joerunde, line 19: what does this mean?

examples/spyre_inference_ppc64le.py
```diff
@@ -0,0 +1,323 @@
...
+    print(string)
+    print(string, file=logbuf)
+
+
+def cleanup_pserver():
+    pserver.terminate()
+    pserver.wait()
+
+
+parser = argparse.ArgumentParser()
```
joerunde, line 35: Can you stick the `description` field in here with a quick summary of what this script is for / what it does?

examples/spyre_inference_ppc64le.py
```diff
@@ -0,0 +1,323 @@
...
+parser.add_argument('-d', '--debug', action='store_true')
+parser.add_argument('-l', '--online', action='store_true')
```
joerunde, line 37: help strings on these would be super great!

examples/spyre_inference_ppc64le.py
```diff
@@ -0,0 +1,323 @@
...
+# For offline inferencing.
+os.environ["VLLM_SPYRE_WARMUP_NEW_TOKENS"] = str(max_tokens)
+os.environ["VLLM_SPYRE_WARMUP_BATCH_SIZES"] = str(batch_size)
+os.environ["VLLM_SPYRE_WARMUP_PROMPT_LENS"] = str(promptlen)
+
+# Verify setup is consistent with number of aius
+if not args.remote:
+    if "VLLM_AIU_PCIE_IDS" not in os.environ or os.environ[
+            "VLLM_AIU_PCIE_IDS"] == "":
+        print("source vllm_start.sh before running scripti."
```
joerunde, line 78: vllm_start.sh doesn't exist externally here :/

Can we update this to instead mention `/etc/profile.d/ibm-aiu-setup.sh`, which is available in the public base image?

examples/spyre_inference_ppc64le.py
```diff
@@ -0,0 +1,323 @@
...
+
+nextid = 0
+logname = logname_pfx + str(nextid) + ".txt"
+while pathlib.Path(logname).exists():
+    nextid += 1
+    logname = logname_pfx + str(nextid) + ".txt"
+
+pathlib.Path(logname).write_text(logbuf.getvalue())
+
+# TODO: multi-aiu doesn't exit. See aiu-app-sw-tracker#246
```
joerunde, line 317: let's not put internal issue links out here in the public

examples/spyre_inference_ppc64le.py
```diff
@@ -0,0 +1,323 @@
...
+                              sampling_params=llm_sparms,
+                              use_tqdm=False)
+        tt = time.perf_counter() - start
+
+        test_output = [
+            prompt[i] + output[i].outputs[0].text for i in range(len(output))
+        ]
+
+        # From the innards of vllm, hopefully accurate
+        end_ft = output[0].metrics.first_token_time - output[
```
joerunde, line 219: Do these metrics work with vllm v1 as well? I think they were dropped from some of the python apis during the move

PR #174, opened 2025-05-20
Disable compile cache in test_spyre_warmup_shapes

sducouedic:
Disable compile cache in `test_spyre_warmup_shapes` because this test fails with cache enabled. Need to debug


waleedqk (approved):
Looks good
we only want this test to skip caching currently, and the CI pipeline can run as is without updates required

PR #175, opened 2025-05-21
🎨 use sync with inexact instead?

prashantgupta24:
Use `--inexact` instead to preserve the existing packages already installed?

### DIFF
--- a/.github/workflows/reminder_comment.yml
+++ b/.github/workflows/reminder_comment.yml
@@ -15,7 +15,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.issue.number,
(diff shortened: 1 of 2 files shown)


prashantgupta24:
cc @joerunde

joerunde:
hmm, I'm going back and forth on adding `--active`, since you'd usually be using uv to manage the project venv. (Which would probably also be the one you have active, but might not be)

Pretty low care amount either way though

joerunde (approved):

prashantgupta24:
> hmm, I'm going back and forth on adding `--active`, since you'd usually be using uv to manage the project venv. (Which would probably also be the one you have active, but might not be)
> 
> Pretty low care amount either way though

Yeah, I'm not sure what prompted me to add the active flag but in one scenario that I was testing, it was needed. Doesn't hurt to have it though, right?

PR #177, opened 2025-05-21
🐛 ignore shell script files within .venv for shellcheck command

prashantgupta24:
Allowing the `shellcheck` command to run against all shell script files within `.venv` gives the error: 

```
xargs: command line cannot be assembled, too long
```

This PR only checks the shell scripts that are git committed.

After ignoring:
```
❯ git ls-files -z | grep -z '\.sh$' | xargs -0 -n 1 printf "%s\n"

format.sh
tools/actionlint.sh
tools/check_repo.sh
tools/doc-lint.sh
tools/mypy.sh
tools/shellcheck.sh
```

Before:
```
❯ find . -name "*.sh"
./tools/shellcheck.sh
./tools/doc-lint.sh
./tools/check_repo.sh
./tools/actionlint.sh
./tools/mypy.sh
./.venv/lib/python3.12/site-packages/bleach/_vendor/vendor_install.sh
./.venv/lib/python3.12/site-packages/debugpy/_vendored/pydevd/pydevd_attach_to_process/linux_and_mac/compile_mac.sh
./.venv/lib/python3.12/site-packages/debugpy/_vendored/pydevd/pydevd_attach_to_process/linux_and_mac/compile_linux.sh
./.venv/lib/python3.12/site-packages/pexpect/bashrc.sh
./.venv/lib/python3.12/site-packages/tqdm/completion.sh
./.venv/lib/python3.12/site-packages/ray/autoscaler/aws/cloudwatch/ray_prometheus_waiter.sh
./format.sh
```

### DIFF
--- a/tools/shellcheck.sh
+++ b/tools/shellcheck.sh
@@ -18,5 +18,8 @@ if ! [ -x "$(command -v shellcheck)" ]; then
     export PATH="$PATH:$(pwd)/shellcheck-${scversion}"
 fi
 
-# TODO - fix warnings in .buildkite/run-amd-test.sh
-find . -name "*.sh" -not -path "./.buildkite/run-amd-test.sh" -not -path "./.venv/*" -print0 | xargs -0 -I {} sh -c 'git check-ignore -q "{}" || shellcheck "{}"'
+# Lists all files currently tracked by Git
+# outputs results separated by a null character
+# matches any filename ending with .sh
+# run shellcheck on every .sh file tracked by Git.
+git ls-files -z | grep -z '\.sh$' | xargs -0 -n 1 shellcheck
--- a/tools/shellcheck.sh
+++ b/tools/shellcheck.sh
@@ -19,4 +19,4 @@ if ! [ -x "$(command -v shellcheck)" ]; then
 fi
 
 # TODO - fix warnings in .buildkite/run-amd-test.sh
-find . -name "*.sh" -not -path "./.buildkite/run-amd-test.sh" -print0 | xargs -0 -I {} sh -c 'git check-ignore -q "{}" || shellcheck "{}"'
+find . -name "*.sh" -not -path "./.buildkite/run-amd-test.sh" -not -path "./.venv/*" -print0 | xargs -0 -I {} sh -c 'git check-ignore -q "{}" || shellcheck "{}"'


tools/shellcheck.sh
```diff
@@ -19,4 +19,4 @@ if ! [ -x "$(command -v shellcheck)" ]; then
 fi
 
 # TODO - fix warnings in .buildkite/run-amd-test.sh
-find . -name "*.sh" -not -path "./.buildkite/run-amd-test.sh" -print0 | xargs -0 -I {} sh -c 'git check-ignore -q "{}" || shellcheck "{}"'
+find . -name "*.sh" -not -path "./.buildkite/run-amd-test.sh" -not -path "./.venv/*" -print0 | xargs -0 -I {} sh -c 'git check-ignore -q "{}" || shellcheck "{}"'
```
prashantgupta24, line 22: Another option is to increase the amount of space for the command which is pointed out in this link"

https://stackoverflow.com/a/77336067
```
Specify the amount of space (in bytes) that -I can use for
             replacements.
```
but given that we are not aware of how many shell scripts might exist in the virtual environment, I feel ignoring the whole environment altogether might be a safer approach.

prashantgupta24: Ideally, we would want to ignore all directories that were git ignored. But I'm not sure if it's worth the effort?

joerunde: I agree we should be ignoring .venv entirely, and any other common names for directories that would be .gitignored.

Can we plop these into a .shellcheckrc file in the repo? https://github.com/koalaman/shellcheck/wiki/Directive#shellcheckrc-file

prashantgupta24: From the documentation it looks like the `shellcheckrc` file allows us to specify a source-path and not ignore directories though right. Also since shellcheck gets files as an argument, we should ideally be filtering the files ourselves, right? See this issue https://github.com/koalaman/shellcheck/issues/2905

Also the problem right now is that the command that we're trying to send to xargs is too long. So even if we somehow add directories as source paths within the `shellcheckrc` file, the error might still remain? 

Also I just noticed we have a `.shellcheckrc` file already in this repo :)

prashantgupta24: I guess we could go the other way around and actually list all files using git that are shell script files and pass them to shellcheck?

wallashss: Interesting solution! Thanks for investigating that, this was bothering me for a while.

However, it still fail for me. I am using other path for virtual environment, and set to ignore this was the solution to make it work.

prashantgupta24: > and set to ignore this was the solution to make it work.

I didn't understand, can you rephrase it? What exactly work for you?

wallashss: Sure, sorry for that. 

I mean that I have virtual environment in the path at `.pyenv`, so I had to place a `-not -path './.pyenv/*'` for the shellcheck run fine.

wallashss: Probably the way we collect the shell scripts with `find . -name "*.sh"` is too sensitive to the working dir of the repo. I also added other scripts there that I don't commit.

prashantgupta24: oh try this?
```
git ls-files -z | grep -z '\.sh$' | xargs -0 -n 1 printf "%s\n"

```

wallashss: I got something like this for `git ls-files --cached --others --exclude-standard -z | grep -z '\.sh$'` (removed the xargs part):

```
jumpbox.shlinting.shservedocs.shformat.shtools/actionlint.shtools/check_repo.shtools/doc-lint.shtools/mypy.shtools/shellcheck.sh
```

It includes other files like you said. But is this format ok? I would expect a delimiter between them.

wallashss, line 22: BTW, we do not have a .buildkite folder, neither the ignored script file
```suggestion
find . -name "*.sh"  -not -path "./.venv/*" -print0 | xargs -0 -I {} sh -c 'git check-ignore -q "{}" || shellcheck "{}"'
```

wallashss: I think there's an extra dquotes in the the end of this commnad, I removed it and did run. 
if it is to the process return 0 without any printing then It runs fine .

wallashss: I guess so

prashantgupta24: Whoops, yeah, my bad, I edited the command. And yeah, I also tried testing it with an actual incorrect line within the shell script and it seems to catch it fine. And once I remove it, it passes the check. So I guess we're good.

prashantgupta24: I guess if you're going down the other path of fetching files through `git` then we don't really need to care about this change then.

wallashss (approved):
LGTM! Thanks for this!

Issue #148, opened 2025-05-08
Incorrect default max_completion_tokens being set


gkumbhat:
### Changes

Trying to run vLLM with `VLLM_SPYRE_WARMUP_NEW_TOKENS` set to X number always results in error:

```
No applicable warmup shape exists for combination of prompt length (Z tokens) and maximum number of output tokens to be generated (Y tokens)
```

Where `Z`  is input token count and `Y` is `Z + VLLM_SPYRE_WARMUP_NEW_TOKENS`

tjohnson31415:
To add some more context: This is observed when using the Chat API while not setting `max_completions_token` or `max_tokens` in the request body.

Turns out that the OpenAI frontend code is where the default value for `max_completions_token`/`max_tokens` is set.
For the completions API, it defaults to 16 ([REF](https://github.com/vllm-project/vllm/blob/a944f8ede7361a5233e112a575ff77c4aaa268a5/vllm/entrypoints/openai/protocol.py#L747)), so is typically overwritten in the incoming request.
For the chat API, it defaults to `max_model_len - prompt_len` ([REF](https://github.com/vllm-project/vllm/blob/a944f8ede7361a5233e112a575ff77c4aaa268a5/vllm/entrypoints/openai/serving_chat.py#L213-L214)). The setting of the default max-tokens happens before the execution gets to any vllm-spyre code 🤔

gmarinho2:
Since the default value for generation tokens is set in the OpenAI frontend code, the only fix seems to involve making a PR to vllm source code:

If a new method is implemented in the class `Plataform` in `vllm/platforms/interface.py`:
```python
    def maybe_update_default_max_tokens(self, prompt_len: int, default_max_tokens: int) -> int:
        return default_max_tokens
```


It can be override in `vllm_spyre/platform.py`:
```python
    @classmethod
    def maybe_update_default_max_tokens(self, prompt_len: int, default_max_tokens: int) -> int:
        max_new_tokens = 1
        for shape in self._warmup_shapes:
            if prompt_len <= shape['prompt_length']:
                max_new_tokens = max(max_new_tokens, shape['new_tokens'])
        
        return max_new_tokens
```

Now, in `vllm/entrypoints/openai/serving_chat.py`, in the class `OpenAIServingChat`, instead of using:
```python
     default_max_tokens = self.max_model_len - len(engine_prompt["prompt_token_ids"])
```


use:
```python
[...]

PR #179, opened 2025-05-22
Add maybe_update_max_tokens for class SpyrePlatform 

gmarinho2:
FIX: https://github.com/vllm-project/vllm-spyre/issues/148

Currently the `class SpyrePlatform` uses the default `max_tokens` set in the OpenAI frontend code. This new method selects the warmup shape that fits the prompt and has the biggest `shape['new_tokens']`. This way `SpyrePlatform` will use the maximum number possible for token generation when `max_tokens` is not set in the request body.

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -276,3 +276,12 @@ class SpyrePlatform(Platform):
             if prompt_len <= shape['prompt_length']
             and max_tokens <= shape['new_tokens']
         ]
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


wallashss:
I think you should probably set this PR as draft and wait for the feedback of the vllm community on https://github.com/vllm-project/vllm/pull/18557. This changes depends on that.

maxdebayser (approved):
LGTM, once the upstream PR gets merged.

PR #181, opened 2025-05-26
[CI] Ignore docs for tests

wallashss:
Accordingly to this documentation: https://docs.github.com/es/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpaths. This should be the way to not run tests when we only update docs.

```
on:
  push:
    paths-ignore:
      - 'docs/**'
```

### DIFF
--- a/.github/workflows/lint_code.yml
+++ b/.github/workflows/lint_code.yml
@@ -7,6 +7,8 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - 'docs/**'
   push:
     branches:
       - main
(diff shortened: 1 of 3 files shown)


wallashss:
@joerunde @ckadner 

Regarding this comment:

```
# Don't use pull_request.paths filter since this workflow is required for
  # all pull requests on main irrespective of file type or location
```

Is it okay use `ignore-paths` then? Am I understand correctly, or I may be missing something?

prashantgupta24:
And as far as I can tell, the `paths-ignore` should be the right way to do what you've added here.

.github/workflows/lint_code.yml
prashantgupta24, line 1: And as far as I can tell, the paths-ignore should be the right way to do what you've added here.

PR #182, opened 2025-05-27
[CB] add min batch size of 2 in decode

nikolaospapandreou:
This PR adds support for minimum batch size of 2 in decode steps.


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -612,13 +613,26 @@ def _update_states(self, scheduler_output):
...
         self,
         new_requests: list[NewRequestData],
     ) -> ModelForwardInputs:
         assert len(new_requests) > 0
         input_token_list: list[torch.Tensor] = []
 
+        # free the blocks used for padding to minimum decode batch size of 2
+        if self.dummy_req_ids2blocks:
+            for freed_block in self.dummy_req_ids2blocks:
+                self.free_blocks.appendleft(freed_block)
```
sducouedic, line 633: I think `appendleft` won't do exactly what you want. It will append the values from `dummy_req_ids2blocks` to the left, but in reverse order. e.g. `[6, 7] --> appendleft --> [7, 6, 8, 9, 10, ...]`. I would suggest to use `append` simply, just as you did in line 620. Or iterate the `dummy_req_ids2blocks` in reverse order (and do the same in 620)

sducouedic (approved):
LGTM (just a minor comment)
nice changes 👍

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -755,6 +769,27 @@ def _prepare_decode(
...
+            dummy_req_indices = torch.zeros(1, dtype=torch.bool, device="cpu")
+            self.model.indices = torch.cat(
+                (self.model.indices, dummy_req_indices), -1)
+            assert self.model.indices.size(dim=0) == 2
+
+            n = self.tkv + 1
+            d = self.BLOCK_SIZE
+            num_blocks = (n + d - 1) // d
+            for i in range(num_blocks - len(self.dummy_req_ids2blocks)):
+                self.dummy_req_ids2blocks.append(self.free_blocks.popleft())
```
joerunde, line 783: Is this popping free blocks for every decode iteration of a batch with a single request? It looks like `len(input_tokens) == 1` will always be true when there's only a single request running, since `cached_requests` comes from the scheduler and won't have our dummy request in it

joerunde: ahhh I see the `num_blocks - len(self.dummy_req_ids2blocks)` is only popping ones as the tkv increases 👍

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -612,13 +613,26 @@ def _update_states(self, scheduler_output):
...
     ) -> ModelForwardInputs:
         assert len(new_requests) > 0
         input_token_list: list[torch.Tensor] = []
 
+        # free the blocks used for padding to minimum decode batch size of 2
+        if self.dummy_req_ids2blocks:
+            for freed_block in self.dummy_req_ids2blocks:
+                self.free_blocks.appendleft(freed_block)
+            self.dummy_req_ids2blocks = []
+
```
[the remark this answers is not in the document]
sducouedic: I guess this would do the cleanup when the last requests finishes?

yannicks1:
> I am not completely sure if this won't automatically be handled correctly in the subsequent _prepare_decode, but probably in reduce_left_padding we should also free the dummy blocks

good point! this will be solved when doing the removal of the left padding in every step (decode **and** prefill). I am taking care of this on this branch: https://github.com/vllm-project/vllm-spyre/tree/ysc-refactor-left-padding-stripping

PR #184, opened 2025-05-27
[Docs] Update contributing docs

rafvasq:
- Adds information about opening issues/bug reports, PR linting, etc. and point to Slack support
- Remove debugpy doc because it doesn't seem vllm-spyre related (same example [here](https://code.visualstudio.com/docs/python/debugging#_example))
- Copy-edit testing docs

### DIFF
--- a/docs/source/contributing/overview.md
+++ b/docs/source/contributing/overview.md
@@ -1,4 +1,4 @@
-# Contributing to vLLM-Spyre
+# Contributing
 
 Thank you for your interest in contributing to Spyre support on vLLM!
 
--- a/docs/source/contributing/overview.md
+++ b/docs/source/contributing/overview.md
@@ -1,12 +1,92 @@
-# Contributing
+# Contributing to vLLM-Spyre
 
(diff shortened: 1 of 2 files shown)


docs/source/contributing/overview.md
```diff
@@ -1,150 +1,125 @@
...
 
-1. (optional) Install dev dependencies (if vllm-spyre was installed without them)
+1. (Optional) Install development dependencies:
   
    ```sh
    uv pip install --group dev
    ```
 
-1. Run the tests:
+1. Run the tests using [markers](https://github.com/vllm-project/vllm-spyre/blob/main/pyproject.toml#L119):
```
prashantgupta24, line 59: Can you do your docs magic and actually import the code from the file instead? The problem with using line numbers is that once the file changes then this link just points to something else entirely.

docs/source/contributing/overview.md
prashantgupta24, line 1: Love the changes. Thanks!

prashantgupta24 (approved):
LGTM!

PR #187, opened 2025-05-29
[FIX] Suppression of stacktrace on a shutdown

wallashss:
This PR adds a workaround to suppress logs that are dumped at the shutdown of the engine (only on V1) when vllm runs with multiprocess. The undesired behavior happens because g3log from Spyre stack/runtime overrides the signal  handler from vLLM when it starts a process for the engine code. Therefore, the engine does not have a chance to gracefully shutdown.

### DIFF
--- a/vllm_spyre/envs.py
+++ b/vllm_spyre/envs.py
@@ -10,6 +10,7 @@ if TYPE_CHECKING:
     VLLM_SPYRE_RM_PADDED_BLOCKS: bool = False
     VLLM_SPYRE_PERF_METRIC_LOGGING_ENABLED: int = 0
     VLLM_SPYRE_PERF_METRIC_LOGGING_DIR: str = "/tmp"
+    VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER: bool = False
 
 # begin-env-vars-definition
 
@@ -69,6 +70,12 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # logs are written to /tmp.
     "VLLM_SPYRE_PERF_METRIC_LOGGING_DIR":
     lambda: os.getenv("VLLM_SPYRE_PERF_METRIC_LOGGING_DIR", "/tmp"),
+
+    # If set, override the signal handler for vllm-spyre on
+    # vLLM V1 + torch_sendnn backend to be able to gracefully
+    # shutdown the engine.
+    "VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER":
+    lambda: os.getenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "0"),
 }
 
 # end-env-vars-definition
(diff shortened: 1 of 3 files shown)


wallashss:
bot:test
TEST_FILE=tests/e2e/test_spyre_basic.py MARKERS="spyre"

joerunde:
I've had a bit of experience messing with signal handlers, and the annoying problem is that setting a handler with `signal.signal(sig, handler)` simply overrides any handler that was previously registered for that signal. One thing I worked on back on caikit was a signal handler that would iteratively call all handlers registered for that signal, instead of just blanket overriding whatever the current handler is: https://github.com/caikit/caikit/blob/ce3fa2c129ce15a5e2095d466a8f01ec2e0c577d/caikit/runtime/server_base.py#L179
But of course, all handlers for a signal had to be added with that utility :(

Maybe we can use a similar strategy here though, to setup a handler that calls the vllm one, plus whatever is already registered:
```
def make_handler(signum):
    original_hander = signal.getsignal(signum)
    def handler(signal_, stack_):
        vllm.run_engine_core.signal_handler(signal_, stack_)
        original_hander(signal_, stack_)

    return handler

...

signal.signal(SIGTERM, make_handler(SIGTERM))
```

However, if we think that the one set by the g3log is not required for proper shutdown, then we should probably just override it to the vllm one and call it a day

vllm_spyre/envs.py
```diff
@@ -68,6 +69,12 @@
     # logs are written to /tmp.
     "VLLM_SPYRE_PERF_METRIC_LOGGING_DIR":
     lambda: os.getenv("VLLM_SPYRE_PERF_METRIC_LOGGING_DIR", "/tmp"),
+
+    # If set, override the signal handler for vllm-spyre on
+    # vLLM V1 + torch_sendnn backend to be able to gracefully
+    # shutdown the engine.
+    "VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER":
+    lambda: os.getenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "0"),
```
tjohnson31415, line 77: This is currently setting the value of `VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER` to the string "0" which is truthy.

```suggestion
    lambda: bool(int(os.getenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "1"))),
```

The default should be to have a graceful shutdown, this should be True by default.

vllm_spyre/envs.py
```diff
@@ -68,6 +69,12 @@
     # logs are written to /tmp.
     "VLLM_SPYRE_PERF_METRIC_LOGGING_DIR":
     lambda: os.getenv("VLLM_SPYRE_PERF_METRIC_LOGGING_DIR", "/tmp"),
+
+    # If set, override the signal handler for vllm-spyre on
+    # vLLM V1 + torch_sendnn backend to be able to gracefully
+    # shutdown the engine.
+    "VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER":
+    lambda: bool(os.getenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "1")),
```
tjohnson31415, line 77: `bool("0")` is True. We need to the `int()` conversion too to disable the override.

wallashss: oh, sorry. I applied the wrong patch.

tjohnson31415 (approved):
This PR took me down a bit of a rabbit hole on signal handling 😅 

Great fix though. LGTM

joerunde:
Ah, yeah TIL about the C++ vs Python signal handling 🤦 

LGTM!

PR #190, opened 2025-05-30
Simplify spyre_setup.py and fix distributed setup

tjohnson31415:
Most of the env vars can be handled automatically in the Sentient backend now, but it seems that flex requires `torchrun` style RANK and WORLD_SIZE env vars for configuring distributed serving.


tjohnson31415:
The test failure with CB is due to another upstream change to the `step()` API, unrelated to the changes in this PR.

tdoublep (approved):
LGTM - nice cleanup, and great that we have a fix for the TP issue.

PR #192, opened 2025-06-02
🐛 fix return type for update_from_output

prashantgupta24:
since upstream `update_from_output` now returns a dict 🙄


joerunde (approved):
Could be worth a comment explaining what the indexing on the dict is? But not super necessary since we're just delegating to the stock scheduler here

PR #193, opened 2025-06-02
[CI] Skip tests on doc changes

ckadner:
**WIP**

**Problem**:

For with PRs that only change *.md files is that:
- if we skip tests, the required PR check for all the tests will never report back to the PR and auto-merge won't be enabled, maintainers need to manually merge ([example PR](https://github.com/vllm-project/vllm-spyre/pull/194))
- if we don't skip tests, auto-merge will be enabled (if required tests succeed), but it takes 10 minutes for PR checks to complete

There are no elegant solutions AFAIK.

**Options**:

- live with either of the situation described above
- don't skip workflow runs, instead use GHA marketplace action to skip costly test steps based on changed files (this PR, creates a more cluttered workflow file)
- have 2 workflow files for each PR check with the same name, one dummy one that succeeds without running (harder to maintain)

If we don't merge this PR, or do similar changes, we could add more "committers" by updating the branch protection rules bypass list to enable forced merges on doc PRs

image


.github/workflows/test.yml
sducouedic, line 1: Maybe add an additional level of steps called something like "Install and Run Tests" which would include all the steps that are marked with the `if: steps.changed-src-files.outputs.any_changed == 'true'`? This would allow to avoid duplicating that line

ckadner: Unfortunately there are no "sub-steps" in GHA. I could create a composite action that "bundles" all the "Install and Run Test" steps into a separate file.
https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action#creating-a-composite-action-within-the-same-repository

This would keep the workflow file cleaner, but the test execution logs would be all in one long log file -- as opposed to now they are individual sections that can fail separately and collapsible, easier and more orderly to debug test failures.

sducouedic (approved):

PR #198, opened 2025-06-03
[SB][CB] mark input tensors refactoring

yannicks1:
WIP

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -432,6 +432,7 @@ class StaticBatchingSpyreModelRunner(SpyreModelRunner):
 
         t0 = time.time()
 
+        self._update_states(scheduler_output)
         # TODO: change to EMPTY_MODEL_RUNNER_OUTPUT, right now this
         # will be a breaking change, or clumsy to make retrocompatible
         # with conditional import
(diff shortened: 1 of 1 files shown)


wallashss:
FYI: I did something similar in this other draft PR. My intention is try have common between static batching and continuous batching and in after that have a structure that is more similar to vLLM upstream.

yannicks1:
thanks for letting me know (I usually do not check draft PRs:)
What do you think about [this](https://github.com/vllm-project/vllm-spyre/pull/198/commits/e6154be2e58e0bf7fc82c58efded5dc77a487012) change? I saw that the order is not the same for SB and CB and just changed it, test seem to pass...

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -916,41 +915,7 @@ def execute_model(
...
-            torch._dynamo.mark_dynamic(model_input.current_tkv_mask, 0)
-            torch._dynamo.mark_dynamic(model_input.left_padded_prompt_mask, 0)
-
-            # sequence
-            torch._dynamo.mark_static(model_input.input_tokens, 1)  # always 1
-            torch._dynamo.mark_dynamic(model_input.block_table, 1)
-            torch._dynamo.mark_static(model_input.slot_mapping, 1)  # always 1
-            torch._dynamo.mark_static(model_input.input_positions,
-                                      1)  # always 1
+        self._mark_input_tensors(model_input)
```
prashantgupta24, line 918: do we really need this function to be using `self`? It only interacts with the `model_input` param

wallashss:
> If @wallashss can confirm above change I will close this PR and he can address all in his?

I'm back to work in this refactoring, so sure I can address those.

PR #202, opened 2025-06-03
Updates for release prep

joerunde:
For multi-spyre support we need a newer build of the base image. This has been synced to `quay.io/ibm-aiu/base` and validated with tensor parallel tests.

Once this is in we can cut a new release, and the resulting image should work with tensor parallel inferencing.


rafvasq (approved):
LGTM

PR #186, opened 2025-05-28
Deprecate sendnn_decoder in favor of sendnn with warmup_mode

tjohnson31415:
If `TORCH_SENDNN_LOG` is set to WARNING instead of CRITICAL, there are logs stating
> You're using a deprecated backend. Please use sendnn in conjunction with warmup_mode

This PR makes the change to sendnn. For backwards compatibility, sendnn_decoder is overwritten to sendnn and a warning is logged.

PR #205, opened 2025-06-03
[do not merge][CB] get number of blocks from compiler mock implementation

yannicks1:
### [CB] get number of blocks from compiler mock implementation
This is a first draft how the message passing from the Spyre compiler to vLLM could work on vLLM side. 

The process consists of the following steps: 

- For warmup vLLM reserves a required minimum of 4 pages/blocks (`num_blocks=4`).
- The `num_blocks` (4) dimension is marked as dynamic (`torch._dynamo.mark_dynamic()`) for warmup forward calls only.
- The Spyre compiler calculates the maximum number of pages/blocks it can accomodate and write this to a `.json` file. 
- `torch_sendnn` reads the value from the `.json` file and can return it via a function
- vLLM calls the above mentioned `torch_sendnn` function to get the number of available blocks/pages and sets `num_blocks=N`
- For actual inference vLLM will then adjust the list of free blocks/pages and the KV cache size accordingly (`num_blocks=N`).


vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -401,11 +401,24 @@ def _warmup_spyre_dynamic_size(self, special_token_ids):
             logger.info("update_lazyhandle() done (duration: %.3fs)",
                         ul_stop_time - ul_start_time)
 
+        # get the number or pages from the actual Spyre card after the warmup
+        # and set it accordingly in the model runner and the kv cache size
+        n_blocks_spyre = self.get_num_blocks_from_compiler_mock()
+        model_runner._set_free_blocks(num_blocks=n_blocks_spyre)
+        model_runner.model.model._set_past_key_value_states(
+            num_blocks=n_blocks_spyre)
```
sducouedic, line 409: I realised all these changes only applies to `sendnn_decoder` backend and we still need the old logic for the other backends.

sducouedic: Maybe everything works by renaming function `get_num_blocks_from_compiler_mock` to `get_num_blocks` and either return `max_batch_size * max_model_len // self.BLOCK_SIZE` if cpu backend or call the torch_sendnn function if `torch_sendnn` backend

yannicks1:
> Rephrasing and update after meeting today: in the previous implementation, the number of blocks were set based on max_model_len, max_batch_size and BLOCK_SIZE and the input request were thoroughly checked against those values. Probably there are new checks to be done in order to avoid obscure errors (e.g. check max_model_len input).

Not entirely sure what you mean here. This code does not change anything, except of setting n_blocks = 4 for the warmup. This should always be enough, since we only do 2 prompts (block size is fixed to 64, Spyre constraint). After warmup it is set to what it was before.

sducouedic:
>Not entirely sure what you mean here. This code does not change anything, except of setting n_blocks = 4 for the warmup. This should always be enough, since we only do 2 prompts (block size is fixed to 64, Spyre constraint). After warmup it is set to what it was before.

I don't have a clear idea neither of the exact checks that need to be done or where things could fail. The idea is that there is a dependency between the `max_model_len`, `max_batch_size`, and the number of blocks. Depending on the values of `max_model_len` and `max_num_seqs` set by the user, we might not have enough blocks to serve a batch of sequences. This issue couldn't arise before because the number of blocks were set based on these values, but it is not the case anymore. Somehow we need to enforce that `max_model_len` and `max_num_seqs` can be served. Tom suggested to do a check and raise an error if that is not the case.

yannicks1:
I think I get you now: you mean to check something like
`num_blocks_spyre >= max_batch_size * max_model_len // block_size`, right? 
We can already incorporate this for the future, yes!

sducouedic:
> I think I get you now: you mean to check something like `num_blocks_spyre >= max_batch_size * max_model_len // block_size`, right? We can already incorporate this for the future, yes!

Yes correct

yannicks1:
@sducouedic I have incorporated your suggested check already [here](https://github.com/vllm-project/vllm-spyre/pull/205/commits/aa5754a973b883613347bf9df26c9c801d7aa6a6)

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -402,13 +402,45 @@ def _warmup_spyre_dynamic_size(self, special_token_ids):
...
+        Will eventually contain a function in torch_sendnn which reads 
+        the actual value provided by the compiler for backend sendnn_decoder"""
+
+        max_batch_size = \
+            self.model_runner.vllm_config.scheduler_config.max_num_seqs
+        max_model_len = \
+            self.model_runner.vllm_config.scheduler_config.max_model_len
+        block_size = self.model_runner.BLOCK_SIZE  # type: ignore[union-attr]
+
+        min_req_num_blocks = max_batch_size * max_model_len // block_size
```
joerunde, line 430: Upstream vllm has this check as:
```
min_req_num_blocks = max_model_len // block_size
```

I think it's more correct to only ensure you have enough blocks to run a single full-size request, because ideally you want to be able to set a high batch size to be able to run many smaller requests on a long-context model. For example say a model has a context length of 1m tokens, and with your current hardware you can only deploy it in a way where you have enough kv-cache available for 1m tokens. You wouldn't want to be forced to set the max batch size to only 1, because in practice very few requests will use the full context length. Ideally, you'd be able to set the max batch size much higher like 256, to still run many smaller requests in parallel.

That requires scheduling with preemption though to kick out request(s) from the batch when you run out of kv-cache blocks. I think we'd need to hook up our kv cache management with the scheduler and implement preemption in the model runner to make that happen. I haven't looked at how hard that would be to do yet

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -402,13 +402,45 @@ def _warmup_spyre_dynamic_size(self, special_token_ids):
...
+
+        if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == 'sendnn_decoder':
```
joerunde, line 432: `sendnn_decoder` no longer exists since https://github.com/vllm-project/vllm-spyre/pull/186

vllm_spyre/v1/worker/spyre_worker.py
yannicks1: Thanks Joe, this makes a lot of sense. changed it.

yannicks1: FYI: we need more than the above `min_req_num_blocks` for [one](https://github.com/vllm-project/vllm-spyre/pull/205/commits/42cb89974abf84dcac69a4f4c8937d7c515b6a40) of the test cases...

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -407,6 +414,52 @@ def _warmup_spyre_dynamic_size(self, special_token_ids):
...
+        # min_req_num_blocks is not enough blocks for the following test:
+        # tests/e2e/test_spyre_cb.py::test_scheduler_cb_steps_tkv
+        # [seqs_max_tokens4-prompts_lengths4-steps_add_reqs4-
+        # checked_steps4-256-False-2-eager-llama-194m]
+
+        if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == 'sendnn':
+            # TODO: replace num_blocks_spyre by calling a function in
+            # torch_sendnn which returns the value set by the Spyre compiler
+            num_blocks_spyre = max_batch_size * min_req_num_blocks
+            assert num_blocks_spyre >= min_req_num_blocks, (
```
JRosenkranz, line 438: If `min_req_num_blocks` stays same as above, we may hit a case where we can handle only a portion of the max_model_len. Should we fail in this case, or just let the user know we may not be able to support up to the full max_model_len for each request?

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -407,6 +414,52 @@ def _warmup_spyre_dynamic_size(self, special_token_ids):
...
+            max_concurrency_spyre = num_blocks_spyre * block_size \
+                / max_model_len
+            logger.info("Spyre KV cache size: %s tokens",
+                        num_blocks_spyre * block_size)
+            logger.info("Maximum concurrency for %s tokens per request: %.2fx",
+                        str(max_model_len), max_concurrency_spyre)
+            return num_blocks_spyre
+        else:  # dynamo backend 'eager'
+            # TODO: how do we get a meaningful value for CPU here
+            num_blocks_cpu = max_batch_size * min_req_num_blocks
```
JRosenkranz, line 451: At least with cuda, I believe you would profile a run and see what the peak memory usage was. Then use some percentage of whatever else was left over to determine the number of blocks that could fit.

https://github.com/vllm-project/vllm/blob/ace5cdaff0cf021ff02ddbe39ea814f2ed2e56b7/vllm/worker/worker.py#L232

There is also a method of this for cpu that calculates based on this:

```python
num_cpu_blocks = int(self.cache_config.cpu_kvcache_space_bytes //
                             cache_block_size)
```

https://github.com/vllm-project/vllm/blob/ace5cdaff0cf021ff02ddbe39ea814f2ed2e56b7/vllm/worker/cpu_worker.py#L239

vllm_spyre/v1/worker/spyre_worker.py
yannicks1: I think Joe's [comment](https://github.com/vllm-project/vllm-spyre/pull/205#discussion_r2129788301) covers this as well. Preemption will hopefully soon be implemented (or reused from upstream)...

yannicks1:
Do you guys think we can merge this PR now and insert the `torch_sendnn` function once it is available? 

Note: The PR in the current form does not change any behavior, but as it is quite some refactoring, it couldn't hurt to merge instead of waiting for the one line change inserting the function... 

@JRosenkranz @tdoublep @joerunde @sducouedic @nikolaospapandreou

sducouedic (approved):

PR #206, opened 2025-06-03
:bug: fix static scheduling issues with long prompts

joerunde:
This fixes an issue where batches of requests with long prompts would not properly be scheduled. Even with a long queue of requests, smaller-than-full batches would be scheduled because requests would be rejected from the schedule once the total number of prompt tokens was >= `--max-num-batched-tokens`.

The `--max-num-batched-tokens` config is designed for chunking up prompts for chunked prefill, and isn't relevant for static batching. This PR sets `--max-num-batched-tokens` to the maximum number of prompt tokens that could be in a full batch, so that the chunked prefill logic doesn't prevent us from creating large batches.

I missed this before because it requires a lower level test that invokes the scheduler directly in order to ensure that the full batches are actually scheduled.

### DIFF
--- a/tests/e2e/test_spyre_basic.py
+++ b/tests/e2e/test_spyre_basic.py
@@ -4,10 +4,15 @@ Run `python -m pytest tests/test_spyre_basic.py`.
 """
 
 import pytest
+from e2e.test_spyre_cb import create_random_request
 from spyre_util import (VLLM_VERSIONS, compare_results, generate_hf_output,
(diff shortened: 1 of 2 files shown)


Daniel-Schenker:
Looks good, I'll get this built and tested on Power. Thanks Joe.

tdoublep (approved):
Nice simple fix and well-written test. LGTM!

PR #209, opened 2025-06-04
:white_check_mark: add more TP sizes to tests

joerunde:
This PR adds more TP sizes to the tests marked `multi`, and skips them if the required cards are not available

### DIFF
--- a/tests/e2e/test_spyre_online.py
+++ b/tests/e2e/test_spyre_online.py
@@ -1,8 +1,7 @@
 import openai
 import pytest
-
-from tests.spyre_util import (VLLM_VERSIONS, get_spyre_backend_list,
-                              get_spyre_model_list)
+from spyre_util import (VLLM_VERSIONS, get_spyre_backend_list,
+                        get_spyre_model_list)
 
 
 @pytest.mark.parametrize("model", get_spyre_model_list())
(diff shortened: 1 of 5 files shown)


tests/e2e/test_spyre_online.py
```diff
@@ -1,8 +1,7 @@
 import openai
 import pytest
-
-from tests.spyre_util import (VLLM_VERSIONS, get_spyre_backend_list,
```
joerunde, line 4: `tests` itself isn't a package (no `__init__.py`) so this fails if you copy just the tests folder over to a machine with the plugin installed and try to run tests.

tests/spyre_util.py
```diff
@@ -537,3 +537,10 @@ def create_text_prompt(model: str, min_tokens: int, max_tokens: int) -> str:
     assert min_tokens < len(tokenizer.encode(prompt)) < max_tokens
 
     return prompt
+
+
+def skip_unsupported_tp_size(size: int):
+    cards = int(os.getenv("AIU_WORLD_SIZE", "0"))
+    if cards < size:
+        pytest.skip(f"Cannot run TP size {size}: "
+                    f"only {cards} cards are available")
```
yannicks1, line 546: I guess you never reach that part with `inductor` backend? e.g. `AIU_WORLD_SIZE` never defaults to 0 where it would skip all the sizes.

tests/spyre_util.py
```diff
@@ -537,3 +537,10 @@ def create_text_prompt(model: str, min_tokens: int, max_tokens: int) -> str:
     assert min_tokens < len(tokenizer.encode(prompt)) < max_tokens
 
     return prompt
+
+
+def skip_unsupported_tp_size(size: int):
```
rafvasq, line 542: Just a nit but something more verbose like `skip_if_insufficient_cards` might be more accurate/readable.

PR #210, opened 2025-06-04
🐛 fix for upstream change

prashantgupta24:
`data_parallel_rank` added to `EngineCoreRequest`

### DIFF
--- a/tests/e2e/test_spyre_cb.py
+++ b/tests/e2e/test_spyre_cb.py
@@ -4,6 +4,7 @@ Run `python -m pytest tests/test_spyre_cb.py`.
 """
 
 import copy
+import inspect
 from collections import deque
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


tests/e2e/test_spyre_cb.py
```diff
@@ -95,16 +96,35 @@ def create_random_request(
...
-                             sampling_params=sampling_params,
-                             eos_token_id=None,
-                             arrival_time=0,
-                             lora_request=None,
-                             cache_salt=None)
+    # Temporary until 'data_parallel_rank' parameter makes it to
+    # a release version in vllm
+    if "data_parallel_rank" in [
+            x[0] for x in inspect.getmembers(EngineCoreRequest)
+    ]:
```
joerunde, line 103: lil nit that we could do this more concisely like:
```
if "data_parallel_rank":
    extra_kwargs = {"data_arallel_rank": None}
else:
    extra_kwargs = {}

return EngineCoreRequest(
    ...
    **extra_kwargs)
```

but my care amount here is low

PR #214, opened 2025-06-05
:bug: fetch tags for versioning in docker build

joerunde:
The docker images don't have a correct version for the package, they all report:
```
$ pip3 list | grep spyre
vllm-spyre                               0.1.dev1
```
This is because we don't have tag info to version from if we don't fetch any history on checkout in the docker build workflow. Fetch-depth 0 should fetch all tags: https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -33,6 +33,11 @@ jobs:
     steps:
     - name: "Checkout"
       uses: actions/checkout@v4
+      # We need to fetch history and tags so that we can correctly version the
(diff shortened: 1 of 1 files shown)


joerunde:
From the build workflow on this PR:
```
#12 35.56  + vllm-spyre==0.3.1.dev17 (from file:///usr/src/vllm-spyre)
```

ckadner (approved):
LGTM

PR #217, opened 2025-06-06
Add support for vllm model implementation

SilverSoldier:
Changing vllm_spyre to use vllm's own model implementation instead of fms. 

Status: Tested on llama and fails at self.attn() in LlamaAttention.

NOTE: Attention, KV Cache and Paging implementation is totally incomplete.


SilverSoldier:
@tdoublep FYI, these are the changes we made in vllm-spyre side apart from the vllm changes mentioned:
- import fusion only if compilation config is set
- replace unsupported ops with their supported equivalents

joerunde:
Tests fail as expected 😉

This looks pretty cool!

PR #219, opened 2025-06-06
fixed issue with warmup_context not capturing full generate

JRosenkranz:
Warmup context needs to capture both prefill/decode. This PR fixes that issue.


yannicks1:
thanks Josh

yannicks1 (approved):
lgtm

tjohnson31415:
@JRosenkranz Thanks for catching this!

I do have a question with the fix: Is it ok that there are 2 full generation passes through the model in the warmup context for the static batching case now?

PR #224, opened 2025-06-09
✨ add top-level spyre version

prashantgupta24:
This PR allows us to do something like this:
```
>>> vllm_spyre.__version__
'0.3.2.dev17'
```

### DIFF
--- a/vllm_spyre/__init__.py
+++ b/vllm_spyre/__init__.py
@@ -1,6 +1,9 @@
+import importlib.metadata
 from logging.config import dictConfig
 
 from vllm.logger import DEFAULT_LOGGING_CONFIG
+
+__version__ = importlib.metadata.version("vllm_spyre")
 
 
 def register():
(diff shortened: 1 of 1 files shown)


vllm_spyre/__init__.py
prashantgupta24, line 1: I copied this from the original vllm version file [here](https://github.com/vllm-project/vllm/blob/main/vllm/version.py). Let me know if we need to add something else.

prashantgupta24: Interesting. Although this has a very unique side effect in which the top level library gets a `version` attribute associated with it. I'm not sure if that's intentional

```
>>> import fms
>>> fms.version
<function version at 0x102894cc0>
>>> fms.version("fms")
'0.0.8'
```

joerunde: oh, yeah that's not ideal though but it could probably be worked around. We could override `__all__`, or do `from importlib.metadata import version as __version`, or 
```
import importlib.metadata
__version__ = importlib.metadata.version("vllm_spyre")
```

joerunde:
Looking at the failing main tests here, it looks like they're deadlocking and timing out?

Very suspicious

vllm_spyre/__init__.py
prashantgupta24: Good point. I should probably open a PR for fms too then lol

prashantgupta24: https://github.com/foundation-model-stack/foundation-model-stack/pull/424

PR #228, opened 2025-06-11
[CB] parametrize example script

sducouedic:
As per request of the compiler team


examples/offline_inference/offline_inference_spyre_cb_test.py
prashantgupta24, line 1: I like the changes!

examples/offline_inference/offline_inference_spyre_cb_test.py
```diff
@@ -7,82 +8,102 @@
...
+    for output in outputs:
+        prompt = output.prompt
+        generated_text = output.outputs[0].text
+        print(f"\nPrompt:\n {prompt!r}")
+        print(f"\nGenerated text:\n {generated_text!r}\n")
+        print("-----------------------------------")
+
+
+if __name__ == "__main__":
+    main()
```
prashantgupta24, line 109: Kind of overkill because I don't see this file being used at import time anywhere, but doesn't hurt to have I guess

prashantgupta24 (approved):
LGTM

examples/offline_inference/offline_inference_spyre_cb_test.py
sducouedic: You're right, I just removed that

PR #229, opened 2025-06-11
:fire::wrench: Remove environment variables specific to hardware conf

gkumbhat:
### Changes

Following environment variables are quite specific to the hardware and platform vllm-spyre is being run on. Therefore, ideally we should rely on the "reasonable" defaults for this from underlying stack rather than changing those defaults in vLLM.

Env variables:
- `FLEX_UNLINK_DEVMEM`
- `FLEX_RDMA_MODE_FULL`


joerunde:
LGTM, 🤞 that soon we'll have public docs to link out to about these so that users actually know how to set them when they're needed

PR #230, opened 2025-06-11
:sparkles: Add support for compiling dynamic shapes

gkumbhat:
### Changes

Since compiler now supports symbolic shapes, we can enable passing these to compiler. This will help in less shapes munging and tracking upfront and allow more utilization of compilation cache.


joerunde:
I think the goal here is to do this with continuous batching, and that's close to a working implementation on spyre cards but everything's not quite all released yet

gkumbhat:
@joerunde I think this one relies on the dynamic tkv symbolic shapes, which was released in compiler sometime back. I am working on testing this out, will update the PR as per my findings.

joerunde:
Gotcha, I just meant that our team isn't spending any of our time on bringing improvements to the static batching path

PR #232, opened 2025-06-12
Exclude vllm v0.9.1 as an allowed version due to breaking bug

tjohnson31415:
vLLM v0.9.1 contains a bug that causes vllm-spyre to hang on boot-up.

The bug is not respecting `num_gpu_blocks_overrides`. It was introduced in https://github.com/vllm-project/vllm/pull/17996 and fixed in https://github.com/vllm-project/vllm/pull/19503.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ license = {text = "Apache 2"}
 dependencies = [
     "fms-model-optimizer>=0.2.0",
     "ibm-fms==1.0.0",
-    "vllm>=0.9.0",
+    "vllm>0.9.0,!=0.9.1",
 ]
 requires-python = ">=3.9"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -13,7 +13,7 @@ license = {text = "Apache 2"}
 dependencies = [
     "fms-model-optimizer>=0.2.0",
     "ibm-fms==1.0.0",
-    "vllm>=0.9.0",
+    "vllm>0.9.0,!=0.9.1",
```
joerunde, line 16: Do we need to say `>0.9.0`? I thought we still worked with `0.9.0`

tjohnson31415: Ah, nice catch!
Fixed

Issue #225, opened 2025-06-09
[CB] Server terminates after cancelling requests


rafvasq:
Similar to #36, it doesn't happen every time but when cancelling requests (e.g. a benchmark run) the server terminates: 

```
...
INFO 06-09 19:38:56 [async_llm.py:327] Request cmpl-5156d6b05e3c4013addf6b25202cbec8-0 aborted.
INFO 06-09 19:38:56 [async_llm.py:420] Aborted request cmpl-07e571960ee14cf7b8bd25f3a8d31770-0.
INFO 06-09 19:38:56 [async_llm.py:327] Request cmpl-07e571960ee14cf7b8bd25f3a8d31770-0 aborted.
INFO 06-09 19:38:56 [async_llm.py:420] Aborted request cmpl-4d48eb0753dc43ce9d66c70559b4981a-0.
INFO 06-09 19:38:56 [async_llm.py:327] Request cmpl-4d48eb0753dc43ce9d66c70559b4981a-0 aborted.
ERROR 06-09 19:38:57 [dump_input.py:68] Dumping input data
[...]

PR #233, opened 2025-06-12
fix: avoid KeyError when cancelling requests that have not been processed

tjohnson31415:
FIX https://github.com/vllm-project/vllm-spyre/issues/225

### DIFF
--- a/tests/e2e/test_spyre_async_llm.py
+++ b/tests/e2e/test_spyre_async_llm.py
@@ -46,7 +46,7 @@ async def generate(
         if cancel_after is not None and count >= cancel_after:
             return count, request_id
 
-        await asyncio.sleep(0.0)
+        await asyncio.sleep(0.01)
 
     return count, request_id
 
--- a/tests/e2e/test_spyre_async_llm.py
+++ b/tests/e2e/test_spyre_async_llm.py
@@ -1,6 +1,5 @@
 import asyncio
 from contextlib import ExitStack
-from typing import Optional
 
 import pytest
 from spyre_util import get_spyre_backend_list, get_spyre_model_list
@@ -18,8 +17,6 @@ async def generate(
     output_kind: RequestOutputKind,
     max_tokens: int,
     n: int = 1,
-    prompt_logprobs: Optional[int] = None,
-    cancel_after: Optional[int] = None,
 ) -> tuple[int, str]:
     # Ensure generate doesn't complete too fast for cancellation test.
     await asyncio.sleep(0.2)
@@ -31,7 +28,6 @@ async def generate(
         output_kind=output_kind,
         seed=42,
         n=n,
-        prompt_logprobs=prompt_logprobs,
     )
     async for out in engine.generate(request_id=request_id,
                                      prompt=prompt,
@@ -42,9 +38,6 @@ async def generate(
             count += num_tokens
         else:
             count = num_tokens
-
-        if cancel_after is not None and count >= cancel_after:
-            return count, request_id
 
         await asyncio.sleep(0.01)
 
--- a/tests/e2e/test_spyre_async_llm.py
+++ b/tests/e2e/test_spyre_async_llm.py
@@ -1,0 +1,170 @@
+import asyncio
+import os
+from contextlib import ExitStack
+from typing import Optional
+
+import pytest
+from spyre_util import get_spyre_backend_list, get_spyre_model_list
+from vllm import PromptType, SamplingParams
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.engine.async_llm_engine import AsyncLLMEngine
+from vllm.sampling_params import RequestOutputKind
+from vllm.v1.engine.async_llm import AsyncLLM
+
+VLLM_VERSIONS = [
+    pytest.param("V0", marks=pytest.mark.v0, id="v0"),
+    pytest.param("V1", marks=pytest.mark.v1, id="v1"),
+]
+
+
+async def generate(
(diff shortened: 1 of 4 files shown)


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -603,15 +603,15 @@ def _update_states(self, scheduler_output):
 
         # Continuous batching stuff
         for req_id in scheduler_output.finished_req_ids:
-            if req_id in self.req_ids2blocks:
+            # requests may be cancelled from the client side while in the queue
+            if req_id in self.requests:
```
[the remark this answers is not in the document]
tjohnson31415: Yeah, thats a good question.
If we look at this code in isolation, each `del` on a map should be checked for presence of they key before the removal (eg. `req_ids2left_pads` should be checked too). But the request is added to all maps during the course of `_update_states`:
https://github.com/vllm-project/vllm-spyre/blob/97d03d6003c7afee846c76790a145287d4774d52/vllm_spyre/v1/worker/spyre_model_runner.py#L650-L688

So the idea is that if it is in `self.requests` it is also in the other maps. But probably better to be safe and not make that assumption.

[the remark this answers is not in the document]
prashantgupta24: Yeah, I'm also trying to understand the problem here. Could it be that there's a race condition between 2 threads and one thread actually deletes the request from `self.requests` while the other is trying to do the same but fails? In that case then Christian's comment above does make sense. Or maybe we can also think about using thread-safe mechanisms here?

ckadner (approved):
Elegant!

vllm_spyre/v1/worker/spyre_model_runner.py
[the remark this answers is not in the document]
joerunde: There shouldn't be any threading in vllm though, as far as I know. The only concurrency should be with async or multiprocessing 😕

tjohnson31415:
I think we will have to wait for https://github.com/vllm-project/vllm-spyre/pull/162 to be merged in to get the tests to pass (they pass in my dev env if I include that PR).

ckadner:
> I think we will have to wait for #162 to be merged in to get the tests to pass (they pass in my dev env if I include that PR).

Is it the `and not cb` marker in the `.github/workflows/test.yml` that will make your new tests work (ignored/skipped) for `V1-e2e`?

tests/e2e/test_spyre_async_llm.py
```diff
@@ -0,0 +1,174 @@
...
+        num_tokens = sum(len(output.token_ids) for output in out.outputs)
+        if output_kind == RequestOutputKind.DELTA:
+            count += num_tokens
+        else:
+            count = num_tokens
+
+        if cancel_after is not None and count >= cancel_after:
+            return count, request_id
+
+        await asyncio.sleep(0.0)
```
yannicks1, line 55: is `await asyncio.sleep(0.0`) enough or do we need `await asyncio.sleep(x>0)`. background: arguments 0 and x>0 might not behave the same (see [here](https://stackoverflow.com/questions/74493571/asyncio-sleep0-does-not-yield-control-to-the-event-loop))

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -602,14 +602,11 @@ def _update_states(self, scheduler_output):
...
         for req_id in scheduler_output.finished_req_ids:
-            if req_id in self.req_ids2blocks:
-                logger.debug("Freeing request id: %s", req_id)
-                for freed_block in self.req_ids2blocks[req_id]:
-                    self.free_blocks.append(freed_block)
-                del self.req_ids2blocks[req_id]
-                del self.req_ids2left_pads[req_id]
-
-            del self.requests[req_id]
+            logger.debug("Finishing request id: %s", req_id)
```
yannicks1, line 605:
```suggestion
            if req_id in self.req_ids2blocks:
                logger.debug("Freeing request id: %s", req_id)
```

This debug statement was for specific for CB to have confirmation that the blocks were actually freed. I would suggest to keep it that way. IMO it should not be a general statement that a request has finished, this could probably (or is already) at some other place in the code, not in the CB specific part.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -602,14 +602,11 @@ def _update_states(self, scheduler_output):
...
+            for block_id in self.req_ids2blocks.pop(req_id, []):
```
yannicks1, line 606: nice hack with the popping ! :)

yannicks1 (approved):
lgtm, once test pass this is ready to be merged

tests/e2e/test_spyre_async_llm.py
tjohnson31415: That is interesting that the behavior is different! Thanks for the link.

The tests seem fine with `0.0`, but I'll add a small value just in case.

tests/e2e/test_spyre_async_llm.py
```diff
@@ -0,0 +1,149 @@
...
+    backend: str,
+    cb: int,
+    warmup_shapes: list[list[int]],
+    output_kind: RequestOutputKind,
+    monkeypatch: pytest.MonkeyPatch,
+):
+    """Test handling of cancelled requests"""
+
+    if cb == 1 and backend != "eager":
+        pytest.skip("CB requires eager")
```
joerunde, line 68: we can come back and fix this up to work on spyre

Issue #150, opened 2025-05-09
Tests `test_spyre_seed.py` timeouts undeterministically


sducouedic:
With `sendnn_decoder` backend, it happens that `test_spyre_seed` tests block until it timeouts. It was previously attributed to compile caching problem, but it was also observed with caching disabled.

joerunde:
@sducouedic is this with the v1 backend and are you using `--forked`?

Wondering if this is related to #146

sducouedic:
So yes v1 and yes `--forked`. It might be due also to compilation caching, I am not 100% sure it happens not only with caching. I have quite some trouble to debug that.

PR #238, opened 2025-06-16
[FIX] lazy import of SpyreCausalLM to avoid issues with pytest-forked

wallashss:
# Description

We were having problems with pytest-forked and cache. After some investigation and debugging I figure it out that we were importing torch_sendnn before it was necessary, so pytest already initialize the runtime and the forking inherit it, which I believe it was the cause of the problems with the cache.

Also I re-enabled a test that was skipped because of this known issue.

## Related Issues

Potentially fix: #150

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -13,7 +13,7 @@ from vllm.utils import is_pin_memory_available
 from vllm.v1.kv_cache_interface import FullAttentionSpec, KVCacheSpec
 from vllm.v1.outputs import SamplerOutput
 
-# from vllm_spyre.model_executor.model_loader.spyre import SpyreCausalLM
+from vllm_spyre.model_executor.model_loader.spyre import SpyreCausalLM
 from vllm_spyre.platform import SpyrePlatform
 from vllm_spyre.v1.worker.spyre_input_batch import (CachedRequestState,
                                                     InputBatch)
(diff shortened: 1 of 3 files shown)


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -109,6 +109,11 @@ def get_model(self) -> nn.Module:
 
     def load_model(self, prompt_lens: Iterable[int],
                    num_decode_tokens: Iterable[int]) -> None:
+        # Lazy import to avoid load torch_sendnn runtime before it is really
+        # necessary. This solve issues of running forked tests that share
+        # some resources from parent to children which can have problems
+        # of caching even though the test run in isolated subprocesses.
+        from vllm_spyre.model_executor.model_loader.spyre import SpyreCausalLM
```
tjohnson31415, line 116: Ideally, we wouldn't require a lazy import of a module in `vllm_spyre`. Could we move the import of `troch_sendnn` further down into the code?

I don't think the import is needed until we run `torch.compile` at:
https://github.com/vllm-project/vllm-spyre/blob/58cb2490cb74f8487f67a1b5f400bbe21993249b/vllm_spyre/model_executor/model_loader/spyre.py#L265

wallashss: makes sense! Also, I just remembered that we two model runner (v0 too). So, I think it will be better, so we can fix both imports.

tdoublep:
Great find @wallashss !

PR #239, opened 2025-06-17
move function create_random_request to spyre_utils.py

yannicks1:
# Description


tests/e2e/test_spyre_tensor_parallel.py
```diff
@@ -1,6 +1,6 @@
 """Verification of vLLM output by comparing with HF
 
-Run `python -m pytest tests/test_spyre_tensor_parallel.py`.
```
sducouedic, line 3: good catch

sducouedic (approved):
LGTM (I wanted to do the same changes 😂)

Issue #168, opened 2025-05-16
[CB] Test deterministic output by comparing to HF API


prashantgupta24:
There are some tests that assert output using certain input values. Ideally, we should be comparing the outputs against HF in all scenarios.

prashantgupta24:
I'm still debating whether we want to test more with the steps function which is already doing so much at the moment. I was thinking we could add a different test that verifies output with HF? Still not 100% sure which way is better.

sducouedic:
yeah I was also hesitant, I agree that it would be better to have a separate test function that checks the HF output. I am just wondering if the output of some of the different situation cases should be checked (e.g. reduce-left-padding) in addition than just checking the scheduler and tkv, and in the latter case then we might as well reuse the different cases implemented for the other test function

prashantgupta24:
Yep, I agree.

PR #240, opened 2025-06-18
Use hugging face as baseline to test CB output

sducouedic:
Closes issue #168 

@joerunde About the llama issue, maybe we can wait until and e2e image gets published with CB enabled before removing it from the list of model for cb? it won't be used in the meantime in the CI/CD anyway, and this allows to test it on cpu for now, what do you think?


tests/e2e/test_spyre_basic.py
prashantgupta24, line 1: uhh there's gonna be a ton of conflicts from main because we _just_ removed the v0 and v1 markers from all tests

joerunde:
> About the llama issue, maybe we can wait until and e2e image gets published with CB enabled before removing it from the list of model for cb? it won't be used in the meantime in the CI/CD anyway, and this allows to test it on cpu for now, what do you think?

I think we can leave it in the list, it sounds like we should be fine to test llama models on cpu going forward. We can update the test configs over on the spyre hardware to swap to granite models

prashantgupta24 (changes requested):
I'm going to request changes just so that we can pause more changes for CB. Right now there's a ton of uncertainty surrounding CB and we don't want to make any changes until we have more clarity

prashantgupta24:
The changes that you have here are amazing But there's still some uncertainty around the output that continuous batching is generating on the actual `spyre` hardware. Ideally, we want to get to the point where this PR works as is, But clearly both `llama` and `granite` models are not generating output which is gonna match what `HF` is producing at the moment. 

Probably worth spending a bit of time understanding why and see if we want to push a "working" implementation which doesn't produce good results in the meantime. In that case, we will not be able to merge this PR as is because the outputs will never match `HF`. That's the only reason I recommended a pause for a little while until the investigation gets us more insight?

sducouedic:
@prashantgupta24 ah I think there is some confusion, the tests were not passing because of some breaking changes in vllm upstream (a new parameter introduced again), fixed in this PR #245. I just synced with main and all the tests are passing now (at least on the CPU)

sducouedic:
Ah ok I see. Fundamentally it doesn't really change to compare with HF output or a hardcoded ground truth. But yes let's freeze the code if that can simplify debugging for spyre 👍

prashantgupta24 (approved):
LGTM! We have a plan on how to handle testing with spyre

Issue #204, opened 2025-06-03
[CB] Test minimal batch size of 2 in decode mode


sducouedic:
In decode mode, the FMS API requires batch of minimum size of 2. When there is only one sequence prompt scheduled, a second dummy request is artificially passed to the model for running. This is done through this PR: https://github.com/vllm-project/vllm-spyre/pull/182

We need to test verify the following points, once the above PR gets merged:
1. If only one sequence is requested, a dummy request is created
2. The dummy request disappear and the respective blocks get free up as soon as a new request joins
3. If `VLLM_SPYRE_RM_PADDED_BLOCKS` is set, the dummy sequence also gets the left padding correctly removed

A possible solution to test this would be through stepping using an vLLM engine core object, and checking the variables of the `ContinuousBatchingSpyreModelRunner`:
`runner_obj = engine_core.model_executor.driver_worker.worker.model_runner` ([ref](https://github.com/vllm-project/vllm-spyre/pull/156#issuecomment-2881219296))

yannicks1:
this should be solved with this PR #259

PR #243, opened 2025-06-18
Add tests for dummy requests creation and removal

gmarinho2:
# Description
Add tests to check if dummy requests are created when there is only one sequence prompt scheduled and removed when a second true request is received.

## Related Issues
https://github.com/vllm-project/vllm-spyre/issues/204

### DIFF
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -1,4 +1,3 @@
-import inspect
 import math
 import os
 import subprocess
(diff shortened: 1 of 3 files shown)


yannicks1:
Hi @gmarinho2, this PR does not look ready yet, do you mind to convert it to a draft while still working on it. Otherwise code owners get notifications to review...

PR #244, opened 2025-06-18
🐛 fix runtime msg

prashantgupta24:
# Description

Earlier, the error showed up as
```
RuntimeError: Warmup shape [4, {shape['prompt_length']}, {shape['new_tokens']}] results in a maximum sequence length of {max_seq_len} which is longer that what the model supports ({max_model_len})
```
After this fix, it shows up correctly:
```
RuntimeError: Warmup shape [4, 128, 20] results in a maximum sequence length of 148 which is longer that what the model supports (128)
```


yannicks1 (approved):
lgtm, nice catch!

PR #248, opened 2025-06-19
✅  add tokenizer option for testing

prashantgupta24:
# Description

Add `VLLM_SPYRE_TEST_TOKENIZER_LIST` as an optional parameter while testing. This is used by `tiny-granite` models which need to supply `tokenizer` as an additional input to vllm.

```
export VLLM_SPYRE_TEST_MODEL_LIST=tiny-granite
```
```
export VLLM_SPYRE_TEST_TOKENIZER_LIST=ibm-granite/granite-3.2-8b-Instruct 
```
then
```
pytest -v -m "spyre and cb" tests
```

will start with
```
test_cb_handling[max_num_seqs(4)-num_prompts(4)-eager-model(models/tiny-granite), tokenizer(ibm-granite/granite-3.2-8b-Instruct)]
.
.
.
```

Also changes `max_model_len` to `256` in CB tests.

### DIFF
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -490,47 +490,6 @@ def get_spyre_backend_list():
     return backends
 
 
-# # get model names from env, if not set then default to "llama-194m"
-# # For multiple values:
-# # export SPYRE_TEST_MODEL_LIST="llama-194m,all-roberta-large-v1"
-# def get_spyre_model_list(isEmbeddings=False, quantization=None):
-#     spyre_model_dir_path = get_spyre_model_dir_path()
-
-#     if isEmbeddings:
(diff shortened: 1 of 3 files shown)


tests/spyre_util.py
```diff
@@ -226,7 +226,7 @@ def generate_cb_spyre_vllm_output(
 
         vllm_model = LLM(
             model=model,
-            tokenizer=model,
+            tokenizer="ibm-granite/granite-3.2-8b-instruct",
```
joerunde, line 229: I think this will have to be set conditionally

prashantgupta24: Yeah just wanted to test for now

PR #253, opened 2025-06-23
[do not merge][CB] requesting only one token via index (fms api change)

yannicks1:
### [do not merge][CB] requesting only one token via index (fms api change)

To minimize data transfer for CB, we only want to request the last prompt tokens logits instead of the the entire prompts logits.
Therefore the flag `only_last_token: bool` will be replaced by the argument `index: int` in the fms forward api. When passing and index i, fms will return the logits for the token at i-th position only. A draft implementation of this in fms can be found [here](https://github.com/foundation-model-stack/foundation-model-stack/tree/only_last_token_rework). 

_Note: we only request the last token logits for static batching by default, as no right padding is ever required there._ 

**changes:** 
- install fms feature [branch](https://github.com/foundation-model-stack/foundation-model-stack/tree/only_last_token_rework) for testing
- passing `index` instead of `only_last_token`
- set number of right pads for decode to 0

solves #254


yannicks1:
#464 has been merged as a intermediate fix/stepping stone. Performance wise not critical anymore to return precisely one token.

joerunde:
closing as done in #464 then!

Issue #256, opened 2025-06-23
Add test for requests finishing at different times


prashantgupta24:
`max_tokens` varies between the requests

PR #257, opened 2025-06-23
✅ CB tests refactoring + adding batch test

prashantgupta24:
# Description

- ✅ change `max_model_len` to `256`
- ✅ edit HF test description
- ✅add `pytest.xfail` for failing CB tests
- ✅ run batch handling test for CB too

# Issues
- Fix https://github.com/vllm-project/vllm-spyre/issues/256

### DIFF
--- a/tests/e2e/test_spyre_cb.py
+++ b/tests/e2e/test_spyre_cb.py
@@ -62,8 +62,8 @@ def test_cb_handling(
     vllm_results = generate_spyre_vllm_output(
         model=model,
         prompts=prompts,
-        max_model_len=2048,
-        block_size=2048,
+        max_model_len=256,
+        block_size=256,
         sampling_params=vllm_sampling_params,
         tensor_parallel_size=1,
         backend=backend,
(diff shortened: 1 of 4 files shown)


tests/e2e/test_spyre_cb.py
```diff
@@ -23,47 +23,118 @@
...
-]])
-def test_cb_handling(
+], )
+
+
+@pytest.mark.cb
+@pytest.mark.parametrize("model", get_spyre_model_list())
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+@pytest.mark.parametrize("prompts", prompts)
+def test_cb_batch_handling(
```
prashantgupta24, line 42: This test is failing on both CPU and spyre hardware, will need to debug

tests/e2e/test_spyre_cb.py
```diff
@@ -643,7 +628,6 @@ def augment_checked_steps(
 @pytest.mark.cb
 @pytest.mark.parametrize("model", get_spyre_model_list())
 @pytest.mark.parametrize("backend", get_spyre_backend_list())
-@pytest.mark.parametrize("max_num_seqs", [2])
 @pytest.mark.parametrize(
     "seqs_max_tokens,prompts_lengths,steps_add_reqs,checked_steps,"
     "max_model_len", [
```
yannicks1, line 633: (I know this is not your change, but) could we move `max_model_len` out of the `get_params*` functions as part of this refactoring PR? It is not test case specific and should not be needed to set at 5 different places (currently it is set to 256 in all functions)

joerunde:
mergin'!

PR #261, opened 2025-06-24
[CB] Scheduling constraints regarding number of available blocks/pages

### LINKED ISSUE #260: [CB] Scheduling constraints regarding number of available blocks/pages
### [CB] Scheduling constraints regarding number of available blocks/pages

`ContinuousBatchingSpyreScheduler` needs to respect the current number of free blocks/pages (`model_runner.free_blocks`) when deciding whether a request can be scheduled in` can_schedule()`. 

related PR

yannicks1:
### [CB] Scheduling constraints regarding number of available blocks/pages

changes:
- [x] move hard coded `BLOCK_SIZE` (64) variable to Platform class and import it where needed instead of defining it in multiple different places. 
- [x] introduce scheduler constraints regarding number of available blocks/pages in `can_schedule()`
- [ ] write unit test for new scheduler constraint

closes #260

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -48,8 +48,9 @@ class ModelForwardInputs:
 
 @dataclass
 class CBSpyreModelRunnerOutput(ModelRunnerOutput):
-    # Add the current tkv to the output
+    # Add the current tkv and the number of free blocks to the output
     tkv: int = 0
+    n_free_blocks: int = 0
 
 
 class SpyreModelRunner:
@@ -965,6 +966,8 @@ class ContinuousBatchingSpyreModelRunner(SpyreModelRunner):
                                             logprobs=None,
                                             prompt_logprobs_dict={},
                                             tkv=0,
+                                            n_free_blocks=len(
+                                                self.free_blocks),
                                             **extra_kwargs)
 
         model_input = self.prepare_model_input(scheduler_output)
@@ -1051,6 +1054,7 @@ class ContinuousBatchingSpyreModelRunner(SpyreModelRunner):
                                   for req_id in req_ids
                                   },  # TODO(wallas?): prompt logprobs too
             tkv=self.tkv,
+            n_free_blocks=len(self.free_blocks),
             **extra_kwargs,
         )
         return model_output
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -579,7 +579,7 @@ class ContinuousBatchingSpyreModelRunner(SpyreModelRunner):
         assert vllm_config.scheduler_config.max_num_seqs >= 2, "Currently, " \
             "continuous batching needs config to set batch_size >= 2"
 
-        self.BLOCK_SIZE = 64  # hardcoded Spyre constraint for now
+        self.BLOCK_SIZE = SpyrePlatform.get_block_size()
 
         # TO DO: move to InputBatch
         self.req_ids2blocks: dict[str, deque[int]] = {}
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -17,6 +17,7 @@ from vllm.model_executor.model_loader.weight_utils import (
 from vllm.model_executor.sampling_metadata import SamplingMetadata
 
 import vllm_spyre.envs as envs_spyre
+from vllm_spyre.platform import SpyrePlatform
 
 try:
     import backends.dynamo_tracer  # noqa: F401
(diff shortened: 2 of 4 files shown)


tests/e2e/test_spyre_cb.py
```diff
@@ -592,7 +606,222 @@ def get_params_test_scheduler_constraints_max_prompt_len():
...
+        {
+            "step": 0,
+            "tkv": 0,
+            "waiting": ["0", "1"],
+            "running": [],
+            "request_outputs": []
+        },
+        {
+            # Prefill sequence 0
+            # total blocks in use: 1
```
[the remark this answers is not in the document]
yannicks1: no, I think this is a nice additional check! Will add this

tests/e2e/test_spyre_cb.py
```diff
@@ -114,35 +114,52 @@ def get_params_test_blocks_borders_aligned_prompts():
...
 
     checked_steps = [
         {
             "step": 0,
             "tkv": 0,
             "waiting": ["0", "1", "2"],
             "running": [],
-            "request_outputs": []
+            "request_outputs": [],
+            "n_reserved_blocks": 0,
```
sducouedic, line 127: For the future, but I am wondering if it is necessary to always add newly tested values to all the previous existing tests, this can quickly become cumbersome to maintain. Maybe in the main running loop we could tests the keys values only if they are found in the reference dictionary, this way we can only add target tested attributes in the corresponding test

sducouedic:
LGTM

sducouedic (approved):

tests/e2e/test_spyre_cb.py
sducouedic: ah absolutely, as I said it is a comment for the future. I didn't mean to remove the existing `request_outputs` or add more logic 👍

vllm_spyre/envs.py
```diff
@@ -74,6 +75,11 @@ def _backend_backwards_compat() -> str:
     "VLLM_SPYRE_USE_CB":
     lambda: bool(int(os.getenv("VLLM_SPYRE_USE_CB", "0"))),
 
+    # If set, use the V1 continuous batching implementation. Otherwise, static
+    # batching mode will be enabled.
```
joerunde, line 79: I don't think this comment is correct? From the code it looks like this is an override for the number of kv cache blocks which is used in place of the simple `max_model_len * max_num_seqs` calculation.

I'm assuming this is for passing in the known good values for the available blocks given a tested model and card combo. Can we also consider using the existing kv cache override instead of setting up a new one? There's `--num-gpu-blocks-override` which will set the available blocks in the scheduler config

tests/e2e/test_spyre_cb.py
joerunde: For the future we could also consider converting these to snapshot tests. The idea there is that for any new test cases you only need to specify the inputs, and then the test can snapshot the outputs automatically for your review. The snapshots are stored as data files read by the test, and when schema changes like this happen it's simple to just regenerate all the snapshots

vllm_spyre/envs.py
yannicks1: yes, the comment is from copy pasting the code from `VLLM_SPYRE_USE_CB` 😆 
good pointer with `--num-gpu-blocks-override`, will look into that!

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -422,7 +422,10 @@ def _warmup_spyre_dynamic_size(self, special_token_ids):
         # get the number or pages from the actual Spyre card after the warmup
         # and set it accordingly in the model runner and the kv cache size
         n_blocks_avail = self._get_num_blocks_available()
-        model_runner._set_free_blocks(num_blocks=n_blocks_avail)
+        # overwrite n_blocks_avail for testing scheduler constraints
+        if envs_spyre.VLLM_SPYRE_N_BLOCKS > 0:
+            n_blocks_avail = envs_spyre.VLLM_SPYRE_N_BLOCKS
+        model_runner._set_blocks(num_blocks=n_blocks_avail)
```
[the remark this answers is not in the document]
yannicks1: although this was not part of this PR, you are absolutely right and I moved the function 😇

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -715,8 +723,52 @@ def __init__(
...
-        self.free_blocks = deque([i for i in range(num_blocks)])
+    def _set_blocks(self, num_blocks: int) -> None:
+        self.n_blocks = num_blocks
+        self.block_pool = deque([i for i in range(self.n_blocks)])
+
+    def _get_num_blocks_available(self) -> int:
+        """Function returns the number of available blocks/pages.
+        Will eventually contain a function in torch_sendnn which reads 
+        the actual value provided by the compiler for backend sendnn"""
+
```
joerunde, line 734: can we consolidate some of the `if envs_spyre.VLLM_SPYRE_N_BLOCKS > 0:` checks into here and return the override from this method?

Issue #255, opened 2025-06-23
[CB] Reduce wastage in prefill compute and pad blocks in homogeneous continuous batching


yannicks1:
### Reduce wastage in prefill compute and pad blocks in homogeneous continuous batching

implement optimization idea by @JRosenkranz: do prefill only on next multiple of block size and then during decode pad with (valid) block id. Reduces computes for prefill and does not waist any valid blocks ids if whole blocks are padded to make tkv homogeneous.

PR #262, opened 2025-06-24
[CB] Optimization: Reduce wastage in prefill compute and pad blocks in homogeneous continuous batching

yannicks1:
### [CB] Reduce wastage in prefill compute and pad blocks in homogeneous continuous batching

implement optimization idea by @JRosenkranz: do prefill only on next multiple of block size and then during decode pad with (valid) block id. Reduces computes for prefill and does not waist any valid blocks ids if whole blocks are padded to make tkv homogeneous.

solves #255


yannicks1:
great news: This runs on Spyre 🎉 

I just ran `cb_spyre_inference.py` which (with the parameters on this branch) exploits all functionality: 
- prefill seq 1 of size 128 (left pads)
- prefill seq 2 of size 64 (left pads) (see #255)
- decode of batch 2
- strips fully padded blocks (see #131 )
- prefill seq 3 by left padding to 66 (align tkv) and right padding to 128 (pad to block boundary)
- decode of batch 2
- decode of batch 1 (see #182)

cc: @tdoublep @JRosenkranz @joerunde @nikolaospapandreou @sducouedic

yannicks1:
bot:test
TEST_FILE=tests/e2e/test_spyre_cb.py MARKERS="spyre"

yannicks1:
bot:test
TEST_FILE=tests/e2e/test_spyre_cb.py MARKERS="spyre"

yannicks1:
6/7 tests passed on the Spyre card! looks like the failure is a known issue unrelated to this PR. 🥳

yannicks1:
bot:test
MARKERS="spyre"

yannicks1:
bot:test
TEST_FILE=tests/e2e/test_spyre_cb_scheduler_step.py MARKERS="spyre"

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -931,16 +941,24 @@ def _prepare_decode(
...
-            # adding new blocks if needed
-            if self.tkv // self.block_size + 1 > len(
-                    self.req_ids2blocks[req_id]):
-                self.req_ids2blocks[req_id].append(self.block_pool.popleft())
-            block_table.append(self.req_ids2blocks[req_id])
+
+            # filling block table with padding blocks (reusing id 0)
+            blocks = self.req_ids2blocks[req_id].copy()
+            for i in range(n_blocks - len(self.req_ids2blocks[req_id])):
+                blocks.appendleft(0)
```
maxdebayser, line 959: Perhaps this is an incredibly stupid question, but why is it ok to use block id 0? Does it make a difference if it free (i.e. it's in self.block_pool) or not?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -884,8 +892,10 @@ def _prepare_prompt(
         # applies left padding to align with tkv of current decode batch
         # and right padding to align with the next block boundary
         input_tokens, position_ids, mask =\
```
prashantgupta24, line 894: Still processing the changes, but wondering if the comment above needs rewording...

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24: Based on Josh's comment in an internal issue,

"The only requirement when padding this is that we choose a block ID that exists in the pool of allotted block ids at server start (this way it will map to a real location in the memory space). In this case, when performing paged attention compute, the placeholder block will be part of compute (we will still take a performance hit during decode - at least until heterogeneous tkv is available), but will not take up any extra space as part of the block allotment at the time of server start."

IIUC I think that's why block 0 makes sense?

prashantgupta24: Although yeah I'm not sure what will happen if block 0 is actually being used by another request?

prashantgupta24: I do see this in a scenario where the block table looks like this for 2 requests:
```
block table: [deque([0, 2]), deque([0, 4])]
```

yannicks1: yes, it is exactly how @prashantgupta24 stated it. the padding id has to be real allocated memory, but it can be used or free. felt natural to use 0 for padding otherwise i would have to check if the padding id is in the `self.block_pool` (e.g. <= `self.n_blocks` - 1). Reserving an extra physical mem block for padding feels like an overkill/waste to me. But I will add some comments to make this clearer:)

maxdebayser (approved):
LGTM

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -859,7 +871,7 @@ def _prepare_prompt(
                 sampling_params=sampling_params,
                 generator=generator,
                 output_token_ids=[],
-                left_padding=left_padding)
+                left_padding=self.tkv - max_prompt_len)
```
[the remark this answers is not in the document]
prashantgupta24: Also if possible a short demo of how the tkv actually changes based on this optimization would be much appreciated as well! The original demo in which you showed how CB works with the table visualization helped me a lot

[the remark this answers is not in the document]
prashantgupta24: > did the refactoring in #335 , would be nice to get it in asap, do rebase this PR and make it more readable. @prashantgupta24 @maxdebayser

That's pretty awesome! If you could rebase this PR on that it would be great

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -924,16 +940,28 @@ def _prepare_decode(
...
-            # adding new blocks for current sequence if needed
-            if self.tkv // self.block_size + 1 > len(
-                    self.req_ids2blocks[req_id]):
-                self.req_ids2blocks[req_id].append(self.block_pool.popleft())
-            block_table.append(self.req_ids2blocks[req_id])
+
+            # filling block table with padding blocks to make it rectangular
+            # Note: the padding block id 0 here is chosen arbitrarily, it can
+            # be any allocated block id on the Sypre card (has to be in range
+            # [0, self.n_blocks - 1]). Further it also be an block id that holds
```
JRosenkranz, line 959: nit: Further, it can also be a block id that holds actual KV cache for another (or the same) sequence.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -924,16 +940,28 @@ def _prepare_decode(
         }
         req_ids = self.input_batch.sorted_requests_ids
 
+        n_blocks = 0  # maximal number of blocks used by any seq in the batch
+        for req_id in req_ids:
+            # adding new blocks if needed
+            if self.tkv % self.block_size == 0:
+                self.req_ids2blocks[req_id].append(self.block_pool.popleft())
+            n_blocks = max(n_blocks, len(self.req_ids2blocks[req_id]))
```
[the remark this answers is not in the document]
yannicks1: after the padding on block level, the block table is of course rectangular...

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -894,11 +906,15 @@ def _prepare_prompt(
...
+        # tkv_prefill aligns with tkv of current decode batch tkv_decode:
+        # tkv_prefill % block_size = tkv_decode % block_size
         # and right padding to align with the next block boundary
         input_tokens, position_ids, mask = self.pad_input_ids(
-            [prompt_token_ids_tensor], min_pad_length=block_padding)
-        mask = mask.unsqueeze(1)
+            [prompt_token_ids_tensor],
+            min_pad_left=left_padding_tkv,
+            min_pad_right=right_padding_tkv)
+        mask = mask.unsqueeze(1).contiguous()
```
joerunde, line 917: Is the `.contiguous()` a new requirement here on the mask?

JRosenkranz: @joerunde This may come from the following: https://github.com/foundation-model-stack/aiu-fms-testing-utils/blob/bd1090ef7aa1323b28587533d29935827b7a6755/aiu_fms_testing_utils/utils/paged.py#L263

```
This view will result in a discontiguous tensor (creates a new graph during compile). For this reason, we must explicitly make contiguous
```

If you are however not doing something like `mask[seq_index][:, -current_tkv:, -current_tkv:]`, it is probably not needed.

yannicks1: yeah, just want to be on the safe side here.  Would be tricky to debug if our mask should become discontiguous for any reason ...

yannicks1:
bot:test
MARKERS="spyre"

joerunde:
bot:test
MARKERS="spyre"

waleedqk:
bot:test
MARKERS="spyre"

PR #274, opened 2025-07-01
✨ Support prompt logprobs with static batching

joerunde:
# Description

This PR enables prompt logprobs with static batching, at batch size 1 only. This enables some experimentation and model evaluation tasks on spyre hardware.

For static batching, the requires us to warm the model up with `only_last_token=False`, which passes back the hidden state tensors for the entire (padded) prompt. This is a big performance penalty, and is also only supported on spyre cards with batch size 1 currently.

So, this PR introduces an environment flag `VLLM_SPYRE_ENABLE_PROMPT_LOGPROBS` which must be set to 1 to enable prompt logprobs. At bootup, we check to ensure that we're running in static batching mode with max batch size 1, and fail to boot otherwise. All requests that ask for `prompt_logprobs` will be rejected unless prompt logprobs are enabled. This is different than the behavior today, where requests always return `[None]` for prompt logprobs.


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -267,13 +272,85 @@ def update_states(self, scheduler_output: SchedulerOutput):
...
+            req_idx = self.get_req_id_to_index(model_inputs.is_prompt)[req_id]
+            hidden_states = hidden_states[req_idx]
+            # The offset needs to account for the left padding that static
+            # batching applies.
+            # TODO: To support continuous batching the offset needs to be
+            # calculated differently.
+            offset = hidden_states.shape[0] - num_prompt_tokens
+
+            prompt_hidden_states = hidden_states[offset:offset + num_logits]
+            logits = self.model.compute_logits(prompt_hidden_states, None)
```
yannicks1, line 332: we already compute the logits in the forward pass on line 407. Would it be possible to reuse this tensor instead of recomputing? 
something like: 
```
def _get_prompt_logprobs_dict(
        self,
        logits: torch.Tensor,
        model_inputs: ModelForwardInputs,
    ) -> dict[str, Optional[LogprobsTensors]]:
        ...
        for loop:
            ...
            # Get the logits corresponding to this req's prompt tokens.
            req_idx = self.get_req_id_to_index(model_inputs.is_prompt)[req_id]
            logits = logits[req_idx]
            # The offset needs to account for the left padding that static
            # batching applies.
            # TODO: To support continuous batching the offset needs to be
            # calculated differently.
            offset = logits.shape[0] - num_prompt_tokens

            logits = logits[offset:offset + num_logits]
            ...
```

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -970,6 +1047,16 @@ def get_req_id_to_index(self, is_prefill: bool) -> dict[str, int]:
...
 
+    def no_prompt_logprob(self, is_prefill: bool) -> bool:
+        if is_prefill:
+            return self.prefill_batch.no_prompt_logprob
+        # If we're not running a prefill then this is a decode-only batch
+        return True
+
+    def get_num_prompt_logprobs(self, is_prefill: bool) -> dict[str, int]:
+        return (self.prefill_batch.num_prompt_logprobs
+                if is_prefill else self.input_batch.num_prompt_logprobs)
```
yannicks1, line 1058: Are the two branches here needed? Do we ever return `self.input_batch.num_prompt_logprobs` (for decode)? differently asked, are we not exiting in line 295 if `self.no_prompt_logprob(model_inputs.is_prompt)` is True for decodes?

tests/e2e/test_spyre_prompt_logprobs.py
```diff
@@ -0,0 +1,119 @@
...
+    '''
+    This test uses expected prompt_logprob values recorded from running vllm
+    v0.9.0 on an A100 GPU.
+    These are valid for the model "ibm-ai-platform/micro-g3.3-8b-instruct-1b"
+    only.
+    '''
+    model = "ibm-ai-platform/micro-g3.3-8b-instruct-1b"
+    num_prompt_logprobs = 5
+
+    json_path = Path(__file__).parent.parent / "expected_prompt_logprobs.json"
```
yannicks1, line 33: is there a way to get these log prompts from the huggingface model instead of reading hard coded values here?

maxdebayser: If it's a small model, we could execute on CPU perhaps with transformers as reference implementation.

tests/expected_prompt_logprobs.json
```diff
@@ -0,0 +1,744 @@
+{
+    "Hello darkness my old friend": [
```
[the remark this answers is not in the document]
joerunde: Replaced with chicken soup prompts :(

But I can add it to the list of stock prompts if you want to keep it

tests/e2e/test_spyre_prompt_logprobs.py
```diff
@@ -0,0 +1,119 @@
...
+    with open(json_path) as f:
+        expected_prompt_logprobs: dict[str, list] = json.load(f)
+
+    monkeypatch.setenv("VLLM_USE_V1", 1)
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+    monkeypatch.setenv("VLLM_SPYRE_ENABLE_PROMPT_LOGPROBS", 1)
+    llm = LLM(model)
```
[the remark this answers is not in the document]
joerunde: For `max_model_len`, `max_num_seqs`, and `block_size` I'd actually rather not set them here because right now this test is specifically only for static batching, so they're unused. Once we support continuous batching though, I agree we can and should set those explicitly.

I could go ahead and parameterize this test for multi-aiu if we want, that would be helpful for it to run when it detects multiple cards.

For `tokenizer`, I agree we shouldn't need to ever set it differently, and if we did it wouldn't work to set the same model name anyway. But... there's no harm in setting it for consistency with other tests I guess

vllm_spyre/v1/worker/spyre_model_runner.py
joerunde: True true, this is very likely to only ever be called after a `no_prompt_logprob` guard so it's probably fine to simplify

joerunde: (this also doesn't work on cb now anyway)

tests/e2e/test_spyre_prompt_logprobs.py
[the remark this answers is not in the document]
joerunde: tensor parallel works 🎉

joerunde:
@yannicks1 @maxdebayser This should be ready for another look

I can follow up with enabling this for CB, that will require touching bits of the model that I think Yannick is currently working on to replace `only_last_token=False` with the token indices that we want to get back from the model

yannicks1 (approved):
LGTM

tests/e2e/test_spyre_tensor_parallel.py
```diff
@@ -1,4 +1,4 @@
-"""Verification of vLLM output by comparing with HF
+"""Tests validating the correctness and configuration of prompt_logprobs.
```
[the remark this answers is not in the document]
joerunde: hah, yeah I clicked on the wrong file, sorry. I'll fix it

PR #276, opened 2025-07-02
:alembic: add more prompts and cpu validation

joerunde:
# Description

This updates the offline cb example test script to:
- add more varied prompts for better testing with different `--max-num-seqs` values
- add a `--num-prompts` option to send more prompts to the model
- add a `--compare-with-cpu` flag to run the same prompts against the transformers model on cpu and ensure the text is identical


joerunde:
Alright, I'll go ahead and merge and we'll get this over to the compiler team today

PR #278, opened 2025-07-03
Duplicate the SamplingMetadata class

maxdebayser:
We were previously reusing the GPU SamplingMetadata class but there have been incompatible changes upstream (PR https://github.com/vllm-project/vllm/pull/16728)

Since it's not clear for now whether we want, should or can reuse the LogitsProcessor implementation as is, I'm making temporarily making a copy of the old versions of the files that we need for the spyre backend.

This won't affect any features for now since the vllm change was an internal refactoring without UX impact.

Follow-up issue to give a definitive solution: [deleted]

### DIFF
--- a/vllm_spyre/v1/sample/metadata.py
+++ b/vllm_spyre/v1/sample/metadata.py
@@ -7,8 +7,6 @@ from typing import Optional
 
 import torch
 
-# This is a copy of the vLLM vllm file prior to PR
-# https://github.com/vllm-project/vllm/pull/16728
 # TODO: Figure out if we want to apply the LogitsProcessor
 # approach here and whether we want to reuse the code. That
 # would require some refactoring since processors like the


joerunde:
> We were previously reusing the GPU SamplingMetadata class

I don't think that class is gpu-specific though, right? It's the sampling metadata used by all in-tree platforms supported by vllm

prashantgupta24 (commented):
Any chance we can rebase this to get merged to `vllm-main-updates` instead of `main`?

prashantgupta24:
> > Any chance we can rebase this to get merged to `vllm-main-updates` instead of `main`?
> 
> Sorry, I'm not sure I understand. These commits here are already in `vllm-main-updates`, rght?

You're right, I said that just in case we wanted to merge any new changes

PR #279, opened 2025-07-03
[CB][Tests] Add CB online test and refactor multi tests

rafvasq:
## Description

- Enable CB in pytest fixture for online server
- Adds CB online server test 
- Consolidates TP tests
- Updates prompts in all tests to use chicken soup prompts


tests/e2e/test_spyre_basic.py
```diff
@@ -56,6 +56,7 @@ def test_output(
         sampling_params=vllm_sampling_params,
         tensor_parallel_size=1,
         backend=backend,
+        use_cb=cb,
```
prashantgupta24, line 59: probably need the marker back?

tests/e2e/test_spyre_online.py
```diff
@@ -8,7 +8,10 @@
 @pytest.mark.parametrize("warmup_shape", [[
     (64, 20, 1),
 ]])
-def test_openai_serving(remote_openai_server, model, warmup_shape, backend):
+@pytest.mark.parametrize("cb",
+                         [pytest.param(1, marks=pytest.mark.cb, id="cb"), 0])
+def test_openai_serving(remote_openai_server, model, warmup_shape, backend,
```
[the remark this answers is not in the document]
yannicks1: also `cb` will be unused, right? Am I missing something here?

tests/e2e/test_spyre_online.py
```diff
@@ -51,9 +54,11 @@ def test_openai_serving(remote_openai_server, model, warmup_shape, backend):
 
 @pytest.mark.parametrize("model", get_spyre_model_list(quantization="gptq"))
 @pytest.mark.parametrize("backend", ["sendnn"])
+@pytest.mark.parametrize("cb",
+                         [pytest.param(1, marks=pytest.mark.cb, id="cb"), 0])
 @pytest.mark.parametrize("quantization", ["gptq"])
 @pytest.mark.parametrize("warmup_shape", [[(64, 20, 1)]])
-def test_openai_serving_gptq(remote_openai_server, model, backend,
+def test_openai_serving_gptq(remote_openai_server, model, backend, cb,
```
yannicks1, line 61: again: `warmup_shape`, `backend`, `quantization`, and `cb` are unused

tests/e2e/test_spyre_online.py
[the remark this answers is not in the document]
joerunde: They're used by the `remote_openai_server` fixture, which I do think needs to be updated to read the `cb` param too.

That said we could convert that fixture to a context manager to make it more clear how the test params are used. I think maybe we had it as a fixture so that we could potentially re-use it across tests like upstream vllm does but we also aren't doing that currently.

tests/e2e/test_spyre_basic.py
```diff
@@ -247,3 +247,72 @@ def test_full_batch_scheduling(model: str, backend: str, monkeypatch):
...
+    "What is the weather today like?", "Who are you?"
+]])
+@pytest.mark.parametrize(
+    "warmup_shapes",
+    [[(64, 20, 4)]])  #,[(64,20,8)],[(128,20,4)],[(128,20,8)]])
+# (prompt_length/new_tokens/batch_size)
+@pytest.mark.parametrize("tp_size", [2, 4, 8])
+@pytest.mark.parametrize(
+    "backend", [b for b in get_spyre_backend_list() if "eager" not in str(b)])
+def test_output_multi(
```
joerunde, line 266: I don't think we actually need a separate test for this though, right? Can we parameterize the existing one like this:
```
@pytest.mark.parametrize("tp_size", [param(1), param(2, marks=multi) ...])
```

yannicks1 (approved):
lgtm

PR #281, opened 2025-07-03
Respect vllm logging configs in vllm_spyre

kazunoriogata:
# Description

vLLM has mechanism to change [logging configuration](https://docs.vllm.ai/en/v0.7.2/getting_started/examples/logging_configuration.html) by using environment variables.  However, vllm_spyre does not refers to these environment variables, but it always configure python logging based on the hard coded default of vLLM.  As a result, changing logging configuration results in mixing log messages of different configurations and generates inconsistent logs.

This PR updates initialization of python logger in vllm_spyer to respect the logging configs and use the same logging configuration as vllm.

### DIFF
--- a/vllm_spyre/__init__.py
+++ b/vllm_spyre/__init__.py
@@ -3,8 +3,8 @@ import json
 from logging.config import dictConfig
 from typing import Any
 
-from vllm.logger import DEFAULT_LOGGING_CONFIG
 from vllm.envs import VLLM_CONFIGURE_LOGGING, VLLM_LOGGING_CONFIG_PATH
+from vllm.logger import DEFAULT_LOGGING_CONFIG
 
 __version__ = importlib.metadata.version("vllm_spyre")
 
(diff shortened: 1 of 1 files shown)


joerunde:
Thanks @kazunoriogata!

Do you think it's possible to read the config from the vllm logger instead of reloading the file without validation?

kazunoriogata:
@joerunde Thank you for the comment, but I don't think we can get the configs from vllm logger.

[`_configure_vllm_root_logger`](https://github.com/vllm-project/vllm/blob/main/vllm/logger.py#L92) does not save the config dict. The config dict is thrown away [after calling `dictConfig()`](https://github.com/vllm-project/vllm/blob/a4113b035cd4ee90ec02de0658b75d47f0007ede/vllm/logger.py#L124).  I couldn't find a method to extract logging config or a list of formats/handlers/loggers from an existing logger (e.g., the root logger).  So, I think we need to reloading the file.

Since errors like [file-not-found](https://github.com/vllm-project/vllm/blob/a4113b035cd4ee90ec02de0658b75d47f0007ede/vllm/logger.py#L107) or [invalid-format](https://github.com/vllm-project/vllm/blob/a4113b035cd4ee90ec02de0658b75d47f0007ede/vllm/logger.py#L114) should have raised an exception and should not reach the vllm-spyre logger initialization, I think we don't need the same file validation again.

joerunde:
Nice, thanks for looking into it! I'll get this merged

PR #273, opened 2025-07-01
🐛 req_ids is now a list in vllm:main

prashantgupta24:
# Description

Issue #271, opened 2025-07-01
Fix plugin code agains vLLM main


yannicks1:
### Fix plugin code agains vLLM main

tests agains main are currently failing due to 
`TypeError: CachedRequestData.__init__() got an unexpected keyword argument 'req_id'`

PR #283, opened 2025-07-04
vllm main updates

prashantgupta24:
# Description

This branch has a fix for:
- Caching the token_ids (now the new tokens are cached in `execute_model` instead of `update_states`. This is because of https://github.com/vllm-project/vllm/pull/20291. )
- Changes from the `CachedRequestData` (https://github.com/vllm-project/vllm-spyre/pull/273)

## Related Issues

Fix for https://github.com/vllm-project/vllm-spyre/issues/271

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
           - name: "default"
             repo: ""
           - name: "vLLM:main"
-            repo: "git+https://github.com/vllm-project/vllm@02cabff207ca68094a73ba21296c82cdbcb1d1a5"
+            repo: "git+https://github.com/vllm-project/vllm@c1909e7e8ccd2037e76536a8e726120c85d3754e"
         test_suite:
           - name: "static batching"
             markers: "cpu and decoder and not cb"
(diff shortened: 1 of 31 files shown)


prashantgupta24:
> quick question before reviewing this one thoroughly: Do we want to ensure backwards compatibility (I see a couple of copied files from upstream, but also failing test with missing `make_empty` attribute for `CachedRequestData`). If yes down to which version do we ensure compatibility, if not we should enforce a minimum version which does not fail the test...

I think the general consensus is that we want to pin to `vllm>=0.9.2`, we were just waiting for vllm to release the version

joerunde:
@yannicks1 Yeah we talked very briefly about it last Thursday, our assertion here is that our current delivery mechanism is images that bundle vllm-spyre with vllm, so we're the ones that have to worry about building images with compatible versions.

The small risk here is that there could be a regression in vllm 0.9.2 with some model or models that we then can't easily roll back, but we haven't matured this stack yet to GA support anywhere so there's currently no chance of that causing a product regression

joerunde (approved):

PR #270, opened 2025-07-01
[do not merge][CB] additional prefill in warmup to fix TTFT

yannicks1:
### [CB] additional prefill in warmup to fix TTFT
TTFT for the first prefill when running continuous batching was way bigger than for second prefill.
Doing an additional prefill during warmup outside the warmup_context solves this issue.

PR #284, opened 2025-07-07
Update warmup log messages and comments

tjohnson31415:
# Description

This came out of a follow up to https://github.com/vllm-project/vllm-spyre/pull/270 to determine why an extra Prefill was necessary after using `warmup_mode`. I learned that the extra Prefill is required to deploy the compiled graph to the Spyre device. This PR does not change any functionality, but updates logging and documentation around warmup to make this clearer.

### DIFF
--- a/vllm_spyre/v1/worker/spyre_worker.py
+++ b/vllm_spyre/v1/worker/spyre_worker.py
@@ -68,7 +68,7 @@ class SpyreWorker(WorkerBaseV1):
 
         num_shape_combinations = len(self.spyre_warmup_shapes)
         logger.info(
-            "[WARMUP] Starting warmup for %d "
+            "[WARMUP] Starting for %d "
             "prompt/decode/batchsize-shape combinations...",
             len(self.spyre_warmup_shapes))
         all_warmup_start_t = time.time()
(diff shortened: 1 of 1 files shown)


tjohnson31415:
I took the liberty of updating most of the warmup logging: using a `[WARMUP]` tag in the message, more consistent formating, and combining adjacent log messages. But let me know if the preference would be to limit the changes to just update the "additional prefill" CB step.

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -594,10 +593,9 @@ def _warmup_spyre_fixed_size(self, prompt_len, num_decode_tokens,
                               batch_size=batch_size,
                               max_tokens=num_decode_tokens,
                               prompt_len=prompt_len)
-        logger.info("Warmup finished.")
         logger.info(
-            "Warmup took %.3fs (for prompt length %d and max output tokens %d)",
-            warmup_total_t, prompt_len, num_decode_tokens)
+            "[WARMUP] Warmup for prompt length %d and max output tokens %d "
```
prashantgupta24, line 597:
```suggestion
            "[WARMUP] for prompt length %d and max output tokens %d "
```

yannicks1 (approved):
lgtm, if possible address @prashantgupta24 nits..

joerunde:
get in!

PR #285, opened 2025-07-07
Small updates for cb tests

joerunde:
# Description

- ~~Runs the basic continuous batching test at batch size 1 to ensure that's a valid configuration (we still pad to batch size 2 internally to function on spyre)~~
- Sets the max batch size to 2 (from 8) on the abort test for continuous batching. We're only targeting bs=4 for the near term future so we shouldn't set 8, and 2 is much more stable than 4 currently.


prashantgupta24:
bot:test
MARKERS="cb and spyre"

prashantgupta24:
bot:test
MARKERS="cb and spyre"

PR #286, opened 2025-07-08
[CI] Tests for graph comparison between vllm and AFTU

wallashss:
# Description

This PR adds tests that compare graphs generated by vLLM vs AFTU (aiu-fms-testing-utils). The tests do an inference with vLLM and execute inference.py from AFTU. For both execution, the tests dump the compiled graphs to files by setting the env variable `DEE_DUMP_GRAPHS=name`. Then, they compare the corresponding graphs generated for both method.

The main limitation of this implementation is AFTU is not a proper package. So, I had to make some workarounds:
- ~~I added an environment variable: `VLLM_SPYRE_TEST_AFTU_SCRIPTS_DIR`. If set, we can use to inform where is the scripts dirs (where it has the `inference.py`) to copy from there and execute the script within the tests.~~ 
- If the above env var is not set, then we try to get from the install by importing the module and using the `aiu_fms_testing_utils.__file__` variable to find the directory.

I added a marker `aftu` to ease select the tests of this PR, and also to filter out from other test suite.


pyproject.toml
```diff
@@ -150,6 +151,7 @@ dev = [
     "pytest-timeout==2.3.1",
     "requests==2.32.3",
     "sentence-transformers==3.4.1",
+    "aiu_fms_testing_utils@git+https://github.com/foundation-model-stack/aiu-fms-testing-utils.git#1a77f630104a5661fff554164c1e536ea08393e3"
```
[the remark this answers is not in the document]
wallashss: updated

tests/aftu/graph_compare_utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+    for i, s in symbol_map.items():
+        content = content.replace(s, f'S#{i}')
+
+    return content
+
+
+def collect_graph_files(input_dir: str) -> dict[str, tuple[str, str]]:
+    # Get G1 graphs, it assumes the input_dir has the folder export_dtcompiler
+    # where are the files
```
rafvasq, line 53: nit: something like

```suggestion
    # Get G1 graphs. 
    # Assumes the input_dir contains 'export_dtcompiler' holding '.ops' files.
```

tests/aftu/graph_compare_utils.py
```diff
@@ -0,0 +1,160 @@
...
+    # Replace value
+    content = re.sub(r'values: ([0-9a-fA-F]{2}\s*)+', 'values: $$', content)
+
+    # Silly regex to find all s#.
+    # We are only considered those that surrounds by space (whole word)
+    # or started with space and terminated with comma
+    # examples:
+    # ' s1 '
+    # ' s1,'
+    # ' s1 s2 '
```
rafvasq, line 35: just a nit: comment could be a bit more succinct, something like:

```suggestion
    # Regex to find all 's#' patterns surrounded by spaces,
    # or starting with a space and ending with a comma.
    # Examples: ' s1 ', ' s1,', ' s1 s2 '
```

tests/aftu/test_compare_graphs.py
```diff
@@ -0,0 +1,176 @@
...
+                                 get_model_path,
+                                 run_inference_py_and_get_graphs)
+from spyre_util import (generate_spyre_vllm_output, get_chicken_soup_prompts,
+                        get_spyre_model_list)
+from vllm import SamplingParams
+
+
+@pytest.mark.aftu
+@pytest.mark.parametrize("model", get_spyre_model_list())
+@pytest.mark.parametrize("backend", ["sendnn"])
```
joerunde, line 21: If this test can only run on the `sendnn` backend then we don't need to parameterize it, we can just set the backend as sendnn in the test.

Also, we should mark all tests running with the sendnn backend with `pytest.mark.spyre` for consistency with the rest of the test suite

tests/aftu/graph_compare_utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+def diff_graph(a_filepath, a_file, b_filepath, b_file) -> Iterator[str]:
+    return difflib.unified_diff(a_file.split("\n"),
+                                b_file.split("\n"),
+                                fromfile=a_filepath,
+                                tofile=b_filepath)
+
+
+def get_aftu_script_dir() -> str:
+    # TODO: since AFTU is not a lib yet, this function does the best
```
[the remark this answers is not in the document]
wallashss: Yeah... I think I saw somewhere that someone said something like "aftu it is not lib" (or a package that we can use everything just by pip installing), and that stuck in my head, but maybe I misread that. 

Anyway, the last time I checked the inference.py is still not included by installing using the git url, so the code and comment is still valid, I guess.

pyproject.toml
```diff
@@ -131,6 +131,7 @@ markers = [
     "multi: Tests that require >1 cards",
     "utils: Tests for utility functions",
     "worker: Tests for worker logic",
+    "aftu: Tests to compare graphs from aiu-fms-testing-utils",
```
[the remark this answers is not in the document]
joerunde: >Let the pytest collect these tests and check in the test if everything is setup to run, and otherwise skip them?

Yup! That's worked really nicely for the TP test cases so far

tests/aftu/test_compare_graphs.py
```diff
@@ -0,0 +1,174 @@
...
+    model: str,
+    max_num_seqs: int,
+    monkeypatch: pytest.MonkeyPatch,
+    runtime_xfail,
+):
+    """Test that the spyre worker correctly outputs
+    continuous batches of requests by comparing to HF"""
+
+    if max_num_seqs > 2:
+        runtime_xfail("CB failures expected for batch size > 2")
```
joerunde, line 34: This should have since been fixed, let's make sure the test passes with --max-num-seqs 4

tests/aftu/graph_compare_utils.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def get_model_path(model_name_or_path):
+    is_local = os.path.isdir(model_name_or_path)
+    model_path = model_name_or_path
+    # Get location of model from HF cache.
+    if not is_local:
+        model_path = download_weights_from_hf(
+            model_name_or_path=model_path,
+            cache_dir=None,
```
joerunde, line 153: I was confused here but I see from vllm's docs that `None` here defaults to the hf cache

tests/aftu/graph_compare_utils.py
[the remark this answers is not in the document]
joerunde: Yeah, looks like the work is currently in progress there to add a pyproject.toml and ship it on pypi. I guess this is the best we can do now- we can open a PR later over there to include the scripts with the release so that this isn't so cumbersome to do here (or manually)

tests/aftu/test_compare_graphs.py
joerunde: Looks like it passes for me on a dev pod 👍 

Should be good to remove this check

tests/aftu/test_compare_graphs.py
```diff
@@ -0,0 +1,174 @@
...
+                                 run_inference_py_and_get_graphs)
+from spyre_util import (generate_spyre_vllm_output, get_chicken_soup_prompts,
+                        get_spyre_model_list)
+from vllm import SamplingParams
+
+
+@pytest.mark.spyre
+@pytest.mark.aftu
+@pytest.mark.parametrize("model", get_spyre_model_list())
+@pytest.mark.parametrize("max_num_seqs", [2, 4],
```
prashantgupta24, line 22: I wonder if we can have only 4 here to speed things up unless I'm missing something?

joerunde (commented):
LPGTM, and passes on my machine. Just a couple comments on cleaning up a bit and I think we're good to go

tests/aftu/test_compare_graphs.py
joerunde: Probably? We initially had 2/4 on all the CB tests because batch size 4 didn't work

tests/aftu/test_compare_graphs.py
```diff
@@ -0,0 +1,174 @@
...
+    aftu_graphs = run_inference_py_and_get_graphs(inference_py_args)
+
+    # VLLM
+    prompts = get_chicken_soup_prompts(4)
+
+    max_new_tokens = warmup_shape[1]
+
+    monkeypatch.setenv("DEE_DUMP_GRAPHS", "vllm_static")
+    # Disable cache to produce the graphs
+    monkeypatch.setenv("TORCH_SENDNN_CACHE_ENABLE", "0")
```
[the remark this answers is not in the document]
wallashss: Yes, I do that in this method:

```py
def run_inference_py_and_get_graphs(
         inference_py_args: list[str],
         extra_env: Optional[dict[str,
                                  str]] = None) -> dict[str, tuple[str, str]]:
     with tempfile.TemporaryDirectory() as tmpdir:

         env = os.environ.copy()
         env.update({
             "DEE_DUMP_GRAPHS": "aftu",
             "TORCH_SENDNN_CACHE_ENABLE": "0"
         })
```

tests/aftu/test_compare_graphs.py
```diff
@@ -0,0 +1,174 @@
...
+
+    assert compare_graphs(vllm_graphs, aftu_graphs)
+
+
+@pytest.mark.spyre
+@pytest.mark.aftu
+@pytest.mark.parametrize("model", get_spyre_model_list())
+@pytest.mark.parametrize("warmup_shape",
+                         [(64, 5, 1), (64, 5, 2),
+                          (64, 5, 4)])  # (prompt_length/new_tokens/batch_size)
```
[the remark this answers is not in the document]
wallashss: I set just a single warmup shape.

tests/aftu/graph_compare_utils.py
```diff
@@ -0,0 +1,157 @@
...
+
+        process = Popen(inference_py_args,
+                        stdout=PIPE,
+                        stderr=PIPE,
+                        env=env,
+                        cwd=tmpdir)
+
+        process.communicate()
+
+        aftu_graphs = collect_graph_files(tmpdir)
```
maxdebayser, line 142: Before collecting the graphs, we should call `process.wait(timeout=600)` to make sure that the process is finished.

maxdebayser: If this raises a timeout exception, the process should be killed. Otherwise it can cause problems for subsequent tests.

tests/aftu/test_compare_graphs.py
```diff
@@ -0,0 +1,177 @@
...
+        "--batch_size",
+        str(max_num_seqs), "--compile_dynamic_sendnn", "--attention_type=paged"
+    ]
+
+    extra_env = {
+        "VLLM_DT_MAX_CONTEXT_LEN": str(max_model_len),
+        "VLLM_DT_MAX_BATCH_SIZE": str(max_num_seqs)
+    }
+    aftu_graphs = run_inference_py_and_get_graphs(inference_py_args,
+                                                  script_dir, extra_env)
```
maxdebayser, line 54: I think this could be factored out as a function or fixture.

tests/aftu/graph_compare_utils.py
wallashss: Thanks for the feedback. I did a refactoring and guess it got more simple and robust. I am using now `subprocess.run` and put some catches for timeout and process that doesn't return code 0. 

> If this raises a timeout exception, the process should be killed. Otherwise it can cause problems for subsequent tests.

I reviewed the documentation and both run and Popen.communicate wait/or kill process if timeout, so I guess we are fine there.

tests/aftu/test_compare_graphs.py
maxdebayser: Ok, that makes sense.

tests/aftu/graph_compare_utils.py
maxdebayser: Awesome!

PR #290, opened 2025-07-08
Integrate upstream logits processors

maxdebayser:
At first it wasn't obvious if it would be easy to integrate the changes of PR https://github.com/vllm-project/vllm/pull/16728 so initially I added PR that copies the sampler files previous to that PR in vllm-spyre. But actually it's easier than I thought because the sampler code is not compiled to the AIU, only the model forward is. 

Currently in the MinP processor there is a tensor for the cpu and for the device. Since only the model forward runs on the AIU, both tensors end up on the CPU, which means that there is an unnecessary copy from one to the other, but the result is still correct.

There is a future upstream PR that will generalize the Logits processor to other sampling parameters:

https://github.com/vllm-project/vllm/pull/19912


maxdebayser:
@joerunde , I've opened an issue for the follow up PR that will add tests: https://github.ibm.com/ai-foundation/aiu-app-sw-tracker/issues/821

maxdebayser:
@joerunde, I've opened an issue for the follow-up PR that will add tests: https://github.com/vllm-project/vllm-spyre/issues/295

prashantgupta24:
bot:test
MARKERS="cb and spyre"

prashantgupta24:
bot:test

maxdebayser:
bot:test

maxdebayser:
The test run 129 successfully executed all tests

PR #291, opened 2025-07-08
feat: detect CPUs and configure threading sensibly

tjohnson31415:
# Description

When running inside a container, it is easy to oversaturate the CPU because the default for threading configurations is to use all CPUs on the system. This is especially true with `--tensor-parallel` becuase each worker creates its own thread pool. This PR adds code to detect the CPU limits allocated to a container (falling back to a simple `os.cpu_count()`) and considers the distributed world size to provide better default behavior of `num_cpus / num_workers` per process. There is also an internal configuration affecting threading during graph compilation called `DT_PARALLEL_THREADS` that does not check `OMP_NUM_THREADS`.

The PR adds a `VLLM_SPYRE_UPDATE_THREAD_CONFIG` env var configuration that is on by default but can be disabled to check and warn about threading configurations but not override anything.


vllm_spyre/platform.py
```diff
@@ -309,3 +312,113 @@ def _get_matching_warmup_shapes(
...
+        """
+
+        # The quay.io/ibm-aiu/spyre-base image includes shell scripts that
+        # automatically set OMP_NUM_THREADS to the result of `nproc --all`.
+        #
+        # vLLM also already has logic around threading to be aware of,
+        #  - sets TORCHINDUCTOR_COMPILE_THREADS=1 (https://github.com/vllm-project/vllm/blob/baba0389f7e810a361fff5229ce20c2d5a2b1fac/vllm/env_override.py#L38-L39)
+        #  - it will set OMP_NUM_THREADS=1 when using multiple workers (https://github.com/vllm-project/vllm/blob/baba0389f7e810a361fff5229ce20c2d5a2b1fac/vllm/executor/multiproc_worker_utils.py#L304)
+        #  - has configurations for OMP thread binding (https://github.com/vllm-project/vllm/blob/baba0389f7e810a361fff5229ce20c2d5a2b1fac/vllm/envs.py#L435-L438)
+        #    - the bind attempts to detect NUMA nodes (https://github.com/vllm-project/vllm/blob/baba0389f7e810a361fff5229ce20c2d5a2b1fac/vllm/v1/worker/cpu_worker.py#L111)
```
sducouedic, line 334: thanks for the pointers

vllm_spyre/platform.py
```diff
@@ -309,3 +312,113 @@ def _get_matching_warmup_shapes(
...
+
+        # Always print current env for awareness
+        threading_envs = [
```
joerunde, line 337: Thoughts on setting these as "constants" in this file?

wallashss:
bot:test
TEST_FILE=tests/e2e/test_spyre_basic.py MARKERS="spyre"

wallashss:
bot:test
TEST_FILE=tests/e2e/test_spyre_basic.py MARKERS="spyre"

wallashss (approved):
I've reviewed and tested the code. LGTM! 

Thanks @tjohnson31415

PR #301, opened 2025-07-10
:bug: fix tensor parallel

joerunde:
# Description

Fixes a bug introduced in #283 where the non-driver workers did not cache the output tokens for the next decode iteration


prashantgupta24:
bot:test
MARKERS="spyre and not quantized and not multi and not embedding"

prashantgupta24:
actually let me try to see if CB works with TP first before merging

PR #304, opened 2025-07-11
add long context example

maxdebayser:


maxdebayser:
Since the example doesn't break any existing code and it's needed ASAP, I'm force merging it for now.

PR #307, opened 2025-07-14
:bug: Workaround ray issue in tests

joerunde:
~~Testing whether or not torch 2.7 will cause the EADDRINUSE failures on vllm 0.9.2~~

Edit: The issue appears to be with the shutdown logic in vllm, with TP running the model executor is not shut down when an `LLM` is garbage collected. This is caused by a circular reference in vllm.executor.ray_utils when ray is not installed. This PR adds a temporary workaround to force the engine shutdowns after each test.

Also updating to torch 2.7.1 everywhere since our next release will have 2.7.1 in the base image as well.

### DIFF
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -23,6 +23,23 @@ ISCLOSE_REL_TOL_CPU = 0.2
 ISCLOSE_REL_TOL_SPYRE = 0.35
 
 
+def force_engine_shutdown(llm: LLM):
+    """
+    🌶️🌶️🌶️
+    This hack is here because of an issue in vllm 0.9.2+ where a circular
+    reference occurs in vllm.executor.ray_utils if ray is not installed. This
+    circular reference holds a copy of the vllm config which contains a
+    reference to the LLM, which means it can never be garbage collected.
+    Since vllm.LLM relies on garbage collection to shut down its engine, the
+    engine never shuts down. When running tensor parallel workloads, if the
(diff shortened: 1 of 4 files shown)


joerunde:
Both `pytest tests -m "spyre and multi" -sv` and `pytest tests -m "spyre and multi" -sv --forked` pass on a machine with 2 spyre cards

tjohnson31415 (approved):
Nice! LGTM

prashantgupta24 (approved):
🌶️

tests/spyre_util.py
```diff
@@ -23,6 +23,23 @@
...
+    circular reference holds a copy of the vllm config which contains a
+    reference to the LLM, which means it can never be garbage collected.
+    Since vllm.LLM relies on garbage collection to shut down its engine, the
+    engine never shuts down. When running tensor parallel workloads, if the
+    engine is never shut down then the TP worker processes are never killed.
+    When the TP worker processes are held open, all future attempts to create a
+    new engine will fail with an EADDRINUSE error.
+    🌶️🌶️🌶️
+    """
+    llm.llm_engine.engine_core.shutdown()
```
prashantgupta24, line 40: I wonder if we could've added this to the `cleanup_fixture` 🤔

tjohnson31415:
Just to link it, here's an upstream vLLM PR to change how the ray_import_err is handled to remove the ref and fix garbage collection:
https://github.com/vllm-project/vllm/pull/21149

PR #310, opened 2025-07-15
:bug: use right attention name

joerunde:
@JRosenkranz says this should be sdpa_causal, but it doesn't matter since we always pass a mask


joerunde:
bot:test

yannicks1 (approved):
lgtm

joerunde:
Works on spyre cards!

PR #311, opened 2025-07-15
[ppc64le] Introduce ppc64le benchmarking scripts

Daniel-Schenker:
This PR introduces two new scripts to the examples directory in the online inference section.

spyre_vllm_benchmark.py -> Benchmarking script for running model inference and collecting performance results.

spyre_vllm_setup_container.sh -> Setup script to initialize container environment for "bare metal" vllm inference.


examples/online_inference/spyre_vllm_setup_container.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash -e
+
+# This script sets up the runtime environment for launching a vLLM API
+# server using Spyre AIU cards.
+# Used for local development and testing in tandem with podman run command.
+# Not for use on an openshift cluster.
+# 1. Validates TORCH_SENDNN cache settings.
+# 2. Detects and configures available AIU devices.
+# 3. Activates the Python virtual environment if not already active.
+# 4. Launches the vLLM server with the computed arguments.
```
joerunde, line 10: Do we want this to be included in the install of vllm-spyre, or do we want it only available with local repo checkouts?

examples/online_inference/spyre_vllm_benchmark.py
```diff
@@ -0,0 +1,437 @@
+#!/usr/bin/env python3
```
joerunde, line 1: Nice, this is a neat little benchmark example

joerunde:
I haven't tested the shell script here since I only have openshift clusters available to me to test on, but it looks reasonable to me

examples/online_inference/spyre_vllm_setup_container.sh
Daniel-Schenker: My plan was to make it available with local repo checkouts mainly, since we want our ppc64le images to include these scripts, I will create a step in our build that copies these scripts into the container image.

joerunde:
Sorry I rarely come back to this- merging it in!

PR #245, opened 2025-06-19
[Priority merge] NewRequestData parameter introduced in vllm upstream

sducouedic:
Temporary hack until the parameter makes it to a new release version. Needs to be merged first for the tests on the other PRs to pass.

(PS: this was actually the error after fixing the merge conflict in PR #240, which had nothing to do with the conflict)

PR #313, opened 2025-07-16
removing legacy backward compatibility

yannicks1:
### removing legacy backward compatibility
due to the requirement `vllm>=0.9.2` we can safely revert some changes by #245 .


maxdebayser (approved):
LGTM

PR #314, opened 2025-07-16
:bug: fix max tokens for continuous batching

joerunde:
# Description

Fixes a bug where continuous batching was always limited to 1 token


prashantgupta24 (commented):
Any chance we can add a failing test to catch it in the future?

joerunde:
Issue added to our internal board to track that!

@gmarinho2 is on the case

prashantgupta24 (approved):
Makes sense to me, based on https://github.com/vllm-project/vllm/pull/18557/files#diff-0e88b32558603cfb041204f63d2f598aa9b102de14688bb31d23d500722209e3

maxdebayser:
I think this solution is correct.

PR #315, opened 2025-07-16
[Fix][Tests] TP param used in tests unconditionally

rafvasq:
# Description

`--tensor-parallel-size` and `tp_size` flag being used regardless of the the block checking if `tp_size` exists.


yannicks1 (approved):
thanks for catching that.

PR #316, opened 2025-07-16
feat: enable FP8 quantized models loading

rafvasq:
# Description

- "Enable" FP8 pre-quantized models (compressed-tensors) to load via `fms_mo`


vllm_spyre/platform.py
```diff
@@ -40,7 +40,7 @@ class SpyrePlatform(Platform):
     # "spyre" device_name no longer worked due to https://github.com/vllm-project/vllm/pull/16464
     device_name: str = "cpu"
     device_type: str = "cpu"
-    supported_quantization: list[str] = ["gptq"]
+    supported_quantization: list[str] = ["gptq", "fp8", "compressed-tensors"]
```
maxdebayser, line 43: I don't see a mention to `compressed-tensors` anywhere else in this PR.

tests/e2e/test_spyre_online.py
```diff
@@ -57,6 +57,7 @@ def test_openai_serving(remote_openai_server, model, warmup_shape, backend,
         assert "warmup" in str(e)
 
 
+@pytest.mark.quantized
```
yannicks1, line 60: why is this line needed? you did not change anything in the gptq path, right ? shouldn't it get marked anyway in line 507 of spyre_util.py?

vllm_spyre/platform.py
yannicks1: do I assume correctly, that if I provide a model in the compressed tensor format and set 
`LLM(..., quantization="compressed-tensors")` this should run by going through the else path (neither fp8 nor gptq) like any other not-quantized model?

yannicks1 (commented):
great work! 
One last question: Do we want to add a test for this? I see you have added support in `tests/spyre_util.py`, but I don't see any test...

PR #322, opened 2025-07-17
[SB] relax constraint on min number of new tokens

yannicks1:
### [SB] relax constraint on min number of new tokens

this is relaxing an old constraint on the number of requested new tokens having to be a min of 3. Turns out it is only important that during the warmup there is at least **one** decode forward pass. Requesting 1 token runs prefill only during warmup -> compiler crashes (I guess it is expecting two graphs, prefill and decode) . Requesting 2+ tokens does at least one decode during warmup and thus produces a decode graph too -> things run smoothly for 2+ tokens ...

### DIFF
--- a/tests/e2e/test_spyre_basic.py
+++ b/tests/e2e/test_spyre_basic.py
@@ -160,32 +160,14 @@ def test_batch_handling(model: str, backend: str, cb: int,
 
     prompts = get_chicken_soup_prompts(4)
 
-    sampling_params1 = SamplingParams(max_tokens=5,
-                                      min_tokens=5,
-                                      temperature=0,
-                                      ignore_eos=True,
-                                      logprobs=0)
-    sampling_params2 = SamplingParams(max_tokens=20,
-                                      min_tokens=20,
-                                      temperature=0,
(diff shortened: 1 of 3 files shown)


tests/e2e/test_spyre_basic.py
```diff
@@ -160,32 +160,14 @@ def test_batch_handling(model: str, backend: str, cb: int,
...
     vllm_sampling_params = [
-        sampling_params1,
-        sampling_params2,
-        sampling_params3,
-        sampling_params4,
+        SamplingParams(max_tokens=max_new_tokens[i],
+                       min_tokens=max_new_tokens[i],
+                       temperature=0,
+                       ignore_eos=True,
+                       logprobs=0) for i in range(len(max_new_tokens))
```
sducouedic, line 170:
```suggestion
        SamplingParams(max_tokens=max_tokens_i,
                       min_tokens=max_tokens_i,
                       temperature=0,
                       ignore_eos=True,
                       logprobs=0) for max_tokens_i in max_new_tokens)
```

sducouedic (approved):
LGTM, thanks for fixing this

PR #326, opened 2025-07-18
:bug: fix cb online test

joerunde:
# Description

We have to override max-model-len since long contexts aren't supported yet

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -91,13 +91,18 @@ def remote_openai_server(request):
             "Error setting up remote_openai_server params") from e
 
     if 'cb' in params:
-        max_num_seqs = params['max_num_seqs']
+        max_model_len = params["max_model_len"]
+        max_num_seqs = params["max_num_seqs"]
         env_dict = {
             "VLLM_SPYRE_USE_CB": "1",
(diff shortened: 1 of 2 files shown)


sducouedic:
so these changes fix `this failure`(https://github.com/vllm-project/vllm-spyre/actions/runs/16356395537/job/46215683015) right? In the test it appears that the server was failing during startup, and this **intermittently**. I don't understand why adding the `max_model_len` as parameter would make it pass every time. I mean here, why was it failing sometimes and sometimes working if the problem is such a determinitic parameter with fixed value?

sducouedic:
adding @yannicks1 as reviewer as he reviewed the PR creating this online test: https://github.com/vllm-project/vllm-spyre/pull/279

prashantgupta24 (approved):
All right, good enough for me

sducouedic:
Ah I see! So this will also probably fix the problem on CPU then.

joerunde:
> So this will also probably fix the problem on CPU then

I hope! If it doesn't then we'll need to take a closer look for sure

joerunde:
Yup, vllm initializes its config based on the model configuration: https://huggingface.co/ibm-ai-platform/micro-g3.3-8b-instruct-1b/blob/main/config.json#L16

PR #329, opened 2025-07-22
[CB] Add scheduling tests

sducouedic:


sducouedic:
bot:test
MARKERS="spyre and not quantized and not multi and not embedding"

prashantgupta24 (commented):
Excited for this!

tests/e2e/test_spyre_cb_scheduler_steps.py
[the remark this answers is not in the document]
yannicks1: I think @sducouedic PR can go first and I will rebase my branch. Testing has higher prio than optimizations rn.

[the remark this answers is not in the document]
sducouedic: thanks @yannicks1, PR is ready for review

tests/e2e/test_spyre_cb_scheduler_steps.py
```diff
@@ -470,6 +475,270 @@ def test_two_sequences_finish_same_time_as_new_arrive(
...
+            * 1: len = 49, max tokens = 119, step joining = 0
+            * 2: len = 14, max tokens = 52, step joining = 0
+            * 3: len = 89, max tokens = 104, step joining = 32
+            * 4: len = 9, max tokens = 64, step joining = 131
+    """
+    # TODO change to 65 max_tokens for last prompt if ever possible
+
+    seqs_max_tokens = [119, 52, 104, 64]
+    prompts_lengths = [49, 14, 89, 9]
+    steps_add_reqs = [0, 0, 32, 131]
```
prashantgupta24, line 503: I wonder if we can lower the values here - the time for CB testing on CPU is on the rise, maybe (if possible) having shorter max_tokens can speed tests up if the test logic remains the same

prashantgupta24: I think the eventual goal could be to reduce the total number of steps - the lesser the steps, the faster the test. I don't think we really need `197` steps for this test case?

prashantgupta24: If this can be made to work with lesser `max_tokens`, then perhaps we can open an issue to change all tests within this file to use lesser values to speed things up?

PR #330, opened 2025-07-22
:test_tube: add long context test

joerunde:
# Description

This adds a failing test case for handling batch size 4 at 4k context length with continuous batching.

This tests the `eager` backend against the `sendnn` backend, which is currently failing on `quay.io/ibm-aiu/spyre-base:2025_07_18-amd64`


yannicks1 (approved):
lgtm

PR #331, opened 2025-07-24
[CB] Override number of Spyre blocks: replace env var with top level argument

yannicks1:
### [CB] Override number of Spyre blocks: replace env var with top level argument

depreciating the env var `VLLM_SPYRE_N_BLOCKS` and use the top level argument `num_gpu_blocks_override` instead to override the number of avail spyre kv cache blocks.


maxdebayser (approved):
LGTM

PR #337, opened 2025-07-25
[CB][Tests] Check output of scheduling tests on Spyre

sducouedic:
This code adds some logic to check the ouput of the scheduling steps tests. The output checking is done only on Spyre to save some compute for the cpu tests.

[Important note]
The output comparison checking of scheduling tests don't pass on CPU neither do they on Spyre. Exact reasons why are not sure, on cpu it might be because of high entropy of the randomly generated prompt tokens.


tests/e2e/test_spyre_cb_scheduler_steps.py
```diff
@@ -174,14 +177,29 @@ def test_prompts_aligned_with_tkv_boundaries(model: str, backend: str,
...
+            prompts=prompts,
+            max_new_tokens=seqs_max_tokens,
+            ignore_eos=True,
+        )
+        compare_results(model=model,
+                        tensor_parallel_size=1,
+                        backend=backend,
+                        vllm_results=cb_outputs,
+                        hf_results=hf_outputs)
+
```
prashantgupta24, line 195: wondering if we can somehow not duplicate this whole block multiple times in the test 🤔

rafvasq: Yeah.. I had the same thought. A version of this block shows up in a handful of other tests too:

https://github.com/vllm-project/vllm-spyre/blob/main/tests/e2e/test_spyre_basic.py#L89-L98

prashantgupta24 (approved):
All tests passed on eager + spyre!

tests/spyre_util.py
```diff
@@ -580,3 +638,46 @@ def get_chicken_soup_prompts(num_prompts: int) -> list[str]:
         prompts = prompts * (math.ceil(num_prompts / 4))
 
     return prompts[:num_prompts]
+
+
+def get_longer_chicken_soup_prompts(num_prompts: int) -> list[str]:
```
prashantgupta24, line 643: this is needed unfortunately since some of the prompt requests are pretty long in the scheduler tests

rafvasq (approved):
Thanks for the updates @prashantgupta24!

tests/spyre_util.py
rafvasq: I think there could be a refactor here with the overlap between the two prompt functions 🤔

tests/spyre_util.py
```diff
@@ -528,22 +566,42 @@ def create_text_prompt(model: str, min_tokens: int, max_tokens: int) -> str:
     return prompt
 
 
-def create_random_request(
-        request_id: int, num_tokens: int,
-        sampling_params: SamplingParams) -> EngineCoreRequest:
+def create_random_request(request_id: int,
+                          num_tokens: int,
+                          sampling_params: SamplingParams,
+                          from_model_vocab: bool = False,
```
joerunde, line 572: It looks like this is never set to true now, so this method no longer creates any random requests

tests/spyre_util.py
```diff
@@ -528,22 +566,42 @@ def create_text_prompt(model: str, min_tokens: int, max_tokens: int) -> str:
...
+        "model's vocabulary: need to provide model."
+
+        valid_token_ids = sorted([
+            v for v in tokenizer.vocab.values()
+            if v not in tokenizer.all_special_ids
+        ])
+        prompt_token_ids = random.choices(valid_token_ids, k=num_tokens)
+    else:
+        # start with existing prompts and tokenize them
+        prompts = get_longer_chicken_soup_prompts(4)
```
joerunde, line 587: 🤔  There's no guarantee that this will give you the number of tokens needed, and given current conversations with the compiler team we may need to add some unit test cases with longer prompts in the near future.

Maybe instead of making this more robust now, we could at least add an assert that the prompt has the correct number of tokens?

tests/spyre_util.py
```diff
@@ -580,3 +638,46 @@ def get_chicken_soup_prompts(num_prompts: int) -> list[str]:
...
+
+    prompts = [
+        template.format(
+            "Provide a list of instructions "
+            "for preparing chicken soup along with "
+            "rice curry to go with it so that"
+            "the flavor is amazing and make sure to follow the" \
+            "recipe that my mum used to make during my " \
+            "childhood so that I can relive my good" \
+            "memories thanks"
```
joerunde, line 657: I like how polite these are!

tests/spyre_util.py
prashantgupta24: Opened an issue for it - https://github.com/vllm-project/vllm-spyre/issues/348

tests/spyre_util.py
prashantgupta24: I wonder if we still need `set_random_seed` now though 🤔

tests/spyre_util.py
```diff
@@ -528,22 +566,42 @@ def create_text_prompt(model: str, min_tokens: int, max_tokens: int) -> str:
...
+        valid_token_ids = sorted([
+            v for v in tokenizer.vocab.values()
+            if v not in tokenizer.all_special_ids
+        ])
+        prompt_token_ids = random.choices(valid_token_ids, k=num_tokens)
+    else:
+        # start with existing prompts and tokenize them
+        prompts = get_longer_chicken_soup_prompts(4)
+        tokenized_prompts = tokenizer(prompts)["input_ids"]
+        prompt_token_ids = [p[:num_tokens] for p in tokenized_prompts][0]
```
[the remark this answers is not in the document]
joerunde: oh lol I didn't see the [0] :D

Issue #336, opened 2025-07-25
re-establish compatibility with vllm main


yannicks1:
### re-establish compatibility with vllm main

Some attributes have changed again, for decoder e.g. 
`AttributeError: 'VllmConfig' object has no attribute 'prompt_adapter_config'`

or embedding models: 
`AttributeError: type object 'Pooler' has no attribute 'from_config_with_defaults'`

maxdebayser:
The embeddings PR fixes the second one.

PR #338, opened 2025-07-25
♻️ Compatibility with vllm main

prashantgupta24:
# Changes

- Remove prompt adapter config
    - Based on upstream vllm changes in https://github.com/vllm-project/vllm/pull/20588
- Implement `get_supported_tasks` in model_runner for online API
    -  only needed for online API, related PR - https://github.com/vllm-project/vllm/pull/21585
    
    Implementation for online API:
    ```
    if envs.VLLM_USE_V1:
        supported_tasks = await engine_client \
            .get_supported_tasks()  # type: ignore
    else:
        supported_tasks = model_config.supported_tasks
    ```

## Related Issues

fix https://github.com/vllm-project/vllm-spyre/issues/336

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -35,6 +35,7 @@ else:
 from vllm.v1.outputs import EMPTY_MODEL_RUNNER_OUTPUT, ModelRunnerOutput
 
 #############################################################
+# from vllm.tasks import GenerationTask, PoolingTask, SupportedTask
 # TODO: remove when we have this in vllm/tasks.py
 #############################################################
 GenerationTask = Literal["generate", "transcription"]
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -3,13 +3,14 @@ import time
 from collections import deque
 from collections.abc import Iterable
 from dataclasses import asdict, dataclass
-from typing import TYPE_CHECKING, Optional, cast
+from typing import TYPE_CHECKING, Literal, Optional, cast, get_args
 
 import torch
 from torch import nn
 from vllm.config import DeviceConfig, VllmConfig
 from vllm.forward_context import set_forward_context
 from vllm.logger import init_logger
+from vllm.model_executor.models.interfaces_base import is_text_generation_model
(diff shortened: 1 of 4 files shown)


maxdebayser:
The pooling task changes are already part of the embedding PR. There are more upstream changes around the pooling tasks and I think they should be part of a follow-up PR. I don't want to pile up more changes in a PR that is already big enough.

prashantgupta24:
> 

Yeah I was trying to get a minimum set of changes in so that the tests pass against vllm:main. I think the pooling tasks are non-breaking, I need to get `get_supported_tasks` in for vllm:main to work atm

vllm_spyre/worker/spyre_embedding_model_runner.py
```diff
@@ -42,11 +42,20 @@ def __init__(
...
+                pooler_config,
+                pooling_type=PoolingType.CLS,
+                normalize=True,
+                softmax=False)
+        else:
+            self.pooler = Pooler.for_embed(
+                pooler_config=pooler_config,
+                default_pooling_type=PoolingType.CLS,
+                default_normalize=True,
+                default_softmax=False)
```
prashantgupta24, line 58: @maxdebayser I think just need this piece of code from your PR to enable vllm:main breaking changes, I've removed all pooling changes from this PR. I'm fine in waiting for your PR to get in first in which case I'll rebase and remove this change

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -375,6 +386,14 @@ def prepare_model_input(
         else:
             return self._prepare_decode(scheduler_output.scheduled_cached_reqs)
 
+    def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
+        tasks = list[SupportedTask]()
+
+        if "generate" in self.model_config.supported_tasks:
+            tasks.extend(["generate"])
```
prashantgupta24, line 393: Ideally we would want this to be coming from the model directly:
```
if is_text_generation_model(model):
            supported_tasks.append("generate")
```
but `SpyreCausalLM` doesn't seem to support it atm

```
type(self.model)
<class 'vllm_spyre.model_executor.model_loader.spyre.SpyreCausalLM'>

is_text_generation_model(self.model)
False

```

yannicks1 (approved):
lgtm! thanks for taking that on!

PR #339, opened 2025-07-25
Add CB API tests on the correct use of max_tokens 

gmarinho2:
# Description

Add tests to check if the API is generating the correct amount of tokens when CB is enabled.

### DIFF
--- a/tests/e2e/test_spyre_cb.py
+++ b/tests/e2e/test_spyre_cb.py
@@ -3,8 +3,9 @@
 Run `python -m pytest tests/e2e/test_spyre_cb.py`.
 """
 
+from openai import BadRequestError
 import pytest
-from spyre_util import (generate_spyre_vllm_output, get_chicken_soup_prompts,
+from spyre_util import (RemoteOpenAIServer, generate_spyre_vllm_output, get_chicken_soup_prompts,
                         get_spyre_model_list)
 from vllm import SamplingParams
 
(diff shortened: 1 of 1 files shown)


tests/e2e/test_spyre_cb.py
```diff
@@ -3,9 +3,10 @@
 Run `python -m pytest tests/e2e/test_spyre_cb.py`.
 """
 
+from openai import BadRequestError
```
ckadner, line 6: move the import `from openai import BadRequestError` after the line `import pytest`

to address the code lint error

tests/e2e/test_spyre_cb.py
```diff
@@ -95,4 +154,4 @@ def test_continuous_batching_with_long_contexts(model, monkeypatch):
     for i in range(len(vllm_cpu_results)):
         # As long as no sequences have top candidate tokens with very close
         # logprobs, the generated text should be identical.
-        assert vllm_cpu_results[i]["text"] == vllm_spyre_results[i]["text"]
+        assert vllm_cpu_results[i]["text"] == vllm_spyre_results[i]["text"]
```
ckadner, line 157: needs a line break at the end of file ⛔

ckadner (changes requested):
Just a quick review on code lint failures

tests/e2e/test_spyre_cb.py
```diff
@@ -44,6 +45,64 @@ def test_cb_max_tokens(
                                    monkeypatch=monkeypatch)
 
 
+@pytest.mark.cb
+@pytest.mark.parametrize("cb", [True])
```
maxdebayser, line 49: Can we also add a test for static batching? Because the logic for that is actually more complicated because the max context length is conditioned on the eligible warmup shapes.

maxdebayser (approved):

tests/e2e/test_spyre_cb.py
```diff
@@ -44,6 +45,64 @@ def test_cb_max_tokens(
...
+@pytest.mark.parametrize("model", get_spyre_model_list())
+@pytest.mark.parametrize("max_model_len", [256])
+@pytest.mark.parametrize("max_num_seqs", [2])
+@pytest.mark.parametrize(
+    "backend", [pytest.param("eager", marks=pytest.mark.cpu, id="eager")])
+def test__api_cb_rejects_oversized_request(
```
prashantgupta24, line 55: is that an extra `_` between `test` and `api`?

PR #341, opened 2025-07-28
♻️ fix vllm:main

prashantgupta24:
It seems `model_config.task` is deprecated, instead from what I understand we can use `model_config.supported_tasks`, which is initialized when an llm engine is instantiated:
https://github.com/vllm-project/vllm/pull/21470/files#diff-7eaad0b7dee0626bf29d10081b0f0c5e3ea15a4af97e7b182a4e0d35f8346953R705-R706

To maintain backward compatibility, it's a bit tricky since:
- Earlier version had `model_config.task` pointing to the task and `model_config.supported_tasks` as a list of all tasks which could contain more than 1 task
    ```
    model_config.task :  generate
    model_config.supported_tasks :  {'embed', 'reward', 'generate', 'classify'}
    ``` 
- Latest `main` now populates `model_config.supported_tasks` as the only task the model supports.
    ```
    model_config.task :  None
    model_config.supported_tasks :  ['generate']
    ```

### DIFF
--- a/vllm_spyre/v1/worker/spyre_worker.py
+++ b/vllm_spyre/v1/worker/spyre_worker.py
@@ -85,7 +85,8 @@ class SpyreWorker(WorkerBaseV1):
             (s["prompt_length"], s["new_tokens"], s["batch_size"])
                 for s in self.spyre_warmup_shapes
         ]):
-            if "embed" not in self.model_config.supported_tasks:
+            if (self.model_config.task and self.model_config.task != "embed"
+                    or "embed" not in self.model_config.supported_tasks):
                 # TODO: remove if spyre supports
                 # lower number of output tokens
                 assert num_decode_tokens >= 2, (
@@ -168,7 +169,8 @@ class SpyreWorker(WorkerBaseV1):
         self.model_runner: \
             Union[StaticBatchingSpyreModelRunner,
                   ContinuousBatchingSpyreModelRunner, SpyrePoolingModelRunner]
-        if "embed" in self.model_config.supported_tasks:
+        if (self.model_config.task and self.model_config.task == "embed"
+                or "embed" in self.model_config.supported_tasks):
             self.model_runner = SpyrePoolingModelRunner(
                 self.vllm_config, self.is_driver_worker)
             self.spyre_warmup_shapes = SpyrePlatform.get_warmup_shapes(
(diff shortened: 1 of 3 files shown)


vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -84,7 +84,10 @@ def compile_or_warm_up_model(self) -> None:
             (s["prompt_length"], s["new_tokens"], s["batch_size"])
                 for s in self.spyre_warmup_shapes
         ]):
-            if self.model_config.task != "embed":
+            # Can be simplified after the model_config change from vllm:main
+            if (self.model_config.task and self.model_config.task != "embed"
+                    or not self.model_config.task
+                    and "embed" not in self.model_config.supported_tasks):
```
yannicks1, line 90: this lines are duplicated many times. would it make sense to have a class attribute to the `SpyreWorker` class: `self.task` and then use this? or set `is_decoder` and `is_pooling` in the platform as attributes and read them?

prashantgupta24: Once main is updated this condition will become very simple, is it worth the effort for a temporary fix?

joerunde: If we simplify the code based on the upcoming changes though then we lose backwards compatibility, right?

For the cached requests overhaul we decided it was too complex to support both versions, so we updated the lower bound to 0.9.2. But this change looks small, if we can refactor this condition to be a little nicer then we can keep the support around.

I'd prefer refactoring this to one spot with a comment explaining the backwards compatibility for 0.10.0. Then @maxdebayser is thinking about ways to have automated checks for dead compatibility code, maybe we could wrap that in some context manager like:

```
with vllm_compatibility("0.10.0"):
   is_embed = # all this stuff
```

which will alert us with a CI failure when the lower bound is lifted to > 0.10.0

maxdebayser: I like the idea of the context manager! I wonder if we could disable mypy within context managers if needed.

yannicks1: > If we simplify the code based on the upcoming changes though then we lose backwards compatibility, right?

oh, my comment was not violate backwards compatibiliy, but rather to refactor to one spot. But I see you already started on that.

prashantgupta24: Yes, for now we will maintain backward compatibility. I have added the two conditions as attributes on the spyre worker.

yannicks1 (approved):
lgtm

maxdebayser (approved):
LGTM

PR #343, opened 2025-07-28
Upstream compatibility tests

maxdebayser:
Since we test against "main" and "default" vLLM versions in our tests, we often add code to make "main" work ahead of our official support. But it's easy to forget this compatibility code and it can become technical debt later. In this PR I'm adding a new test file to that verifies "main" and "default" imports from vLLM so that the "default" tests starts failing one a compatibility code is not longer needed.

### DIFF
--- a/tests/utils/test_upstream_compatibility.py
+++ b/tests/utils/test_upstream_compatibility.py
@@ -1,6 +1,7 @@
 import os
 
 import pytest
+from spyre_util import get_spyre_model_list
 
 pytestmark = pytest.mark.compat
 
@@ -29,11 +30,12 @@ def test_vllm_bert_support():
 
 
 @pytest.mark.cpu
-def test_model_config_task():
+@pytest.mark.parametrize("model", get_spyre_model_list())
+def test_model_config_task(model: str):
 
     from vllm.engine.arg_utils import EngineArgs
 
-    vllm_config = EngineArgs(model="facebook/opt-125m").create_engine_config()
+    vllm_config = EngineArgs(model=model).create_engine_config()
     model_config = vllm_config.model_config
 
     task = getattr(model_config, "task", None)


vllm_spyre/platform.py
```diff
@@ -71,8 +71,13 @@ class SpyrePlatform(Platform):
...
         # TODO: temporary hack while BertModels
         # inherit SupportsV0Only in vllm upstream.
+        import vllm.model_executor.models as me_models
         from vllm.config import ModelConfig
-        ModelConfig.is_v1_compatible = is_v1_compatible
+
+        # no need to patch after the model_config change
+        if 'model_config' not in \
+                inspect.getfullargspec(me_models.ModelRegistry.is_v1_compatible).args:
+            ModelConfig.is_v1_compatible = is_v1_compatible
```
prashantgupta24, line 80: Ah interesting I had this solved in another way - https://github.com/vllm-project/vllm-spyre/pull/341/files#diff-16ac04c4e75668ccde20cd2cfb82fa496d5d98dbe169ffede15652dc60a16066R54-R59

tests/utils/test_upstream_compatibility.py
```diff
@@ -0,0 +1,25 @@
...
+
+    bert_supports_v0_only = getattr(BertEmbeddingModel, "supports_v0_only",
+                                    False)
+
+    if VLLM_VERSION == "main":
+        assert not bert_supports_v0_only
+    else:
+        assert bert_supports_v0_only, (
+            "The currently supported vLLM version already"
+            "supports Bert in V1. Remove the compatibility workarounds.")
```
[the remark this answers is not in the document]
maxdebayser: Yes, I thought that the default was our lowest supported version. If it's not, then perhaps should add another configuration for our lower bound in the test scripts. Because otherwise we're not going to know whether we're actually supporting it.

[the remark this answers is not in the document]
joerunde: Yeah, would you mind seeing about adding in a `lowest` version to the matrix, but only running a small set of tests with it?

tests/e2e/test_spyre_embeddings.py
```diff
@@ -12,6 +12,8 @@
                         st_embeddings)
 from vllm import LLM
 
+pytestmark = pytest.mark.basic
```
prashantgupta24, line 15: I don't understand why we had to add this marker 🤔

prashantgupta24: ah okay

tests/utils/test_upstream_compatibility.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pytest
+
+pytestmark = pytest.mark.compat
+
+VLLM_VERSION = os.getenv("TEST_VLLM_VERSION", "default")
+
+
+@pytest.mark.cpu
+def test_vllm_bert_support(monkeypatch):
```
prashantgupta24, line 11: do we need to add `monkeypatch`?

maxdebayser: No, I forgot to remove it.

tests/e2e/test_spyre_online.py
```diff
@@ -2,6 +2,8 @@
 import pytest
 from spyre_util import get_spyre_backend_list, get_spyre_model_list
 
+pytestmark = pytest.mark.basic
```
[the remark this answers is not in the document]
maxdebayser: Actually it would be nice to run the 3 test in this file, but with just one combination of parameters. I wonder if there is a way to do that.

joerunde (approved):
LPGTM, would ideally like a smaller set of tests for backwards compatibility but we can get to that later

.github/workflows/test.yml
```diff
@@ -48,8 +48,21 @@ jobs:
             markers: "cpu and cb"
             flags: "--timeout=300"
           - name: "worker and utils"
-            markers: "not e2e"
+            markers: "cpu and not e2e"
```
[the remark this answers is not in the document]
maxdebayser: done

PR #277, opened 2025-07-02
V1 embeddings

maxdebayser:
# Description

This PR enables embedding models on vllm V1. In contrast with the V1 GPU implementation, here I added a separate model runner because for most of the embedding models there is no need for continuous batching. To avoid code repetition, I refactored the input batch and model runner classes into a class hierarchy with common base classes.

@gmarinho2 contributed a test that verifies that the returned embeddings don't change with batch size.

PR #344, opened 2025-07-29
[v1] remove v0 code

yannicks1:
### [v1] remove v0 code

Now as we have v1 support for embedding models (#277 ), we can finally delete the v0 code. 
Note: for decoder models v0 support was depreciated some time ago.


tests/e2e/test_spyre_embeddings.py
```diff
@@ -34,7 +32,7 @@ def test_output(
     '''
 
     monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
-    monkeypatch.setenv("VLLM_USE_V1", "1" if vllm_version == "V1" else "0")
+    monkeypatch.setenv("VLLM_USE_V1", "1")
```
joerunde, line 35: Maybe we can replace all of these with one `os.environ["VLLM_USE_V1"] = 1` in conftest.py?

maxdebayser: But this env var is not needed anymore right. Should be not add an assert in the llm engine to verify that it is an instance of the the V1 Engine class?

maxdebayser:
Nice!

yannicks1:
I did a very radical change and removed all lines where we set `VLLM_USE_V1=1`. We don't support any v0 code anymore in `vllm-spyre`. `vllm` enables v1 by default [here](https://github.com/vllm-project/vllm/blob/7b49cb1c6b1bf154c134a962888f2a6e12f9fc18/vllm/envs.py#L652), we do a check in platform.py [here](https://github.com/vllm-project/vllm-spyre/blob/5b7db096bd63046981cc00ca6e232b21d9b2de0b/vllm_spyre/platform.py#L115-L116) and tell the user to set it if it is unset for any reason. IMO this is safe enough. What do you think @joerunde @maxdebayser ?

tests/e2e/test_spyre_embeddings.py
maxdebayser: I think at some point `not envs.VLLM_USE_V1` will be removed. Also, this flag doesn't make sure that the current vLLM instance is not running as V0 as vLLM currently can fallback to V0 is VLLM_USE_V1 is unset.

joerunde:
Yeah I think this is good enough then, this prevents anybody from running with `VLLM_USE_V1=0`

tests/e2e/test_spyre_embeddings.py
maxdebayser: Can we check this flag with `os.getenv` instead? In this way the plugin won't crash when `envs.VLLM_USE_V1` is removed upstream

yannicks1: We have `VLLM_USE_V1` in other places in the code. These will have to be removed anyway when the `envs.VLLM_USE_V1` is removed upstream...

waleedqk:
bot:test
MARKERS="spyre"

tests/e2e/test_spyre_embeddings.py
yannicks1: I saw that the only remaining occurrences were for a hack testing both v0 and v1 engines. As we don't use that anymore, I removed it and incorporated you check via `os.getenv`. see [this](https://github.com/vllm-project/vllm-spyre/pull/344/commits/a286247cb86be4e5953daae50e573438b273151a) commit.

waleedqk:
bot:test
MARKERS="spyre and not quantized and not multi and not cb"

joerunde (approved):
negative diff let's goooo

docs/user_guide/supported_features.md
```diff
@@ -10,7 +10,7 @@ This table summarize the status of features on Spyre. By default, those features
 | Prompt Adapter                |   ⛔   | Being deprecated in vLLM [vllm#13981](https://github.com/vllm-project/vllm/issues/13981) |
 | Speculative Decoding          |   🗓️   |      |
 | Guided Decoding               |   🗓️   |      |
-| Pooling                       |   ⚠️   | Works with V0. V1 still being developed in vLLM [vllm#18052](https://github.com/vllm-project/vllm/issues/18052) |
+| Pooling                       |   ✅   |      |
```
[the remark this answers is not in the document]
maxdebayser: Since we don't support all pooling applications, I think it's better to remove this and leave just Embedding below.

PR #349, opened 2025-07-30
♻️ enable offline mode in GHA tests

prashantgupta24:
# Description

We should be able to run the tests in offline mode to avoid multiple calls to HF unnecessarily.

This also caught a bug (I think?) in one of the tests where we were unnecessarily trying to download the `facebook/opt-125m` model.

### DIFF
--- a/tests/e2e/test_spyre_prompt_logprobs.py
+++ b/tests/e2e/test_spyre_prompt_logprobs.py
@@ -75,27 +75,29 @@ def test_prompt_logprobs_must_be_enabled(monkeypatch: pytest.MonkeyPatch):
 
 @pytest.mark.cpu
 @pytest.mark.decoder
+@pytest.mark.parametrize("model", get_spyre_model_list())
 def test_prompt_logprobs_not_supported_with_cb(
-        monkeypatch: pytest.MonkeyPatch):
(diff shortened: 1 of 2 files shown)


tests/e2e/test_spyre_prompt_logprobs.py
```diff
@@ -75,27 +75,29 @@ def test_prompt_logprobs_must_be_enabled(monkeypatch: pytest.MonkeyPatch):
...
-        monkeypatch: pytest.MonkeyPatch):
+        model: str, monkeypatch: pytest.MonkeyPatch):
     # Server shouldn't boot with both prompt logprobs and continuous batching
     # enabled
     monkeypatch.setenv("VLLM_SPYRE_ENABLE_PROMPT_LOGPROBS", "1")
     monkeypatch.setenv("VLLM_SPYRE_USE_CB", "1")
 
     with pytest.raises(ValueError, match="continuous batching"):
-        VllmConfig(model_config=ModelConfig(task="generate"))
+        VllmConfig(model_config=ModelConfig(model=model, task="generate"))
```
prashantgupta24, line 87: This seems like a bug? No point in downloading the default facebook model for this test right?

joerunde (approved):
lpgtm!

PR #350, opened 2025-07-30
[Tests][FP8]: Add fp8 test

rafvasq:
# Description

- Adds an online unit test for fp8
- Updates spyre_model_list to use `quantized` instead of `quantization`, I think the latter can get confused with dynamic quantizing and the former refers to pre-quantized model type


tests/spyre_util.py
```diff
@@ -527,11 +527,16 @@ def _get_or_default(env: str, default: str) -> str:
...
             "sentence-transformers/all-roberta-large-v1")
         marks = [pytest.mark.embedding]
-    elif quantization == "gptq":
+    elif quantized == "gptq":
         # TODO: need a HF hub reference here as a default
         user_test_model_list = _get_or_default("VLLM_SPYRE_TEST_MODEL_LIST",
                                                "granite-3.0-8b-instruct-gptq")
         marks = [pytest.mark.decoder, pytest.mark.quantized, pytest.mark.spyre]
+    elif quantized == "fp8":
+        # TODO: need a HF hub reference here as a default
```
prashantgupta24, line 536: Maybe I'm missing some context here but if I click on the `quantizations` link on the model page:

Image

I see 2 models listed: https://huggingface.co/models?other=base_model:quantized:ibm-ai-platform/micro-g3.3-8b-instruct-1b

joerunde: I see that M Radermacher is quantizing a bunch of models? https://huggingface.co/mradermacher/micro-g3.3-8b-instruct-1b-GGUF

Seems like they work for https://www.nethype.de/

🤷

.github/workflows/test.yml
```diff
@@ -50,6 +50,9 @@ jobs:
           - name: "worker and utils"
             markers: "not e2e"
             flags: "--timeout=300"
+          - name: "quantized"
+            markers: "quantized"
```
joerunde, line 54: This would need to be `cpu and quantized` if we wanted to run on GHA

But this test would also already run in the `static batching` job as well, since it matches `cpu and decoder and not cb`. I would slightly prefer leaving it in the static batching job instead of creating anew job, just because we have so many entries in the test matrix now, and each one takes a couple minutes of CI worker time to spin up, load cache, and prepare the test environment

joerunde:
Super 🌶️🌶️🌶️ that this test is passing here!

.github/workflows/test.yml
```diff
@@ -137,6 +140,7 @@ jobs:
           
           download_tinygranite() {
             python -c "from transformers import pipeline, AutoTokenizer; pipeline('text-generation', model='ibm-ai-platform/micro-g3.3-8b-instruct-1b'); tokenizer=AutoTokenizer.from_pretrained('ibm-ai-platform/micro-g3.3-8b-instruct-1b')"
+            python -c "from transformers import pipeline, AutoTokenizer; pipeline('text-generation', model='ibm-ai-platform/micro-g3.3-8b-instruct-1b'); tokenizer=AutoTokenizer.from_pretrained('ibm-ai-platform/micro-g3.3-8b-instruct-1b-FP8')"
```
maxdebayser, line 143: Why is the pipeline initialized with `'text-generation', model='ibm-ai-platform/micro-g3.3-8b-instruct-1b` and the tokenizer with `ibm-ai-platform/micro-g3.3-8b-instruct-1b-FP8`?

tests/spyre_util.py
```diff
@@ -528,11 +528,16 @@ def _get_or_default(env: str, default: str) -> str:
...
+    elif quantized == "gptq":
         # TODO: need a HF hub reference here as a default
         user_test_model_list = _get_or_default("VLLM_SPYRE_TEST_MODEL_LIST",
                                                "granite-3.0-8b-instruct-gptq")
         marks = [pytest.mark.decoder, pytest.mark.quantized, pytest.mark.spyre]
+    elif quantized == "fp8":
+        user_test_model_list = _get_or_default(
+            "VLLM_SPYRE_TEST_MODEL_LIST",
+            "ibm-ai-platform/micro-g3.3-8b-instruct-1b-FP8")
+        marks = [pytest.mark.decoder, pytest.mark.quantized]
```
prashantgupta24, line 540: do we need to add `spyre` here?

joerunde (approved):
🌶️

tests/spyre_util.py
joerunde: I don't think so? Maybe the gptq one had it because it didn't work on cpu?

prashantgupta24: Verified

```
❯ VLLM_SPYRE_TEST_BACKEND_LIST=sendnn pytest --collect-only -m "spyre and quantized" tests
<Dir vllm-spyre>
  <Dir tests>
    <Dir e2e>
      <Module test_spyre_online.py>
        <Function test_openai_serving_gptq[warmup_shape0-gptq-sendnn-ibm-ai-platform/micro-g3.3-8b-instruct-1b]>
        <Function test_openai_serving_fp8[warmup_shape0-sendnn-ibm-ai-platform/micro-g3.3-8b-instruct-1b]>
```

prashantgupta24:
Opened an issue to "just rip out the gptq test since it doesn't work and we don't currently have a requirement to support it"

https://github.com/vllm-project/vllm-spyre/issues/356

PR #355, opened 2025-07-31
[Draft] fix logprobs + prompt logprobs

joerunde:
# Description

It turns out when prompt_logprobs are enabled, the static batching model runner will crash when calculating logprobs. 

This draft has a failing test, should be a simple fix

### DIFF
--- a/tests/e2e/test_spyre_prompt_logprobs.py
+++ b/tests/e2e/test_spyre_prompt_logprobs.py
@@ -61,6 +61,31 @@ def test_prompt_logprobs(
     force_engine_shutdown(llm)
 
 
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+@pytest.mark.parametrize("model", get_spyre_model_list())
+def test_prompt_logprobs_work_with_logprobs(
+    backend: str,
+    model: str,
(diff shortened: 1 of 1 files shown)


tests/e2e/test_spyre_prompt_logprobs.py
```diff
@@ -61,6 +61,31 @@ def test_prompt_logprobs(
...
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+    monkeypatch.setenv("VLLM_SPYRE_ENABLE_PROMPT_LOGPROBS", "1")
+    llm = LLM(model, tokenizer=model)
+
+    responses: list[RequestOutput] = llm.generate(
+        prompts, sampling_params=SamplingParams(prompt_logprobs=5, logprobs=5))
+
+    for r in responses:
+        assert r.outputs[0].logprobs is not None
+    force_engine_shutdown(llm)
```
prashantgupta24, line 86: can we add some comments on when we have to use this function? Also I wonder if we can make it such that it's part of pytest teardown automatically in case someone forgets to add this which would result in the weird addr-in-use error

PR #359, opened 2025-08-04
♻️ use fp8 model for testing SB + CB

prashantgupta24:
# Description

The general idea is that the FP8 model can be used in all SB/CB scenarios.

This requires unreleased changes from fms and fms-mo in order to work:
- https://github.com/foundation-model-stack/foundation-model-stack/pull/457
- https://github.com/foundation-model-stack/fms-model-optimizer/pull/176

While the tests may pass on cpu, they will continue to fail on spyre

## Related Issues

Also addresses: https://github.com/vllm-project/vllm-spyre/issues/356


tests/e2e/test_spyre_online.py
```diff
@@ -35,6 +55,9 @@ def test_openai_serving(remote_openai_server, model, warmup_shape, backend,
     assert len(completion.choices) == 2
     assert len(completion.choices[0].text) > 0
 
+    # rest are SB tests
```
prashantgupta24, line 58: I think?

.github/workflows/test.yml
```diff
@@ -41,9 +41,6 @@ jobs:
           - name: "static batching"
             markers: "cpu and decoder and not cb"
             flags: "--timeout=300"
-          - name: "quantized"
-            markers: "cpu and quantized"
-            flags: "--timeout=300"
```
prashantgupta24, line 46: maybe we could leave the `quantized` flag to only run tests with the quantized model?

prashantgupta24: That could also potentially speed things up as we can run the test with quantized model in parallel

joerunde:
bot:test

PR #360, opened 2025-08-05
[docs][CB] remove warning that no output correctness is asserted for scheduler step tests

yannicks1:
### [docs][CB] remove warning that no output correctness is asserted for scheduler step tests

we now do assert output correctness for the scheduling test since #337 , hence this warning should be removed.


maxdebayser (approved):
LGTM

PR #361, opened 2025-08-06
[embedding] support newest vllm main branch 

yannicks1:
### [embedding] support newest vllm main branch 

parameters `default_normalize` and `default_softmax` are deprecated


maxdebayser:
Yeah, I think it's better to add a test specifically for this, since the default_normalize refactoring is a discrete change in vllm.

yannicks1:
@maxdebayser, I added the test in `tests/utils/test_upstream_compatibility.py`. If you could have a quick look and approve that would be great.

maxdebayser (approved):
LGTM

PR #362, opened 2025-08-06
[CB] hard code number of spyre blocks to 2080

yannicks1:
### [CB] hard code number of Spyre blocks to 2080

We have received a number from the compiler team for the [model](https://huggingface.co/ibm-granite/granite-3.3-8b-instruct) of interest with TP 4 and hard code it here. For all other models/ TP sizes the number of blocks defaults to `max_batch_size * max_model_len / block_size`.

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -848,6 +848,8 @@ class ContinuousBatchingSpyreModelRunner(SpyreModelRunner):
             # hard coded value for tensor parallel size 4 with the below model
             # https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
             NUM_BLOCKS_SPYRE = 2080
+            logger.info("Model granite-3.3-8b-instruct and tensor parallel " \
+            "size 4 detected. Using NUM_BLOCKS_SPYRE = %d", 2080)
         else:
             # default value for any other model/ tensor parallel size
             NUM_BLOCKS_SPYRE = max_batch_size * min_req_num_blocks
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -835,19 +835,19 @@ class ContinuousBatchingSpyreModelRunner(SpyreModelRunner):
         Will eventually contain a function in torch_sendnn which reads 
         the actual value provided by the compiler for backend sendnn"""
 
-        max_batch_size = self.vllm_config.scheduler_config.max_num_seqs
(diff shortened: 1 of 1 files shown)


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -835,19 +835,19 @@ def _get_num_blocks_available(self) -> int:
...
-        # tests/e2e/test_spyre_cb.py::test_scheduler_cb_steps_tkv
-        # [seqs_max_tokens4-prompts_lengths4-steps_add_reqs4-
-        # checked_steps4-256-False-2-eager-llama-194m]
+
+        # TODO: replace NUM_BLOCKS_SPYRE_GRANITE_3_3_8B_INSTRUCT by calling
+        # a function in torch_sendnn which returns the value set by the Spyre
+        # compiler for the model being run, in the meantime we hard code the
+        # value for https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
+
+        NUM_BLOCKS_SPYRE_GRANITE_3_3_8B_INSTRUCT = 2080  # HARD CODED VALUE
```
joerunde, line 847: We should probably check to see if this is the actual model loaded, right? And this also applies at TP=4?

yannicks1: I checked with them, agreed on default values and updated the lines:)

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -839,15 +839,25 @@ def _get_num_blocks_available(self) -> int:
...
-        # checked_steps4-256-False-2-eager-llama-194m]
+
+        # TODO: replace the hard coded NUM_BLOCKS_SPYRE by calling a function
+        # in torch_sendnn which returns the value set by the Spyre compiler.
+
+        if ('granite-3.3-8b-instruct' in self.model_config.model
+                and self.parallel_config.world_size == 4):
+            # hard coded value for tensor parallel size 4 with the below model
+            # https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
+            NUM_BLOCKS_SPYRE = 2080
```
joerunde, line 850: Can we stick a log in here about this override happening?

Otherwise lgtm!

yannicks1: @joerunde done!

aluu317: This check` if 'granite-3.3-8b-instruct' in self.model_config.model` seems very specific. Can you write a unit test for this, so when we decide to switch to granite 3.4 or something else, the test will yell at you to change this check?

PR #363, opened 2025-08-06
[CB] set and respect compiler constraint VLLM_DT_MAX_BATCH_TKV_LIMIT 

yannicks1:
[CB] set and respect compiler constraint VLLM_DT_MAX_BATCH_TKV_LIMIT

The Spyre compiler has a constraint on the product of the batch size and tkv. For our [reference model](https://huggingface.co/ibm-granite/granite-3.3-8b-instruct) and TP=4 it cannot exceed `VLLM_DT_MAX_BATCH_TKV_LIMIT=131072`. The default value for all other models/ TP sizes is `VLLM_DT_MAX_BATCH_TKV_LIMIT=max_model_len * max_batch_size
`

changes:
- set `VLLM_DT_MAX_BATCH_TKV_LIMIT` depending on model and TP size
- introduce additional scheduler condition respecting this new constraint
- add drawing explaining the new scheduler check [here](https://github.com/vllm-project/vllm-spyre/pull/363#issuecomment-3173605517)

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -210,6 +210,8 @@ class SpyrePlatform(Platform):
             # hard coded value for tensor parallel size 4 with the below model
             # https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
             os.environ["VLLM_DT_MAX_BATCH_TKV_LIMIT"] = str(131072)
+            logger.info("Model granite-3.3-8b-instruct and tensor parallel " \
+            "size 4 detected. Using VLLM_DT_MAX_BATCH_TKV_LIMIT = %d", 131072)
         else:
             # default value for any other model/ tensor parallel size
             default_max_batch_tkv_limit = \
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -209,9 +209,10 @@ class SpyrePlatform(Platform):
                 and parallel_config.world_size == 4):
             # hard coded value for tensor parallel size 4 with the below model
             # https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
-            os.environ["VLLM_DT_MAX_BATCH_TKV_LIMIT"] = str(131072)
+            os.environ["VLLM_DT_MAX_BATCH_TKV_LIMIT"] = str(128 * 1024)
             logger.info("Model granite-3.3-8b-instruct and tensor parallel " \
-            "size 4 detected. Using VLLM_DT_MAX_BATCH_TKV_LIMIT = %d", 131072)
+            "size 4 detected. Using VLLM_DT_MAX_BATCH_TKV_LIMIT = %d",
+            128 * 1024)
         else:
             # default value for any other model/ tensor parallel size
             default_max_batch_tkv_limit = \
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -203,6 +203,8 @@ class SpyrePlatform(Platform):
         # min decode batch size is 2 due to symbolic shape constraint in torch
         os.environ["VLLM_DT_MAX_BATCH_SIZE"] = str(
             max(vllm_config.scheduler_config.max_num_seqs, 2))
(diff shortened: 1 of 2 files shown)


vllm_spyre/platform.py
```diff
@@ -204,6 +204,25 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         os.environ["VLLM_DT_MAX_BATCH_SIZE"] = str(
             max(vllm_config.scheduler_config.max_num_seqs, 2))
 
+        # max product of batch size x tkv supported by the Spyre compiler
+        if ('granite-3.3-8b-instruct' in model_config.model
+                and parallel_config.world_size == 4):
+            # hard coded value for tensor parallel size 4 with the below model
+            # https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
+            os.environ["VLLM_DT_MAX_BATCH_TKV_LIMIT"] = str(131072)
```
wallashss, line 212: NIT: 
```suggestion
            os.environ["VLLM_DT_MAX_BATCH_TKV_LIMIT"] = str(128 * 1024)
```

I think this would improve the readability/search for 128K constraint

vllm_spyre/v1/core/scheduler.py
```diff
@@ -237,5 +238,37 @@ def can_schedule(self, request) -> bool:
...
+                # If the new request is shorter, it limits the batch volume
+                max_batch_tkv = max(max_batch_tkv, batch_size * new_req_tkv)
+                break
+            else:
+                # Otherwise, use the current (longer) request's volume
+                max_batch_tkv = max(max_batch_tkv, batch_size * decode_req_tkv)
+                # decrease batch_size by 1 as the current request finished
+                batch_size -= 1
+
+        cond6 = max_batch_tkv <= int(max_batch_tkv_limit)
```
joerunde, line 272: I'd extract this whole condition into its own method with a detailed docstring and link to a diagram

yannicks1:
Drawing explaining the function `check_batch_tkv_limit()`
[max_batch_tkv_limit_ysc.pdf](https://github.com/user-attachments/files/21710920/max_batch_tkv_limit_ysc.pdf)

vllm_spyre/v1/core/scheduler.py
```diff
@@ -237,5 +242,64 @@ def can_schedule(self, request) -> bool:
             (self.tkv - request.num_prompt_tokens) / self.block_size)
         num_blocks_required -= num_fully_padded_blocks
         cond5 = num_blocks_required <= self.n_free_blocks
+        # check that batch size x tkv is smaller than the max supported number
+        cond6 = self.check_batch_tkv_limit(request)
+
         return start_new_batch or (cond1 and cond2 and cond3 and cond4
-                                   and cond5)
+                                   and cond5 and cond6)
```
wallashss, line 249: Food for thought:

First of all, I think later we should rename these `cond*` for something more semantic, while it works, I think with short good names we will have a nice summary of what is needed to satisfy the condition as a whole.

Based on the above comment, the final code would be something better than the suggestion below. I see that `check_batch_tkv_limit` is a more complex function than the other condition, there are sort, and some iterations over the requests. So, if you move this code like the suggestion we would use the short-circuit and only runs this function if the other conditions hold, saving some computation.

```suggestion
                                   and cond5 and self.check_batch_tkv_limit())
```

Another follow up of this suggestion is make sure the function is a _pure function_, that is, without side effects that could mutate `self` for example, otherwise we would have weird behaviors if the logic we write would need always this function to be executed.

That is it, in a nutshell this is an exercise that we would do to think about code readability|simplicity vs performance.

yannicks1: great ideas @wallashss ! There is going to be a follow up PR with more runtime optimizations, such as caching the function calls, as to save even more computations if the decode batch did not change in the meantime. Will definitely make it a pure function then. I will also have more 'early exits' as you suggest it. 
We are trying to get a working version in asap for the compiler team to work with the env var. @joerunde

PR #365, opened 2025-08-08
[Tests] Add long context batch tests

rafvasq:
# Description

Adds six cases:

- 32 prompts with 512 tokens each
- 16 prompts with 1.5k tokens each
- 8 prompts with 3k tokens each
- 4 prompts with 5k tokens each
- 2 prompts with 9k tokens each
- 1 prompt with 17k tokens

### DIFF
--- a/tests/e2e/test_spyre_cb.py
+++ b/tests/e2e/test_spyre_cb.py
@@ -189,7 +189,7 @@ def test_long_context_batches(
     prompt_len: int,
     monkeypatch: pytest.MonkeyPatch,
 ):
-    """Test continuous batching with various batch sizes and prompt lengths."""
+    """Tests continuous batching with various batch sizes and prompt lengths."""
 
     max_model_len = 32768
     max_num_seqs = 32
--- a/tests/e2e/test_spyre_cb.py
+++ b/tests/e2e/test_spyre_cb.py
@@ -3,10 +3,13 @@
 Run `python -m pytest tests/e2e/test_spyre_cb.py`.
 """
 
+import math
+
 import pytest
 from openai import BadRequestError
 from spyre_util import (RemoteOpenAIServer, generate_spyre_vllm_output,
-                        get_chicken_soup_prompts, get_spyre_model_list)
+                        get_chicken_soup_prompts, get_spyre_backend_list,
+                        get_spyre_model_list)
 from vllm import SamplingParams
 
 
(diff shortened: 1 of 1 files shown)


tests/e2e/test_spyre_cb.py
```diff
@@ -155,3 +158,68 @@ def test_continuous_batching_with_long_contexts(model, monkeypatch):
...
+)
+@pytest.mark.parametrize(
+    "batch_size,prompt_len",
+    [
+        (32, 512),
+        (16, 1500),
+        (8, 3000),
+        (4, 5000),
+        (2, 9000),
+        (1, 17000),
```
joerunde, line 181: I really want this to work, but we're already up to about 17 minutes for the continuous batching job on github :(

We gotta do something about that soon, but maybe for now we could loop through these combos in the body of the test so that we only have to construct the LLM once?

tests/e2e/test_spyre_cb.py
```diff
@@ -155,3 +157,88 @@ def test_continuous_batching_with_long_contexts(model, monkeypatch):
...
+                tuple([
+                    req_output.outputs[0].logprobs[i][t].decoded_token
+                    for i, t in enumerate(token_ids)
+                ]),
+                "logprobs":
+                tuple([
+                    req_output.outputs[0].logprobs[i][t].logprob
+                    for i, t in enumerate(token_ids)
+                ]),
+            })
```
prashantgupta24, line 240: This seems duplicated in `generate_spyre_vllm_output`. Perhaps we can make it a separate function and edit `generate_spyre_vllm_output` to use that?

tests/e2e/test_spyre_cb.py
```diff
@@ -155,3 +157,88 @@ def test_continuous_batching_with_long_contexts(model, monkeypatch):
...
+
+        assert len(results) == batch_size
```
prashantgupta24, line 242: Still gaining some context but do we care about result quality? (i.e. should we compare against HF?)

tests/e2e/test_spyre_cb.py
```diff
@@ -155,3 +157,88 @@ def test_continuous_batching_with_long_contexts(model, monkeypatch):
...
+        max_tokens=max_tokens,
+        temperature=0,
+        ignore_eos=True,
+        logprobs=0,
+    )
+
+    for batch_size, prompt_len in cases:
+        prompt = create_text_prompt(model,
+                                    min_tokens=prompt_len,
+                                    max_tokens=prompt_len + 1)
```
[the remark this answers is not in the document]
prashantgupta24: @wallashss are you getting confused by the variable name? From what I understand, we need to create prompts that are of the `prompt_len` defined in the test cases, then send in a request where `max_tokens` are requested using `SamplingParams`
```
SamplingParams(
        max_tokens=max_tokens, ...)
```

tests/e2e/test_spyre_cb.py
joerunde: For this case we've often used the good old `seq` prompt. It's not perfect, but something like this would give you prompts where the next few tokens should have candidates with very high logprobs:

```
# generate "0 1 2 3 4 5 ..."
text_prompt = " ".join([i for i in range(prompt_len)])

# truncate it to correct length
tokens = tokenizer(text_prompt)[0:prompt_len]

# put it back as a string
text_prompt = tokenizer.decode(tokens)
# Remove trailing number that may have been split
text_prompt = text_prompt.rsplit(" ",1)[0]
```

That would be much better than the current `🌶️ 🌶️ 🌶️ 🌶️' prompt to compare logprobs against the transformers model

joerunde: (I agree we should probably test against HF)

tests/e2e/test_spyre_cb.py
```diff
@@ -155,3 +157,88 @@ def test_continuous_batching_with_long_contexts(model, monkeypatch):
         # As long as no sequences have top candidate tokens with very close
         # logprobs, the generated text should be identical.
         assert vllm_cpu_results[i]["text"] == vllm_spyre_results[i]["text"]
+
+
+@pytest.mark.cb
+@pytest.mark.parametrize("model", get_spyre_model_list())
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
```
yannicks1, line 164: do we want to run that for CPU, or would it just be efficient to run on Spyre? it will take a long time..

tests/e2e/test_spyre_cb.py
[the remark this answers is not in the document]
yannicks1: how do you not get [this](https://github.com/vllm-project/vllm-spyre/blob/5902f0cee0f1d6b3c36ca9ce4267fa2deb0d28b5/tests/spyre_util.py#L570) assertion error? 
```
assert min_tokens < len(tokenizer.encode(prompt)) < max_tokens
```
when setting 
```
min_tokens=prompt_len,
max_tokens=prompt_len + 1
```

I think I am missing something 😆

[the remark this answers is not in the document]
yannicks1: also +1 on changing the args of `create_text_prompt()` to someting like `min_token_length` and `max_token_length`

tests/e2e/test_spyre_cb.py
yannicks1: yes please reuse code whenever we can:)

tests/e2e/test_spyre_cb.py
yannicks1: although I love the iconic 🌶️ , a sequence might actually be a better fit here...

tests/e2e/test_spyre_cb.py
```diff
@@ -155,3 +158,80 @@ def test_continuous_batching_with_long_contexts(model, monkeypatch):
...
+    batch_token_pairs = [
+        (32, 512),
+        (16, 1500),
+        (8, 3000),
+        (4, 5000),
+        (2, 9000),
+        (1, 17000),
+    ]
+
+    skip_unsupported_tp_size(tp_size, backend)
```
prashantgupta24, line 200: probably a nit but we can move this up the test. We don't really have to do anything if we plan on skipping the test

tests/spyre_util.py
```diff
@@ -552,26 +557,45 @@ def _default_test_models(isEmbeddings=False):
...
+    # Generate "0 1 2 3 ..." with extra so we have enough tokens to trim
+    text_prompt = " ".join(str(i) for i in range(token_length * 2))
+
+    # Take exactly the needed token count
+    tokens = tokenizer.encode(text_prompt)[:token_length]
+
+    # Remove trailing number that may have been split
+    text_prompt = tokenizer.decode(tokens).rsplit(" ", 1)[0]
+
+    return text_prompt
```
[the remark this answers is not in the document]
yannicks1: (used that in the past for some experiments with different tokenizers and it worked:)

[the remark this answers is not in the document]
rafvasq: Makes sense! I updated the function and tested it out, got exact token lengths 👍

PR #368, opened 2025-08-11
add optional --max_tokens input

kkvtran:
# Description

Update the openai_spyre_inference.py to include an option to specify the value for --max_tokens option.

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -204,6 +204,26 @@ class SpyrePlatform(Platform):
         os.environ["VLLM_DT_MAX_BATCH_SIZE"] = str(
             max(vllm_config.scheduler_config.max_num_seqs, 2))
 
+        # max product of batch size x tkv supported by the Spyre compiler
+        if ('granite-3.3-8b-instruct' in model_config.model
+                and parallel_config.world_size == 4):
+            # hard coded value for tensor parallel size 4 with the below model
+            # https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
+            os.environ["VLLM_DT_MAX_BATCH_TKV_LIMIT"] = str(128 * 1024)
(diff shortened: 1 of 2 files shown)


examples/online_inference/openai_spyre_inference.py
```diff
@@ -15,6 +15,24 @@
...
+parser.add_argument(
+    "--max_tokens",
+    type=int,
+    help="Maximum tokens.",
+)
+
+args = parser.parse_args()
+if args.max_tokens is None:
+    print(f"max_tokens is set to 20.")
+    max_tokens = 20
```
prashantgupta24, line 33: You can also think about using 20 as the `default` value within the `parser.add_argument` function

kkvtran: done.  thanks.

maxdebayser:
@kkvtran , to fix linting and DCO you could squash the two commits, fix linting and sign-off like this

To squash, pick one commit and squash the other
```
git rebase -i HEAD~2
```
Run linter:
```
./format.sh
git add <any file that was changed by the linter>
```

Fix DCO:
```
git commit -s --amend
```

Push 
```
git push --force
```

yannicks1:
@kkvtran, we have merge conflicts with main. The file you touch in this PR has not changed on main, so there should not be any merge conflicts... Please do not manually add commits from main to your branch. instead sync your fork to keep up to date with main. 

A possible way to fix this would be to reset your branch to [be3a458](https://github.com/vllm-project/vllm-spyre/pull/368/commits/be3a458c7b42129f9d79b1978b369d3bab762538). Then you go to https://github.com/kkvtran/vllm-spyre/tree/update-example and click on the sync button (description [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork)). Then we should be ready to merge:)

Issue #364, opened 2025-08-07
Provide upper bound for vllm version


prarit:
vllm-spyre's pyproject.toml provides a minimum version for vllm

dependencies = [
    "fms-model-optimizer>=0.2.0",
    "ibm-fms==1.1.0",
    "vllm>=0.9.2",
]

but does not provide a maximum supported version of vllm.

It would be good if this value had an upper bound as well as a lower bound.

PR #369, opened 2025-08-11
📌 pin vllm upper bound

prashantgupta24:
# Description

Pin vllm upper bound

## Related Issues

fix https://github.com/vllm-project/vllm-spyre/issues/364


joerunde (approved):
lgtm!

PR #370, opened 2025-08-11
feat: tests to check swapping of decode program

wallashss:
# Description

This PR add tests that verify the swap of decoding programs from compiler depending on the batch shape.


tests/e2e/test_spyre_cb.py
```diff
@@ -155,3 +156,89 @@ def test_continuous_batching_with_long_contexts(model, monkeypatch):
...
+                                              tensor_parallel_size=tp_size,
+                                              backend=backend,
+                                              max_num_seqs=max_num_seqs,
+                                              monkeypatch=monkeypatch,
+                                              block_size=max_model_len,
+                                              max_model_len=max_model_len,
+                                              use_cb=True)
+
+    # If we passed here then, everything is alright
+    assert len(vllm_results) == max_num_seqs
```
prashantgupta24, line 244: do we care about checking against HF?

wallashss: We could, but it can long so much to generate response for max_new_tokens in the order of 16K and 32K. So we need to think better how to run this test in a reasonable time.

wallashss: Yes. This is the only solution that I can think about now.

wallashss:
WIP: creating cache for this.

yannicks1 (approved):
lgtm

PR #373, opened 2025-08-12
:zap: cache hf results in tests

joerunde:
# Description

In an effort to reduce test runtime, this uses `functools.lru_cache` to cache text generation results from hf transformers. This should shave about 15% off the cpu runtime based on some quick measurements on an M3.

NB: This probably will not work when running with `--forked`, but we don't fork the tests on github actions runs.


tests/spyre_util.py
```diff
@@ -247,14 +248,18 @@ def generate_spyre_vllm_output(
 
 
 # Hugging Face
+# This uses lru_cache to cache the generated text so that we don't have to
+# always load and run the transformers model, nor manage a set of files of
+# expected results.
+@functools.lru_cache
```
ckadner, line 254: Should we limit the cache size?

https://docs.python.org/3/library/functools.html#functools.lru_cache

```Python
@functools.lru_cache(maxsize=128)
```

Though, I don't suppose there a reason to be concerned about a growing cache?

joerunde: Yeah I wasn't too concerned because this should be caching relatively small objects

joerunde:
🤔  This also doesn't appear to be reducing the runtime on github actions at all, which is a bit odd. Something seems off

tests/spyre_util.py
```diff
@@ -402,10 +407,12 @@ def compare_results(
 
 def check_output_against_hf(model, backend, max_new_tokens, vllm_results,
                             prompts) -> None:
+    # To cache results from hf, we have to pass tuples instead of lists
```
prashantgupta24, line 410: I do see an error on GHA - `TypeError: unhashable type: 'list'`

prashantgupta24:
> 🤔 This also doesn't appear to be reducing the runtime on github actions at all, which is a bit odd. Something seems off

I was hoping the unhashable error was the reason the caching wasn't working

joerunde:
Ah, so the reduction in time here is actually much better when static batching and continuous batching tests are run together, since they use the same prompts and would share cache for the expected results. Separated, there are far fewer cache hits :(

Also the tests like the tkv scheduler tests that generate specific-length prompts on the fly don't benefit either. So the speedup here isn't great, but could work better if we can manage to do things like:
- Run sb and cb tests together
- Standardize the prompts we run more across tests
- Pre-fetch prompts to be used in the suite and run them through hf up-front

joerunde:
Okay- swapping this over to use a file-based cache which should avoid loading the model w/ transformers at all in these test runs

tests/hf_result_cache.py
```diff
@@ -0,0 +1,78 @@
...
+    def write_cache(self):
+        """
+        Write the current cache to 'hf_cache.json' if it has been modified.
+        """
+        if self.dirty:
+            with open(self.cached_results_file_path, 'w') as f:
+                json.dump(self.cached_results, f)
+            self.dirty = False
+
+    def get_cached_result(self, model: str, prompt: str,
```
maxdebayser, line 45: Shouldn't the type annotation for prompt be `Union[str, list[int]]`?

tests/hf_result_cache.py
```diff
@@ -0,0 +1,78 @@
...
+        Returns an empty dictionary if no cache entry is found.
+        """
+        if isinstance(prompt, list):
+            prompt = self._token_ids_to_string(prompt)
+        max_tokens = str(max_tokens)
+
+        return self.cached_results.get(model, {}).get(prompt,
+                                                      {}).get(max_tokens, {})
+
+    def add_to_cache(self, model: str, prompt: str, max_tokens: int,
```
maxdebayser, line 58: Same comment about the prompt type annotation

maxdebayser (approved):
LGTM. The failing main tests are due to a new upstream change fixed in https://github.com/vllm-project/vllm-spyre/pull/380

Issue #371, opened 2025-08-12
fix embedding model upstream


yannicks1:
### fix embedding model upstream

we have to again update our api: 
`TypeError: Pooler.for_embed() got an unexpected keyword argument 'default_pooling_type'`

maxdebayser:
Yes

PR #374, opened 2025-08-12
Add compat code for changing Pooler function signature

maxdebayser:
# Description

Adds compatibility code for the `default_pooling_type` argument that was removed

## Related Issues

https://github.com/vllm-project/vllm-spyre/issues/371

### DIFF
--- a/tests/utils/test_upstream_compatibility.py
+++ b/tests/utils/test_upstream_compatibility.py
@@ -94,15 +94,31 @@ def test_pooler_default_args():
     has_from_config = hasattr(Pooler, "from_config_with_defaults")
 
     if not has_from_config:
-        annotations = inspect.getfullargspec(Pooler.for_embed).annotations
+        args = inspect.getfullargspec(Pooler.for_embed).args
         if VLLM_VERSION == "vLLM:main":
-            assert 'default_normalize' not in annotations
-            assert 'default_softmax' not in annotations
+            assert 'default_normalize' not in args
+            assert 'default_softmax' not in args
         elif VLLM_VERSION == "vLLM:lowest":
-            assert 'default_normalize' in annotations
-            assert 'default_softmax' in annotations
+            assert 'default_normalize' in args
+            assert 'default_softmax' in args
             # The compat code introduced in the PR below can now be removed:
             # https://github.com/vllm-project/vllm-spyre/pull/361
+
+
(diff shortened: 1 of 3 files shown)


tests/utils/test_upstream_compatibility.py
```diff
@@ -94,17 +94,33 @@ def test_pooler_default_args():
     has_from_config = hasattr(Pooler, "from_config_with_defaults")
 
     if not has_from_config:
-        annotations = inspect.getfullargspec(Pooler.for_embed).annotations
+        args = inspect.getfullargspec(Pooler.for_embed).args
```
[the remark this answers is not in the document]
maxdebayser: Yes, with `args` it makes it more obvious to understand what is going on when you glance over the code. With `annotations` you have to go to the `inspect` docs first.

yannicks1 (changes requested):
I am a bit concerned with the change from `annotations` to `args` for code that is not part of that fix. They are not the same thing and cannot be used interchangeably. I remember that I had to use `annotations` and not `args` in one of my PRs. Unless you manually check that all the gating still works for the different vllm versions, I suggest to revert it.

yannicks1:
Yeah, we can use args for all new cases where it works, just not changing it for older ones if we don't manually check. Thanks

vllm_spyre/platform.py
```diff
@@ -108,7 +108,7 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         model_config = vllm_config.model_config
         cache_config = vllm_config.cache_config
 
-        if scheduler_config.is_multi_step:
```
yannicks1, line 111: as we don't check anymore if the scheduler is multi step does that mean they deprecated multi step scheduling upstream?

maxdebayser:
All tests are passing now.

yannicks1 (approved):
lgtm, maybe answer the question before merging...

vllm_spyre/platform.py
maxdebayser: Yes, it was removed in V1. Woosuk  removed it in https://github.com/vllm-project/vllm/pull/22138 and https://github.com/vllm-project/vllm/pull/22779

yannicks1: great, thanks

PR #375, opened 2025-08-12
🐛  fix a bug in tests, add DISABLE_ASSERTS

prashantgupta24:
# Description

- Added `DISABLE_ASSERTS ` to the scheduling tests - It's very useful if you want to debug without actually asserting the values
- Fixed a bug where `server_args.extend` should have be called in tests


prashantgupta24:
bot:test
TEST_FILE=tests/e2e/test_spyre_online.py MARKERS="spyre"

tests/conftest.py
```diff
@@ -152,12 +152,11 @@ def remote_openai_server(request):
             "VLLM_SPYRE_USE_CB": "1",
             "VLLM_SPYRE_DYNAMO_BACKEND": backend
         }
-        server_args = [
+        server_args.extend([
```
prashantgupta24, line 155: found a 🐛 !

yannicks1:
just seen this... the modification to the model runner can be removed (see #384). The two internal issues on the tracker were coupled I guess. does this solve the remaining part of (internal) issue 977?

tests/scheduling_utils.py
```diff
@@ -11,6 +11,8 @@
 
 from vllm_spyre.v1.core.scheduler import ContinuousBatchingSpyreScheduler
 
+DISABLE_ASSERTS = True  # used for debugging
```
prashantgupta24, line 14: remember to change it back to False!

tests/scheduling_utils.py
```diff
@@ -140,54 +146,61 @@ def check_scheduler_inference_steps(
...
-            for req_id in req_ids2blocks:
-                # current number of used blocks should be less than reserved
-                assert len(
-                    req_ids2blocks[req_id]) <= req_ids2reserved_blocks[req_id]
-                # update requested/reserved blocks to check in last step
-                # Note: overwrite and not max because of reduce_left_padding()
-                requested_blocks[req_id] = len(req_ids2blocks[req_id])
-                reserved_blocks[req_id] = req_ids2reserved_blocks[req_id]
+            # 'MultiprocExecutor' object has no attribute 'driver_worker'
+            if tensor_parallel_size == 1:
```
wallashss: This is nice, I already faced this limitation. What happens on 'else', does nothing? If so, I think a comment to make this explicit would be nice.

wallashss:
Thanks for rename the title, it is much better! 😄 
(I was about to suggest it)

tests/scheduling_utils.py
prashantgupta24: I removed TP from the tests for now. If needed we can add it back later

Issue #295, opened 2025-07-09
Add tests for logits processor correctness


maxdebayser:
PR https://github.com/vllm-project/vllm-spyre/pull/290 updated our spyre input batch to use the new LogitsProcessor programming model.

We need to make sure not only that all sampling parameters that we support work for batch size 1, but also that they continue working as requests enter and leave the batch. For example, if we have a batch with multiple requests and one of them has a min_p setting, if another request leaves the batch, changing the size of the current batch, the dense indices are adjusted correctly so that min_p is not accidentally applied to another request.

PR #379, opened 2025-08-13
Add Sampling Params tests

gmarinho2:
# Description

This PR introduces individual tests to make sure the sampling params are being utilized when the batch size is 1.
 
## Related Issues

https://github.com/vllm-project/vllm-spyre/issues/295


tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,272 @@
...
+@pytest.fixture
+@pytest.mark.parametrize("model", get_spyre_model_list())
+def spyre_model(model: str, max_model_len: int, max_num_seqs: int,
+                block_size: int, tensor_parallel_size: int) -> LLM:
+    return LLM(
+        model=model,
+        tokenizer=model,
+        max_model_len=max_model_len,
+        max_num_seqs=max_num_seqs,
+        block_size=block_size,
```
maxdebayser, line 15: You can remove the `tokenizer`, `block_size`, `max_num_seqs` and `tensor_parallel_size`. The defaults are fine.

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,323 @@
...
+
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+def test_spyre_batch1_stop_sequence(spyre_model: LLM, backend: str,
+                                    monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+    monkeypatch.setenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "1")
+
+    stop_str = "London"
+    prompt = f"The best way to travel from Paris to {stop_str} is by train."
+    params = SamplingParams(stop=[stop_str], max_tokens=50)
```
maxdebayser, line 50: We should also test without stop and make sure that the finish_reason is different.

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,323 @@
...
+    prompt = "Repeat over and over again: a new day. Repeat: a new day."
+
+    param1 = SamplingParams(presence_penalty=2.0, max_tokens=50)
+    output = spyre_model.generate(prompt, param1)[0]
+
+    param2 = SamplingParams(presence_penalty=0.0, max_tokens=50)
+    no_penalty = spyre_model.generate(prompt, param2)[0]
+    no_penalty_count = no_penalty.outputs[0].text.lower().count("a new day")
+
+    assert output.outputs[0].text.lower().count("a new day") < no_penalty_count
```
maxdebayser, line 73: nice!

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,323 @@
...
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+def test_spyre_batch1_min_tokens(spyre_model: LLM, backend: str,
+                                 monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+    monkeypatch.setenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "1")
+
+    prompt = "Hello."
+    params = SamplingParams(temperature=0, min_tokens=20, max_tokens=25)
+
+    output = spyre_model.generate(prompt, params)[0]
```
maxdebayser, line 181: ditto

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+def test_spyre_batch1_ignore_eos(spyre_model: LLM, backend: str,
+                                 monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+    monkeypatch.setenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "1")
+
+    prompt = "One plus one equals two."
+    params = SamplingParams(temperature=0, ignore_eos=True, max_tokens=100)
```
maxdebayser, line 193: ditto

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,323 @@
...
+
+    params_with_special = SamplingParams(skip_special_tokens=False,
+                                         max_tokens=5)
+    output_with_special = spyre_model.generate(prompt, params_with_special)[0]
+
+    params_without = SamplingParams(skip_special_tokens=True, max_tokens=5)
+    output_without = spyre_model.generate(prompt, params_without)[0]
+
+    assert tokenizer.eos not in output_without.outputs[0].text
+    assert tokenizer.eos in output_with_special.outputs[0].text
```
maxdebayser, line 218: nice!

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,323 @@
...
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+def test_spyre_batch1_bad_words(spyre_model: LLM, backend: str,
+                                monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+    monkeypatch.setenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "1")
+
+    tokenizer = spyre_model.get_tokenizer()
+    prompt = "The capital of France is"
+
+    params = SamplingParams(max_tokens=5,
```
maxdebayser, line 244: ditto

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,355 @@
...
+@pytest.mark.parametrize("model", get_spyre_model_list())
+def spyre_model(model: str) -> LLM:
+    return LLM(model=model)
+
+
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+def test_spyre_batch1_temperature(spyre_model: LLM, backend: str,
+                                  monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
+    monkeypatch.setenv("VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER", "1")
```
maxdebayser, line 16: If these patches are required for all functions, you can put them in a autouse fixture with module scope

gmarinho2:
bot:test
MARKERS="spyre and quantized"

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,238 @@
+import pytest
+from spyre_util import get_spyre_backend_list
+from vllm import LLM, SamplingParams
+
+pytestmark = pytest.mark.basic
+
+
+@pytest.fixture(scope="module")
+def spyre_model() -> LLM:
+    return LLM(model="ibm-granite/granite-3.1-2b-instruct")
```
wallashss, line 10: You should probably set some limits in this model like set max-seq-lens and max-model-len, maybe the defaults for these are a too high for this model and you'll use more resources than you need for these. This also helps to run these tests faster.

prashantgupta24, line 10: Do we have to test with the full model?

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+@pytest.fixture(scope="function", autouse=True)
+def setenv(monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", get_spyre_backend_list())
```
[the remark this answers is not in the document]
prashantgupta24: Yeah I think we should use how we have it in other tests

tests/e2e/test_sampling_params.py
maxdebayser: What do you mean by full model and what are possible alternatives?

prashantgupta24: We usually unit test with the micro granite model `ibm-ai-platform/micro-g3.3-8b-instruct-1b"`

maxdebayser: Oh, right. I think the micro model output quality is not high enough to respond in a predictable manner to the prompt.

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,237 @@
...
+
+@pytest.fixture(scope="module")
+def spyre_model() -> LLM:
+    return LLM(model="ibm-granite/granite-3.1-2b-instruct", max_model_len=128)
+
+
+@pytest.fixture(scope="function", autouse=True)
+@pytest.mark.parametrize("backend", get_spyre_backend_list())
+def setenv(monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", "eager")
```
prashantgupta24, line 16: Should this also follow the convention used by other tests? That gives us the dynamic marking with different backends.

tests/e2e/test_sampling_params.py
gmarinho2: Only because it was the smallest model that could satisfy the minimum output quality needed for the tests

prashantgupta24 (commented):
Also maybe I'm missing some context but which GH test suite actually runs these tests? The file seems to be marked with `basic`, but the only test suite that runs basic tests is the `backward compat` which I'm guessing we don't want to include these tests?

tests/e2e/test_sampling_params.py
gmarinho2: Oh, this was a mistake. `"eager"` was meant to be `backend`. Or do you mean to set the env inside each test ([like here](https://github.com/vllm-project/vllm-spyre/blob/22a135b15fedec33ddee9d289c6385435123b5a9/tests/e2e/test_spyre_embeddings.py#L39))?

prashantgupta24: So if you see here - https://github.com/vllm-project/vllm-spyre/blob/main/tests/conftest.py#L67-L72
```
        _add_param(
            "backend",
            get_spyre_backend_list(),
            metafunc,
            existing_markers,
        )
```
This piece of code automatically adds `backend` as a parameter to any test function that just has `backend` in the argument list. So you can safely remove the line
```
@pytest.mark.parametrize("backend", get_spyre_backend_list())
```
and just add `backend` to this function.

And yeah it should be `backend` and not `eager`.

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def test_spyre_batch1_temperature(spyre_model: LLM):
```
prashantgupta24, line 19: also worth checking out `use_llm_cache` - https://github.com/vllm-project/vllm-spyre/blob/main/tests/e2e/test_spyre_cb.py#L25

We definitely need some sort of caching here

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+@pytest.fixture(scope="module")
+def spyre_model() -> LLM:
+    return LLM(model="ibm-granite/granite-3.1-2b-instruct", max_model_len=128)
+
+
+def test_spyre_batch1_temperature(spyre_model: LLM, backend, monkeypatch,
+                                  use_llm_cache):
+    monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
```
prashantgupta24, line 14: Oh I actually liked the earlier way in which we had a fixture to set this `monkeypatch`. We can always use the `backend` feature in that fixture and avoid duplication of this line across all tests

tests/e2e/test_sampling_params.py
```diff
@@ -0,0 +1,275 @@
+import pytest
+from vllm import LLM, SamplingParams
+
+pytestmark = pytest.mark.basic
```
prashantgupta24, line 4: These tests seem to be marked as basic which is why they are running with the `backward compat` test suite. I'm pretty sure that's not what we want?

maxdebayser: Yeah. We need another marker for testing features that we support at the vllm level which have nothing to do with SB or CB. Do you have a suggestion for a short name? (perhaps "functional"?)

prashantgupta24: We have a marker called `full_model` that uses the full model for testing, maybe we can mark all of these as `full_model` too

gmarinho2: Does the `full_model ` marker works for static batching? Static batching seems to be using other model then `ibm-granite/granite-3.3-8b-instruct`:
```bash
FAILED tests/e2e/test_sampling_params.py::test_spyre_batch1_temperature[JackFram/llama-160m-eager] - AttributeError: 'str' object has no attribute 'generate'
```

prashantgupta24: `full_model` marker should not be running on GH actions

prashantgupta24: also I think your issue is with the fact that `model` isn't being populated. Maybe the dynamic injection doesn't work with the global pytest marker 🤷

maxdebayser:
Or we could already create a `nightly` marker.

gmarinho2:
bot:test

gmarinho2:
bot:test
MARKERS="spyre and other_e2e"

wallashss (approved):
LGTM!

PR #380, opened 2025-08-14
[Compat]: Fix renamed NewRequestData argument

maxdebayser:
The `NewRequestData` argument `mm_inputs` was renamed to `mm_kwargs`.


maxdebayser:
> why are the CB tests taking even longer now? :(

Is `cleanup_dist_env_and_memory` fast? Since we call it from an `autouse` fixture it's run for every test.

PR #383, opened 2025-08-15
Remove block_size from arguments on LLM constructor.

yannicks1:
### Remove block_size from arguments on LLM constructor.

people might be confused by the top level argument `block_size`. We set it somewhat inconsistently throughout our examples/tests. It has no influence at all as it gets overridden in any case SB/CB by [this](https://github.com/vllm-project/vllm-spyre/blob/e6c0d335cad32b65c5f93119dd35ccd25f490fb6/vllm_spyre/platform.py#L188) line. 
For CB this value is the top level arg max_model_len, for SB it depends on the warmup shapes. 

This PR removes it to have more clarity that it does not need to be passed (and has no impact if passed).

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -123,6 +123,11 @@ class BaseSpyreModelRunner(ABC, Generic[InputBatchT, RequestStateT,
             if self.model_config.get_sliding_window():
                 logger.warning("Sliding window is not supported on Spyre. "
                                "The model will run without sliding window.")
+            assert (
+                self.cache_config.block_size == self.model_config.max_model_len
+            ), ("cache_config.block_size must be set to model_config."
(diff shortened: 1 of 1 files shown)


tests/e2e/test_spyre_cb.py
```diff
@@ -59,7 +58,7 @@ def test_cb_max_tokens(
 @pytest.mark.parametrize("max_num_seqs", [2])
 @pytest.mark.parametrize(
     "backend", [pytest.param("eager", marks=pytest.mark.cpu, id="eager")])
-def test__api_cb_rejects_oversized_request(
+def test_api_cb_rejects_oversized_request(
```
prashantgupta24, line 61: 🌶️

jberkhahn:
> LGTM! Is it worth adding a test where we send in a random block_size value and assert that it is indeed overwritten to `max_model_length`?

Probably worth adding a test that asserts on all the specific config functionality in there, no?

PR #385, opened 2025-08-15
[CB] add warning when exceeding 32K context length

yannicks1:
### [CB] add warning when exceeding 32K context length

Warn the user that we do only support 32K context length as of now. Requesting more might still compile, but is untested ground.

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -203,7 +203,7 @@ class SpyrePlatform(Platform):
         if (envs_spyre.VLLM_SPYRE_USE_CB
                 and vllm_config.model_config.max_model_len > 32 * 1024):
             logger.warning(
-                'Max context length too big. Currently only 32K (32768) ' \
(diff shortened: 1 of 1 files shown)


vllm_spyre/platform.py
```diff
@@ -200,6 +200,13 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         # set env vars for torch_sendnn to consume
         os.environ["VLLM_DT_MAX_CONTEXT_LEN"] = str(
             vllm_config.model_config.max_model_len)
+        if (envs_spyre.VLLM_SPYRE_USE_CB
+                and vllm_config.model_config.max_model_len > 32 * 1024):
+            logger.warning(
+                'Max context length too big. Currently only 32K (32768) ' \
```
wallashss, line 206: Nit:  missing a verb, also for me long would sounds better than big.
```suggestion
                'Max context length is too big. Currently only 32K (32768) ' \
```

PR #398, opened 2025-08-20
[GHA] Triggering test agains vLLM:main for ready labels added by bot

yannicks1:
### [tests] Triggering test agains vLLM:main for ready labels added by bot

previously PR #388 did only respect manually added `ready` labels. This PR extends the triggering the test against vLLM:main for `ready` labels added by the bot (by enabling auto-merge).


joerunde (approved):
Nice, I like the better formatting on the exclude condition too

PR #399, opened 2025-08-20
🔥 trim local envs not required anymore

prashantgupta24:
# Description

Trim local envs not required anymore

### DIFF
--- a/_local_envs_for_test.sh
+++ b/_local_envs_for_test.sh
@@ -4,13 +4,10 @@
 
 # Need to be set for tests to run
 export MASTER_ADDR=localhost
-export MASTER_PORT=29500
+export MASTER_PORT=12345
 
 # Run on CPU
 export VLLM_SPYRE_DYNAMO_BACKEND=eager
-
-# TODO: Tests don't work on CPU with MP enabled?
-export VLLM_ENABLE_V1_MULTIPROCESSING=0
 
 # Test related
 export VLLM_SPYRE_TEST_BACKEND_LIST=eager
--- a/_local_envs_for_test.sh
+++ b/_local_envs_for_test.sh
@@ -10,6 +10,10 @@ export MASTER_PORT=12345
 export VLLM_SPYRE_DYNAMO_BACKEND=eager
 
 # Test related
+
+# This makes debugging easy on local setup
+export VLLM_ENABLE_V1_MULTIPROCESSING=0
+
 export VLLM_SPYRE_TEST_BACKEND_LIST=eager
 # We have to use `HF_HUB_OFFLINE=1` otherwise vllm might try to download a
 # different version of the model using HF API which might not work locally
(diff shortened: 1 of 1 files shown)


_local_envs_for_test.sh
```diff
@@ -4,14 +4,11 @@
...
 # Need to be set for tests to run
 export MASTER_ADDR=localhost
-export MASTER_PORT=29500
+export MASTER_PORT=12345
 
 # Run on CPU
 export VLLM_SPYRE_DYNAMO_BACKEND=eager
 
-# TODO: Tests don't work on CPU with MP enabled?
-export VLLM_ENABLE_V1_MULTIPROCESSING=0
```
prashantgupta24, line 13: don't need this anymore

_local_envs_for_test.sh
```diff
@@ -4,14 +4,11 @@
 
 # Need to be set for tests to run
 export MASTER_ADDR=localhost
-export MASTER_PORT=29500
```
prashantgupta24, line 7: 12345 seems like a more reasonable default

_local_envs_for_test.sh
yannicks1: does that mean we can also remove `VLLM_ENABLE_V1_MULTIPROCESSING` from the [worker](https://github.com/vllm-project/vllm-spyre/blob/d0dd0e1ec4eda0bcc6bea4b4840513e30bb9cbda/vllm_spyre/v1/worker/spyre_worker.py#L675)? 

```
def maybe_override_signals_handler():
    if not (envs.VLLM_ENABLE_V1_MULTIPROCESSING
            and envs_spyre.VLLM_SPYRE_OVERRIDE_SIGNALS_HANDLER):
        return
```

_local_envs_for_test.sh
joerunde: ah but right now only `29500` works for me 😭

joerunde:
What about the model override? On my mac I don't set an override which makes it easier to run the embedding tests too, and I don't seem to run into the issue commented here about downloading a broken model version

_local_envs_for_test.sh
prashantgupta24: Not sure if that's related? In earlier versions I used to get non-deterministic output from the tests unless I explicitly disabled MP.

_local_envs_for_test.sh
maxdebayser: Why do we even need to set these? Can't we just run with defaults?

_local_envs_for_test.sh
prashantgupta24: To be clear it's not required to run the tests but does help in debugging locally

_local_envs_for_test.sh
joerunde: Yeah these are still required, but we could just set them in conftest.py

joerunde (approved):

PR #400, opened 2025-08-21
[GHA] 🐛 fix: Save HF models cache for all jobs

yannicks1:
### [GHA] 🐛 fix: Save HF models cache for all jobs

Since #393 , we only load a subset of the cached models instead of all. However, the condition in "Save HF models cache" assumed that all caches across the matrix are the same (as previously the case) and only saved the one for the first job. I removed that condition to save in all cases.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,7 +164,7 @@ jobs:
           key: ${{ runner.os }}-hf-model-${{ env.model_key }}
 
       - name: "Download HF models"
-        if: steps.changed-src-files.outputs.any_changed == 'true'
+        if: steps.changed-src-files.outputs.any_changed == 'true' && steps.cache_restore.outputs.cache-hit != 'true'
         run: |
           # We are caching HF models (HF_HUB_CACHE) for reliability rather than speed, since HF downloads are flaky for concurrent jobs.
           # Be careful when adding models to the cache here, as the GHA cache is limited to 10 GB.
@@ -209,7 +209,7 @@ jobs:
           esac
 
       - name: "Save HF models cache"
-        if: ( steps.changed-src-files.outputs.any_changed == 'true' && github.event_name != 'pull_request' )
+        if: ( steps.changed-src-files.outputs.any_changed == 'true' && github.event_name != 'pull_request' && steps.cache_restore.outputs.cache-hit != 'true' )
         uses: actions/cache/save@v4
         with:
           path: ${{ env.model_path }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,7 +164,7 @@ jobs:
           key: ${{ runner.os }}-hf-model-${{ env.model_key }}
 
       - name: "Download HF models"
-        if: steps.changed-src-files.outputs.any_changed == 'true' && steps.cache_restore.outputs.cache-hit != 'true'
+        if: steps.changed-src-files.outputs.any_changed == 'true'
         run: |
           # We are caching HF models (HF_HUB_CACHE) for reliability rather than speed, since HF downloads are flaky for concurrent jobs.
           # Be careful when adding models to the cache here, as the GHA cache is limited to 10 GB.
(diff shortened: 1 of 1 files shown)


yannicks1:
There is a small amount of redundant work:
for jobs: "continuous batching", "worker and utils", and "compatibility" the cache is the same. We currently save it for all of them. Could only do it for one (the first to be executed), but not sure if hard coding that is worth it. We could potentially run into a similar issue that we ran into here in the future when changing the test suite. Note that this only gets executed on main and if the source files have changed. 

That actually raises another question: Why is changing the source files affecting the cache? Aren't we just caching the model weights? Do we simply want to update the cache from time to time as the weights might have changed on hugging face? 
My naive condition for download and saving the cache would be: `if steps.cache_restore.outputs.cache-hit != 'true'` (note that I introduced that in #393 for download, because I noticed that we still spent some time there even though we had cache hits. But I believe it might interfere with the original logic by @ckadner...) Can you share your thought @ckadner ? cc @joerunde

.github/workflows/test.yml
```diff
@@ -164,7 +164,7 @@ jobs:
           key: ${{ runner.os }}-hf-model-${{ env.model_key }}
 
       - name: "Download HF models"
-        if: steps.changed-src-files.outputs.any_changed == 'true' && steps.cache_restore.outputs.cache-hit != 'true'
+        if: steps.changed-src-files.outputs.any_changed == 'true'
```
ckadner, line 167: This was a "proxy" condition to not run this step if no sources were changed, i.e. for documentation only updates.

The same awkward condition got sprinkled all over the test workflow to keep the test workflow required for all PRs (and run it all the time), yet not actually execute the logic inside the workflow. All other options seemed worse.

.github/workflows/test.yml
```diff
@@ -209,7 +209,7 @@ jobs:
           esac
 
       - name: "Save HF models cache"
-        if: ( steps.changed-src-files.outputs.any_changed == 'true' && github.event_name != 'pull_request' && strategy.job-index == 0 )
+        if: ( steps.changed-src-files.outputs.any_changed == 'true' && github.event_name != 'pull_request' )
```
ckadner, line 212: If we save the cache for all jobs, there can be concurrent cache write errors. I had added the `strategy.job-index == 0` condition to make sure it only got saved once per workflow run.

.github/workflows/test.yml
ckadner: If we want to keeping the same logic to run all steps yet not execute the logic inside the steps, we probably could word the condition a bit more succinctly.

```yaml
if: steps.changed-src-files.outputs.any_changed == 'true'
```

```yaml
if: env.SKIP_TESTS == 'true'
```

.github/workflows/test.yml
joerunde: @ckadner are those concurrent write errors only for concurrent writes with the same cache key? In this case we're trying to write multiple smaller cache entries

.github/workflows/test.yml
joerunde: Should we be keeping the `&& steps.cache_restore.outputs.cache-hit != 'true'` condition around though? We don't want to run the download step if the cache hit, right?

ckadner (changes requested):
We should keep the `cache-hit != 'true'` condition on step "Download HF models"

```yaml
  - name: "Download HF models"
    if: steps.changed-src-files.outputs.any_changed == 'true' && steps.cache_restore.outputs.cache-hit != 'true'
```

.github/workflows/test.yml
ckadner, line 212: We should add the condition `&& steps.cache_restore.outputs.cache-hit != 'true'` here as well

ckadner (changes requested):

yannicks1:
bot:test

joerunde (approved):
🌶️

ckadner (approved):
LGTM 👍🏻

PR #401, opened 2025-08-21
[GHA] 🎨 refactor test yaml

yannicks1:
### [GHA] 🎨 refactor test yaml

refactoring test.yaml. Removing duplicated model strings. Use of for loop for downloads.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -202,7 +202,8 @@ jobs:
                 download_roberta_large "$model" &
                 ;;
               *)
-                echo "No downloads needed for: $model"
+                echo "No download method found for: $model";
+                exit 1
                 ;;
             esac
           done
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,7 +164,7 @@ jobs:
           key: ${{ runner.os }}-hf-model-${{ env.model_key }}
 
       - name: "Download HF models"
-        if: steps.changed-src-files.outputs.any_changed == 'true' && steps.cache_restore.outputs.cache-hit != 'true'
+        # if: steps.changed-src-files.outputs.any_changed == 'true' && steps.cache_restore.outputs.cache-hit != 'true'
         run: |
           # We are caching HF models (HF_HUB_CACHE) for reliability rather than speed, since HF downloads are flaky for concurrent jobs.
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -172,41 +172,42 @@ jobs:
...
+                download_tinygranite "$model" &
+                ;;
+              "JackFram/llama-160m")
+                download_tinyllama "$model" &
+                ;;
+              "sentence-transformers/all-roberta-large-v1")
+                download_roberta_large "$model" &
+                ;;
+              *)
+                echo "No downloads needed for: $model"
```
ckadner, line 205: is this really true? should this message read something like "No download method for: $model" and exit with an error?

```suggestion
                echo "No download method found for: $model";
                exit 1
```

.github/workflows/test.yml
```diff
@@ -172,41 +172,42 @@ jobs:
...
+          hf_models="${{ matrix.test_suite.hf_models || env.DEFAULT_HF_MODEL }}"
+
+          if [[ "$hf_models" == "micro-g3.3_roberta-large" ]]; then
+            models=("ibm-ai-platform/micro-g3.3-8b-instruct-1b" "sentence-transformers/all-roberta-large-v1")
+          else
+            models=("$hf_models")
+          fi
+
+          for model in "${models[@]}"; do
+            case "$model" in
```
ckadner, line 194:
```suggestion
            echo "Downloading $model ..."
            case "$model" in
```

ckadner (changes requested):
Nice improvement. Though every time a PR comes up that modifies this step, I keep thinking we should pull this logic out into a little helper script though that may finding errors harder 🤔

To make debugging a little easier, I think we may want to add a little log message for each download.

.github/workflows/test.yml
yannicks1: good point, didnt touch that print message, but this sounds way more reasonable 😅

rafvasq (approved):
LGTM. Agree with @ckadner about a helper script but only if the model list continues to grow.

yannicks1:
bot:test

yannicks1:
@ckadner looks like you have to unblock by approving here. Also please just push to the branch and merge if there should be something else...

PR #402, opened 2025-08-21
feat: FP8 initial support on continuous batching

wallashss:
# Description

This PR add **initial** support for FP8 on Continuous Batching

# Changes

- Included FP8 logic in spyre.py which need to set the scale to the weights for CB
- [UDPATE]: Added padding for bs=1
- [EXTRA] Added decoding of generation on scheduler `test_spyre_cb_scheduler_steps.py` for better debugging later

# TODOS

- Set tolerance for logprobs difference for quantized model during tests
- Currently, the matrix of tests does not include `tests/e2e/test_spyre_cb_scheduler_steps.py` for FP8, we have to figure it out a clean way to include them. Moreover, most of these tests are failing and need better thinking before activating them.


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -324,7 +325,11 @@ def pad_input_ids(
         # this is a causal mask for generation
         mask = (mask.unsqueeze(-1) == mask.unsqueeze(-2)).tril()
         mask = torch.where(mask.logical_not(), -torch.inf, 0.0)
-        mask = mask.to(self.model.model.dtype)
+        # Currently, using this quantization will generate NaN, keep it float32
```
prashantgupta24, line 328: this is still suspicious behavior right? I'm investigating this issue separately

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -324,7 +325,11 @@ def pad_input_ids(
...
+        if self.model.model.dtype in [torch.float8_e4m3fn]:
+            mask = mask.to(torch.float32)
+        else:
+            mask = mask.to(self.model.model.dtype)
```
prashantgupta24, line 332: Also since we control the `self.model.model.dtype` (through the get_dtype` function), can we not move this logic there instead and make sure that `self.model.model.dtype` is always what we want it to be?

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -296,9 +302,6 @@ def __init__(
         rank: int,
     ) -> None:
 
-        if model_config.quantization:
-            raise ValueError("FP8 is not supported with continuous batching")
```
maxdebayser, line 300: Wait, should the PR name then be `[WIP] feat: FP8 support on static batching`?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1188,11 +1194,14 @@ def build_attn_metadata(
...
         # update only the SpyreAttentionMetadata
+
         return SpyreAttentionMetadata(
             slot_mapping=model_input.slot_mapping,
             current_tkv_mask=model_input.current_tkv_mask,
             left_padded_prompt_mask=model_input.left_padded_prompt_mask,
-            block_table=model_input.block_table)
+            block_table=model_input.block_table,
+            scale_indices=model_input.scale_indices,
+            is_prefill=model_input.is_prompt)
```
maxdebayser, line 1204: Are you setting the scale_indices here just because it has to be set? Because this pr is only for static batching, right?

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -430,25 +438,32 @@ def _set_past_key_value_states(self, num_blocks) -> None:
             # TODO: This does not work yet. The scale needs to be handled, see:
             # https://github.com/foundation-model-stack/aiu-fms-testing-utils/blob/v0.1.0rc3/aiu_fms_testing_utils/utils/paged.py#L306-L319
             from fms_mo.aiu_addons.fp8.fp8_utils import ScaledTensor
+            batch_size = max(2, self.scheduler_config.max_num_seqs)
```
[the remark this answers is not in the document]
wallashss: At least batch size. Currently, I found that had to set bs=2, but maybe it is not necessary for now. I'll revert it.

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: It's still CB+FP8

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: Which place? I didn't understand. 

Can I put there a TODO there and you fix that in your following PR?

joerunde: I thought that for running on spyre we wanted to set the mask as fp16 though, not necessarily fp32?

Maybe we should scrap `model.dtype` completely, and instead specify the dtypes that we need for specific tensors. e.g.

```
def self.get_mask_dtype(self):
  return fp16 if spyre else fp32

def self.get_kv_cache_dtype(self):
  return fp8 if self.is_fp8_model else fp16
```

tests/scheduling_utils.py
```diff
@@ -212,7 +221,12 @@ def check_scheduler_inference_steps(
                 new_token_ids[0])
             collected_outputs[output.request_id]["logprobs"].append(
                 new_logprobs[0][0])
+            collected_outputs[output.request_id]["tokens"].append(
```
maxdebayser, line 224: I'm not sure I understand why the decoding is needed. Is it just to print text for debugging instead of token indices?

vllm_spyre/model_executor/model_loader/spyre.py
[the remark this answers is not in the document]
maxdebayser: Then maybe we should update the scheduler_config in `Platform.check_and_update_config`.

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: Sorry, but I missed the last comment from Joe. 

I changed to something similar to his suggestion. Running the tests to check if everything's still alright. 

BTW, this change breaks the cache.

vllm_spyre/model_executor/model_loader/spyre.py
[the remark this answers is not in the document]
joerunde: Either way is probably fine, but does everything still work if the user sets `--max-num-seqs 1`? If not then I'd prefer overriding it in platform.py

wallashss:
bot:test

joerunde:
force merging- the internal tests passed but they just didn't report the status correctly

PR #404, opened 2025-08-21
🐛 set COMPILATION_MODE and FLEX_OVERWRITE_NMB_FRAME conditionally

prashantgupta24:
# Description

It looks like we have to set `COMPILATION_MODE` based on the model.

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -134,9 +134,13 @@ class SpyrePlatform(Platform):
 
         cls._check_threading_config(parallel_config.world_size)
 
-        # set COMPILATION_MODE based on the model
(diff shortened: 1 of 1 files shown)


vllm_spyre/platform.py
```diff
@@ -134,6 +134,10 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
 
         cls._check_threading_config(parallel_config.world_size)
 
+        # set COMPILATION_MODE based on the model
+        os.environ["COMPILATION_MODE"] = "offline_decoder" if \
+            is_decoder else "offline"
```
prashantgupta24, line 139: I guess I could check `is_pooling` explicitly too

joerunde (approved):
get in

PR #416, opened 2025-08-26
[test] 🎨 fix test description string

yannicks1:
### [test] 🎨 fix test description string

changes:
- number of available blocks was missing (note we don't state it if set to 1000 (no restriction)
- number of available blocks was stated wrongly (8 instead of 4)


yannicks1:
bot:test

PR #418, opened 2025-08-26
[CB] optimization: cache volumetric constraint in scheduler

yannicks1:
### [CB] optimization: cache volumetric constraint in scheduler

the result of the volume check (batch x tkv < limit) can be cached if the decode batch and the new request do not change between two calls (tkv can change). 

changes:
- implement cache for function `check_batch_tkv_limit()`. Two levels of cache: support multiple new requests per decode batch. Clear the cache when the decode batch changes (cache cannot grow bigger than the number of new requests (length of the scheduler waiting queue))
- only call the function if all other conditions are met (make it a lambda function)

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -156,7 +156,7 @@ class ContinuousBatchingSpyreScheduler(SpyreScheduler):
         assert self.max_batch_tkv_limit != '-1', (
             "Expecting the env var VLLM_DT_MAX_BATCH_TKV_LIMIT to be set in "
             "platform.py")
-        # cache for self.check_batch_tkv_limitouter: key: tuple(request_ids),
+        # cache for self.check_batch_tkv_limit() outer key: tuple(request_ids),
         # inner key: (request_id, max_batch_tkv_limit), value: (lower, upper)
         self._cache_check_batch_tkv_limit: dict[tuple, dict[tuple, tuple]] = {}
 


yannicks1:
bot:test

vllm_spyre/v1/core/scheduler.py
```diff
@@ -341,4 +362,31 @@ def check_batch_tkv_limit(self, request, tkv) -> bool:
...
+
+        if outer_key in cache:
+            # decode batch has not changed
+            if inner_key not in cache[outer_key]:
+                # adding new request to present decode batch
+                cache[outer_key][inner_key] = (-float("inf"), float("inf"))
+        else:
+            # decode batch has changed, empty the cache in place
+            cache.clear()
+            cache[outer_key] = {inner_key: (-float("inf"), float("inf"))}
```
wallashss, line 375: NIT: possible micro-optimization, `float("inf")` makes think it will parse, while we have already a constant for that.
```suggestion
            cache[outer_key] = {inner_key: (-math.inf, math.inf)}
```

vllm_spyre/v1/core/scheduler.py
```diff
@@ -341,4 +362,31 @@ def check_batch_tkv_limit(self, request, tkv) -> bool:
...
+        return_value = max_batch_tkv <= int(max_batch_tkv_limit)
+
+        if outer_key in cache:
+            # decode batch has not changed
+            if inner_key not in cache[outer_key]:
+                # adding new request to present decode batch
+                cache[outer_key][inner_key] = (-float("inf"), float("inf"))
+        else:
+            # decode batch has changed, empty the cache in place
+            cache.clear()
```
maxdebayser, line 374: Is my understanding correct that will only ever be one outer_key in the cache at the same time? If yes, is there perhaps a cheaper way to check if the batch hasn't changed? Since outer key is a tuple of a possibly large size, `outer_key in cache` will perhaps become very expensive as the number of running requests grows.

yannicks1: yes, valid point! I thought about building a key based on the request ids, but then thought it is an overkill and just took the tuple... But I can do that to ensure that this will never become a bottleneck in the future with bigger batches. good catch !

yannicks1: > Has the number of cache clearing events increased or diminished? It could be that some other attributes of the Request object are being updated if the clearing events have diminished.

It has diminished, and I think you are absolutely right! 
For `tests/e2e/test_spyre_cb_scheduler_steps.py` the cache clearances went from 21 to 14. 

The request object has at least one attribute (`self.num_computed_tokens`) which can change.

I will check tomorrow if the numbers add up (14/21) and use:
```
outer_key = tuple(r.request_id for r in running)
```
instead of 
```
outer_key = tuple(running)
```
in case it is explainable by the above.

vllm_spyre/v1/core/scheduler.py
```diff
@@ -247,13 +249,17 @@ def can_schedule(self, request) -> bool:
...
         # check that batch size x tkv is smaller than the max supported number
-        cond6 = self.check_batch_tkv_limit(request=request, tkv=self.tkv)
+        cond6 = lambda: self.check_batch_tkv_limit(request=request,
+                                                   tkv=self.tkv,
+                                                   running=self.running,
+                                                   max_batch_tkv_limit=self.
+                                                   max_batch_tkv_limit)
 
-        if cond1 and cond2 and cond3 and cond4 and cond5 and cond6:
+        if cond1 and cond2 and cond3 and cond4 and cond5 and cond6():
```
joerunde, line 258: if `cond6()` is the only thing that is false here, then we'll call it once and move onto the next block where we will immediately call `cond6()` again, right?

The lambda is super slick, this is better than not having it, but maybe if we re-organize the code to exit earlier and then only call `self.check_batch_tkv_limit` once we know we need to evaluate it, it'll be a bit more readable and we'll avoid the double-call edge case.

Since the caching makes the second call fast though, this suggestion doesn't have to block the PR

yannicks1: good point, I put it here because other wise I would need an additional line for cond6 not as a lambda, but as a boolean (and of course a different name). Alternative is as you say have `self.check_batch_tkv_limit(....)` directly, which makes the and concatenation less readable...

vllm_spyre/v1/core/scheduler.py
```diff
@@ -280,13 +286,17 @@ def can_schedule(self, request) -> bool:
...
-                                                   tkv=tkv_updated)
+        # with updated tkv (cond6) -> only call if the other cond are met
+        cond6_updated = lambda: self.check_batch_tkv_limit(
+            request=request,
+            tkv=tkv_updated,
+            running=self.running,
+            max_batch_tkv_limit=self.max_batch_tkv_limit)
 
-        return cond4_updated and cond5_updated and cond6_updated
+        return cond4_updated and cond5_updated and cond6_updated()
```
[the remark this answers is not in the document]
yannicks1: the lambda is for readability here... the 'and' concatenation is more readable this way. 

But I do actually agree that the function has become quite difficult to read with all these early exits and other gating. I can try to make this somewhat more linear/straight forward in a follow up PR (as independent of the caching)

vllm_spyre/v1/core/scheduler.py
yannicks1: Let me first reply to @maxdebayser 

I found the difference between the two approaches which yield 14/21 cache clearances! 
It is not caused by an attribute that changes for the request object. It is coming from the engine caching across the different tests. If not clearing the cache, the request ids start enumerating from 0 ... n again. I added a line [here](https://github.com/vllm-project/vllm-spyre/pull/418/files#diff-9396aeddee9a4c71244cf681f7029e5f36cdfaf8cf81188f259d6e8d04ec518b), and now the behavior is identical.

yannicks1:
bot:test

PR #422, opened 2025-08-27
:art: make hf_cache.json prettier

joerunde:
# Description

@sducouedic this PR updates the hf cache to add some indentation so that it is more human readable. It also checks that there are no cache misses during CI so that we don't accidentally delete cache and cause slow test runs.


sducouedic (approved):
LGTM

PR #428, opened 2025-08-28
♻️ [tests] Full model testing

prashantgupta24:
# Description

Introduces `pytest_generate_tests` - [pytest_generate_tests](https://docs.pytest.org/en/latest/how-to/parametrize.html#pytest-generate-tests) allows one to define custom parametrization schemes or extensions.

`pytest_generate_tests` is called when collecting a test function. Through the passed in `metafunc` object you can inspect the requesting test context and, most importantly, you can call `metafunc.parametrize()` to cause parametrization.

This was required for full model testing, in which case we want some of the parameters to change based on certain runtime configurations.

Running `pytest -m "full_model and cb"` will inject `full_model` CB arguments into the test marked with `full_model` .

_Note: We make sure to not parametrize something if it exists explicitly on the test._

⚠️ ⚠️ ⚠️ 

> SB with full model testing is broken since SB is picking up `--max-model-len` parameter instead of overriding it based on the warmup shapes. I have opened an issue on our internal board for this.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -128,6 +128,7 @@ def pytest_generate_tests(metafunc):
 
 def _add_param(param_name: str, param_value, metafunc,
                existing_markers) -> None:
+    """helper function to parametrize stuff"""
     if (param_name in metafunc.fixturenames
             and param_name not in existing_markers):
         metafunc.parametrize(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,6 +36,9 @@ def pytest_generate_tests(metafunc):
     ]
 
     marker = metafunc.config.option.markexpr  # From CLI
+    # TODO: make this condition better
+    # it can accidentally be triggered if we say
+    # -m "not full_model"
     if "full_model" in marker:
         # When -m full_model is called, all tests tagged with
         # full_model mark will be injected with these custom values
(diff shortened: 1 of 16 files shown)


prashantgupta24:
bot:test
MARKERS="not cb and full_model"

prashantgupta24:
bot:test
MARKERS="cb and full_model"

tests/conftest.py
```diff
@@ -17,6 +18,128 @@
...
+    default_max_num_seqs = [4]
+    default_max_model_len = [256]
+
+    existing_markers = [
+        marker.name if marker.name != "parametrize" else marker.args[0]
+        for marker in metafunc.definition.own_markers
+    ]
+
+    marker = metafunc.config.option.markexpr  # From CLI
+    if "full_model" in marker:
```
joerunde, line 39: Won't this also trigger in cases like `-m "not full_model"`? Is that okay because we never expect to negate the `full_model` marker?

tests/conftest.py
```diff
@@ -17,6 +18,128 @@
...
+                metafunc,
+                existing_markers,
+            )
+            _add_param(
+                "backend",
+                ["sendnn"],
+                metafunc,
+                existing_markers,
+            )
+            if "not cb" in marker:
```
joerunde, line 55: 🤔 

Trying to understand why this conditional is here- don't we always want to run full models with the larger warmup shapes and max model length?

prashantgupta24: I guess? I wasn't sure what the default for CB should be

tests/conftest.py
```diff
@@ -17,6 +18,128 @@
...
+
+    # apply to all
+    _add_param(
+        "max_num_seqs",
+        default_max_num_seqs,
+        metafunc,
+        existing_markers,
+    )
+
+    if "cb" in metafunc.fixturenames and "cb" not in existing_markers:
```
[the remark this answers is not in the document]
joerunde: Sure, I was thinking like accepting lists for values and marks and zipping them together:
```
_add_param(
    "cb", 
    values=[1, 0], 
    marks=[pytest.mark.cb, None]
)
```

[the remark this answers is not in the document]
prashantgupta24: nice! Let me add a TODO for that as well lol

prashantgupta24:
bot:test
MARKERS="cb and full_model"

PR #429, opened 2025-08-28
:hammer: Add debug log redirection option

joerunde:
# Description

This PR adds the `VLLM_SPYRE_WORKER_LOG_REDIRECT_DIR` environment variable config, which will redirect _all_ output from each worker process into a rank-specific logfile to debug tensor parallel crashes.


tdoublep (approved):
LGTM

PR #425, opened 2025-08-28
[cb][test] add tests for volumetric constraint with prefill optimization

yannicks1:
### [cb][test] add tests for volumetric constraint with prefill optimization

Adds tests for the volumetric constraint (PR #363) `VLLM_DT_MAX_BATCH_TKV_LIMIT` combined with the prefill optimization (PR #340) (currently behind a feature flag `VLLM_SPYRE_ENABLE_PREFILL_OPTIMIZATION`). 

Changes:
- `test_requests_use_full_batch_tkv_limit_prefill_opt`: just big enough volume -> can prefill sequence with optimization 
- `test_requests_exceed_batch_tkv_limit_prefill_opt`: exceeding volume -> cannot prefill sequence with optimization, have to decode current sequence first...

PR #432, opened 2025-09-02
🐛 solve undetected merge conflict with main

yannicks1:
### 🐛 solve undetected merge conflict with main

There was a undetected merge conflict of #425 and #428. This PR fixes the conflict in the added tests.


wallashss (approved):

yannicks1:
@joerunde should we maybe enforce syncing with main again before merging PRs to prevent bugs like this in the future? I believe the initial reason for dropping it was triggering GHA less frequently, or was there another motivation too?

PR #435, opened 2025-09-02
:alembic: draft supported model tests

joerunde:
# Description

Just an idea. This adds an easy way to specify (model, config) pairs to test, that we can use to maintain a documented set of supported runtime configurations.


joerunde:
Closing- we built better internal automation running on our spyre clusters to maintain coverage of supported models

PR #440, opened 2025-09-04
[cb] scheduler heuristic 2: unblock long prompts

yannicks1:
### [cb] scheduler heuristic 2: unblock long prompts

Introducing `VLLM_SPYRE_MAX_WAITING_TIME_PREFILL`, which is an upper bound on the waiting time [sec] of any request. After a request has waited for longer the current decode batch is locked and will finish decoding. The request will be either added to that locked batch or prefilled into a new exclusive locked batch.


vllm_spyre/v1/core/scheduler.py
```diff
@@ -179,14 +183,20 @@ def schedule(self) -> "SchedulerOutput":
         To avoid additional specialization, some requests are held back from the
         base scheduler but are restored after.
         """
+        # unlock the current decode batch if no requests are in running queue
+        if len(self.running) == 0:
+            self.batch_is_locked = False
+            logger.debug("Unlocking the current decode batch as no requests "
+                         "are in running queue")
```
tdoublep, line 190: This might log very frequently, even when the logic didn't kick it to lock the batch in the first place. Maybe we want to log conditionally on the batch being locked?

yannicks1: good idea!

vllm_spyre/v1/core/scheduler.py
```diff
@@ -182,14 +183,20 @@ def schedule(self) -> "SchedulerOutput":
...
             self.holdback_queue.append(self.waiting.popleft())
 
         # Check if new requests can be scheduled.
         while self.holdback_queue:
-            if self.can_schedule(self.holdback_queue[0]):
+            if not self.batch_is_locked and self.can_schedule(
+                    self.holdback_queue[0]):
                 # Add request to the waiting queue
                 self.waiting.append(self.holdback_queue.popleft())
             else:
```
sducouedic, line 205: shouldn't this be tested directly in the `can_schedule()` function? Maybe it can be the first tested condition, and return False directly if wrong

vllm_spyre/envs.py
```diff
@@ -113,6 +113,12 @@ def _backend_backwards_compat() -> str:
     "VLLM_SPYRE_N_TOKENS_PREFILL_PRIO":
     lambda: int(os.getenv("VLLM_SPYRE_N_TOKENS_PREFILL_PRIO", "-1")),
 
+    # scheduling heuristic: maximal waiting (blocking) time for prefill
+    # Prefills waiting longer than VLLM_SPYRE_MAX_WAITING_TIME_PREFILL
+    # seconds will have priority after the current decode batch has finished.
+    "VLLM_SPYRE_MAX_WAITING_TIME_PREFILL":
+    lambda: int(os.getenv("VLLM_SPYRE_MAX_WAITING_TIME_PREFILL", "-1")),
```
maxdebayser, line 120: Could this also be a float so that the user can specify 0.5 for 500ms?

vllm_spyre/v1/core/scheduler.py
```diff
@@ -182,14 +186,20 @@ def schedule(self) -> "SchedulerOutput":
...
         # First purge the full waiting queue into our holdback queue, preserving
         # priority
         while self.waiting:
             self.holdback_queue.append(self.waiting.popleft())
 
         # Check if new requests can be scheduled.
         while self.holdback_queue:
-            if self.can_schedule(self.holdback_queue[0]):
+            if not self.batch_is_locked and self.can_schedule(
+                    self.holdback_queue[0]):
```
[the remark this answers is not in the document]
sducouedic: you decide, my thought what that the decision of scheduling or not was entirely in one place. but I see your point also

sducouedic:
bot:test

tests/spyre_util.py
```diff
@@ -333,7 +333,7 @@ def create_random_request(
             mm_placeholders=None,
             sampling_params=sampling_params,
             eos_token_id=None,
-            arrival_time=0,
+            arrival_time=time.time(),
```
tdoublep, line 336: I would suggest using `time.monotonic()` instead to avoid issues with daylight savings etc.

vllm_spyre/envs.py
```diff
@@ -113,6 +113,12 @@ def _backend_backwards_compat() -> str:
     "VLLM_SPYRE_N_TOKENS_PREFILL_PRIO":
     lambda: int(os.getenv("VLLM_SPYRE_N_TOKENS_PREFILL_PRIO", "-1")),
 
+    # scheduling heuristic: maximal waiting (blocking) time for prefill
+    # Prefills waiting longer than VLLM_SPYRE_MAX_WAITING_TIME_PREFILL
+    # seconds will have priority after the current decode batch has finished.
+    "VLLM_SPYRE_MAX_WAITING_TIME_PREFILL":
```
tdoublep, line 119: The name should reflect the units of time that are being used (e.g., `VLLM_SPYRE_MAX_WAITING_TIME_SECONDS`) or something. Should we also consider using an integer instead of a float?

vllm_spyre/v1/core/scheduler.py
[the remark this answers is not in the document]
tdoublep: > Having it here is less code and avoids jumping into can_schedule() if we already know it is gonna return False

Couldn't it just be the first thing we check in `can_schedule` ?

tdoublep (commented):
Couple of minor comments but looks clean to me

yannicks1:
don't merge yet, I found something...

yannicks1:
bot:test

vllm_spyre/v1/core/scheduler.py
```diff
@@ -224,11 +233,29 @@ def can_schedule(self, request) -> bool:
         max_prompt_batch_size = 1
         max_context_len = self.scheduler_config.max_model_len
 
+        # if the batch is locked by a request which has been waiting for
+        # longer then VLLM_SPYRE_MAX_WAITING_TIME_SECONDS, we cannot
+        # schedule the current sequence until we have served this request
+        if self.batch_is_locked:
+            return False
```
joerunde, line 240: instead of locking the batch entirely, shouldn't we just disallow any skipping of requests in the queue until the request at the head of the waiting queue schedules?

I haven't followed super closely but my assumption is that the blocked request may be able to be scheduled before the full batch finishes. E.g. with the 128k limit, a 64k request could potentially schedule once the batch has drained down to a single other request, so we wouldn't need to wait for the last one to finish.

PR #441, opened 2025-09-04
✨ try absolute tolerance

prashantgupta24:
# Description

### DIFF
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -332,7 +332,11 @@ def compare_results(
 
                 if hf_token_id != vllm_token_id:  # different tokens
                     if backend == 'sendnn' and math.isclose(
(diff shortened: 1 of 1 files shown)


tests/spyre_util.py
```diff
@@ -309,7 +310,9 @@ def compare_results(
                             vllm_result['logprobs'])):
                 logprob_abs_diff = math.fabs(hf_logprob - vllm_logprob)
                 logprob_abs_diff_list.append(logprob_abs_diff)
-                logprob_rel_diff = math.fabs(logprob_abs_diff / hf_logprob)
+                logprob_rel_diff = math.fabs(
+                    logprob_abs_diff /
+                    max(math.fabs(hf_logprob), math.fabs(vllm_logprob)))
```
prashantgupta24, line 315: We should always be dividing with the max of the values?

PR #443, opened 2025-09-04
Bump vllm to v0.10.1 and add compatibility code

maxdebayser:


pyproject.toml
```diff
@@ -13,7 +13,7 @@ license = {text = "Apache 2"}
 dependencies = [
     "fms-model-optimizer[fp8]>=0.6.0",
     "ibm-fms>=1.2.1",
-    "vllm>=0.9.2,<=0.10.0",
+    "vllm>=0.9.2,<=0.10.1",
```
prashantgupta24, line 16: do we still need to support 0.9.2?

maxdebayser: Not sure... we only bumped the patch version and not the minor one...

maxdebayser:
bot:test

PR #445, opened 2025-09-05
Manage supported model configurations

ckadner:
# Description

- Add config file for models and runtime parameters
- Use configuration file in [documentation](https://vllm-spyre--445.org.readthedocs.build/en/445/user_guide/supported_models.html)
- Add validation code to compare requested model and runtime parameters with supported configurations
- Log warning when requested configuration is not supported
  <details><summary>Example:</summary>
  <code>
  WARNING 09-05 18:46:43 [runtime_config_validator.py:107] The requested configuration is not supported for model 'ibm-ai-platform/micro-g3.3-8b-instruct-1b': RuntimeConfiguration(platform=<PlatformName.AMD: 'x86_64'>, cb=True, tp_size=1, max_model_len=128, max_num_seqs=2, num_blocks=0, warmup_shapes=None)
  </code>
  </details>

**TODO:**
- [x] code cleanup
- [x] review/revise model-config YAML file structure
- [x] add a YAML field to ignore testing models/configurations for tiny model unit tests
- [x] what to use for `num_blocks` (cpu, gpu ...override)?
- [x] revise config validation logic and messaging
  - 2 stage config matching ... top level fields first, set containment for warmup_shapes second
- [x] update configs after release (candidate) testing
- [x] remove option to error out on unknown configuration
- [x] match models by config if they are mounted locally
- [ ] integrate model/runtime configurations into tests (#435)
- ❓ `get_warmup_shapes_from_envs()` does not yield same as `platform.py`:`cls._warmup_shapes`

**Review suggestions:**
- [@maxdebayser](https://github.com/vllm-project/vllm-spyre/pull/445#discussion_r2355712777)

> I wonder if it's feasible to test the warm up shapes like this. Maybe we could do something like:
> - [x] in the know configuration file, [only keep the] upper bound
> - [x] Validate that the prompts are multiples of 64
> - [x] Validate that prompt + new_tokens <= max_model_len
> - [x] Validate that the batch size is <= a tested upper bound.

## Related Issues

#435


vllm_spyre/config/supported_configurations.yaml
```diff
@@ -0,0 +1,85 @@
...
+# List of supported runtime configurations:
+#  - platform: [x86_64|s390x|ppc64le] amd64, zOS, PowerPC
+#  - cb: True: continuous batching mode; False: static batching mode
+#  - tp_size: [1|2|4|8|16] tensor parallel size
+#  - warmup_shapes := [(fixed_prompt_length, max_new_tokens, batch_size)] if cb: True
+#  - max_model_len: context length
+#  - max_num_seqs: number of sequences in a batch (per instance)
+#  - num_blocks: number of KV cache blocks
+runtime_configs:
+  - model: "ibm-ai-platform/micro-g3.3-8b-instruct-1b"
```
joerunde, line 12: ignore list for these, past the fold

vllm_spyre/config/supported_configurations.yaml
```diff
@@ -0,0 +1,85 @@
...
+      { platform: "x86_64", cb: True,  tp_size: 4, max_model_len: 256, max_num_seqs: 4, num_blocks: 0 },
+      { platform: "x86_64", cb: True,  tp_size: 8, max_model_len: 256, max_num_seqs: 4, num_blocks: 0 },
+    ]
+  - model: "ibm-granite/granite-3.3-8b-instruct"
+    configs: [
+      { platform: "x86_64",  cb: True,  tp_size: 4, max_model_len: 2048, max_num_seqs: 4 },
+      { platform: "x86_64",  cb: True,  tp_size: 4, max_model_len: 4096, max_num_seqs: 4 },
+      { platform: "x86_64",  cb: True,  tp_size: 4, max_model_len: 8192, max_num_seqs: 4 },
+      { platform: "s390x",   cb: True,  tp_size: 4, max_model_len: 8192, max_num_seqs: 4 },
+      { platform: "s390x",   cb: False, tp_size: 4, warmup_shapes: [[7168, 1024,  1]] },
```
joerunde, line 54: let's not publicly document Z and power support for now- there are no public builds for them.

For the simple case, let's start by running this validation while ignoring the platform flag

vllm_spyre/config/runtime_config_validator.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def report_error(msg: str, raise_error: bool = False):
+    if raise_error:
+        raise ValueError(msg)
+    else:
+        logger.warning(msg)
+
+
+def validate_runtime_configuration(model_config: ModelConfig,
```
joerunde, line 110: this only needs to run if the backed if `VLLM_SPYRE_DYNAMO_BACKEND=sendnn` ie we're running on spyre cards

tests/utils/test_model_config_validator.py
```diff
@@ -0,0 +1,147 @@
...
+    validate_runtime_configuration as validate)
+
+
+def setup_log_capture(caplog: LogCaptureFixture, level=logging.INFO):
+    """
+    Setup log capture for the test.
+    """
+    caplog.set_level(level)
+    if caplog.handler not in runtime_config_validator.logger.handlers:
+        runtime_config_validator.logger.addHandler(caplog.handler)
```
joerunde, line 18: oh this is super cool

vllm_spyre/config/runtime_config_validator.py
```diff
@@ -0,0 +1,162 @@
...
+    if len(matching_configs) > 0 and not use_cb:
+        supported_warmup_shapes = set([
+            ws for config in matching_configs
+            for ws in config.warmup_shapes or []
+        ])
+
+        requested_warmup_shapes = set(requested_config.warmup_shapes or [])
+
+        if not requested_warmup_shapes.issubset(supported_warmup_shapes):
+            report_error(
```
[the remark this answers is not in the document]
ckadner: Hi @maxdebayser I added validation code and unit tests for:
- [x] > test [requested warmup_shapes against] upper bound for prompt length, batch size and max_new tokens
- [x] > sum of prompt + max_new_tokens is smaller than the max_model_len [of supported configs]
- [x] > prompt size is a multiple of 64

Kindly take another look? Thank you! 🙏🏻

maxdebayser:
@ckadner , I was having trouble to explain my thoughts as review comments, so I put them in code form: https://github.com/ckadner/vllm-spyre/pull/19 .

maxdebayser:
@ckadner , my assumptions aren't correct. Please disregard some of my previous comments about upper bounds.

ckadner:
One last item to do: detect the model similar to what is done for granite, not rely on name which works for download from HF but not for file mounts

ckadner:
> [...] **question**: what is the motivation for maintaining the pre-downloaded `config.json` files? Is it to avoid downloads during testing or perhaps to make sure that the configurations are not updated remotely?

I wanted to have one unit test which verifies that we can consistently match models by their config. So, in that unit test, I need to instantiate a model config for each of our supported models and then use those to verify my runtime configuration validation code can reliably "guess" the correct model. Which is needed to actually verify the runtime configurations.

Since we run our unit tests in "offline" mode in our GitHub Actions tests, we need to have those configs available before test execution. 

We could add a pre-step in our GHA test workflow to download those configs (from HF or GHA cache) instead of keeping them in our unit tests folder. But @joerunde had already set that precedent of keeping 2 config.json files, and I like following precedent :-)

PR #446, opened 2025-09-05
:arrow_up: bump default vllm version to 0.10.1.1

joerunde:
# Description

Bumps the default vllm version to 0.10.1.1


prashantgupta24 (approved):
do we need to mention AFTU in the PR title?

joerunde:
Updated the description!

I think the ruff failures are because we dropped 3.9 and 3.10 support due to aftu as well :(

PR #448, opened 2025-09-05
:fire: Remove vllm 0.9.2 support

joerunde:
# Description

Removes support for vllm 0.9.2


prashantgupta24 (approved):
lgtm! The `vllm.tasks` test is already added

PR #450, opened 2025-09-05
chore: Use Python 3.11 by default for type-check

ckadner:
# Description

Currently, type-check with `mypy` defaults to Python 3.9 which generates incorrect type hint errors and we longer support Python 3.9 (or 3.10)

## Related Issues

N/A


yannicks1 (approved):
LGTM

yannicks1:
bot:test

PR #454, opened 2025-09-09
fix: block_size to be multiple of max_batch_size

wallashss:
# Description

fix block_size to be multiple of max_batch_size


vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -389,7 +389,9 @@ def get_num_blocks_available(self) -> int:
                 and self.parallel_config.world_size == 4):
             # hard coded value for tensor parallel size 4 with the below model
             # https://huggingface.co/ibm-granite/granite-3.3-8b-instruct
-            NUM_BLOCKS_SPYRE = 2080
+
+            # num_blocks_spyre must be multiple of max_batch_size
+            NUM_BLOCKS_SPYRE = max_batch_size * (2080 // max_batch_size)
```
prashantgupta24, line 394: do we really need this? Shouldn't the assert below be enough? If BS is 3, we get:

```
>>> bs = 3
>>> bs * (2080 // bs)
2079
```
Not sure what we get by setting this to 2079 then asserting False below?

wallashss: Why `False` ?

```
>>> 2079 / 3
693.0
>>> 2079 % 3 == 0
True
```

yannicks1:
bot:test

vllm_spyre/model_executor/model_loader/spyre.py
prashantgupta24: Oh my bad - I thought we were planning to never change the `NUM_BLOCKS_SPYRE` irrespective of BS. Hmm

prashantgupta24: I don't know if we have enough test coverage for this change though? Honestly we haven't tested enough with anything other than BS 4

prashantgupta24:
> Is that a new constraint from the compiler?

This is for the cases where `NUM_BLOCKS_SPYRE` have to be a multiple of BS which we found out here - https://github.com/vllm-project/vllm-spyre/pull/453.

Since this test was passing earlier, it does seem like something was added/changed in the compiler 🤷

yannicks1:
@prashantgupta24 @wallashss I did check with the compiler team and they confirmed that the current stack does require `NUM_BLOCKS_SPYRE` to be dividable by `max_num_seqs`. This restriction will go away with the next updates to the SW stack. But the decision is to have this in place for the time being (just in case). 

@joerunde should we merge this?

PR #458, opened 2025-09-12
Fix dimension of tensor passed to transformer classifier

maxdebayser:
Since we're using the classifier from the transformer module directly, we need to make sure that the tensor that is passed always has the batch dimension. When the batch size is larger than one, the hidden states are passed as a list to the pooler (up until v0.10.1.1), with each hidden state have the shape [prompt_len, hidden_dim].


maxdebayser:
bot:test

maxdebayser:
All tests are passing.

PR #464, opened 2025-09-16
[CB] optimization only return last block of prefill logits

yannicks1:
**---->>> TODO: REMOVE [THESE](https://github.com/vllm-project/vllm-spyre/pull/464#discussion_r2379297835) LINES BEFORE MERGING. <<<-----**
### [Do not merge][CB] optimization only return last block of prefill logits

please do not merge yet.

To reduce the overhead from data transfer for continuous batching, only the last blocks logits (instead of the entire prompts) is requested and returned from fms. Requires changes in foundation model stack. 

Even the static batching api had to be changed too, its logic is not affected by this. 

Note: Currently consuming this fms feature branch: https://github.com/foundation-model-stack/foundation-model-stack/tree/last_n_tokens

**---->>> TODO: REMOVE [THESE](https://github.com/vllm-project/vllm-spyre/pull/464#discussion_r2379297835) LINES BEFORE MERGING. <<<-----**


vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -529,7 +531,7 @@ def forward(
             mask=mask,
             past_key_value_states=self.past_key_value_states,
             use_cache=use_cache,
-            only_last_token=False,
+            last_n_tokens=SpyrePlatform.get_block_size() if is_prompt else 1,
```
yannicks1, line 534: we could potentially deduct from the shape of e.g. `input_ids` if it is a prompt or decode. I thought it wouldn't hurt to pass the boolean `is_prompt` that we have anyway to make things more readable. any thoughts?

.github/workflows/test.yml
```diff
@@ -244,6 +244,8 @@ jobs:
           # `uv run`, to avoid having `uv run` re-sync any dependencies or 
           # re-install the vllm_sypre package from source
           source .venv/bin/activate
+          # overwrite fms main with feature branch last_n_tokens
```
yannicks1, line 247: to be removed before merging!!!

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -692,26 +698,26 @@ def forward(
...
-            # In order to calculate prompt logprobs, we have to return the
-            # hidden states from the whole prompt. The static graphs need to be
-            # compiled with this set one way or the other.
-            only_last_token = False
-        else:
-            only_last_token = True
+        # In order to calculate prompt logprobs, we have to return the
+        # hidden states from the whole prompt. The static graphs need to be
+        # compiled with this set one way or the other.
+        last_n_tokens = 0 if envs_spyre.VLLM_SPYRE_ENABLE_PROMPT_LOGPROBS else 1
```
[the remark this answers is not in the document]
joerunde: prompt logprobs is busted anyway, I screwed up the implementation and it breaks decoding :( 

As long as the regular case works, I think we're good

.github/workflows/test.yml
joerunde: yes please!

We'll need to set the lower bound to the new fms release with this feature, and update the uv.lock file

JRosenkranz (approved):
lgtm

joerunde:
Upgraded to ibm-fms 1.4.0!

Will merge if tests pass here to get new RC out

joerunde (approved):
If @sducouedic and @JRosenkranz are good with it, I am too!

PR #465, opened 2025-09-16
🎨 scheduler: make holdback queue a local variable 

yannicks1:
### 🎨 scheduler: make holdback queue a local variable 

`holdback_queue` is always empty outside of schedule() -> does not need to live across different calls and therefore does not need to be a class variable.

Note: this change is primarily to make the code more readable and the logic easier to understand.


maxdebayser (approved):
Thanks for simplifying the code!

PR #466, opened 2025-09-17
[WIP][CB][FP8] fix batch size 1

yannicks1:
### [CB][FP8] fix batch size 1

We require an effective batch size of >= 2 for warmup and decodes when running fp8 continuous batching. This PR allows `--max_num_seqs 1` still to be set and served. 

As we pad single sequences, which may also occur when serving `--max_num_seqs>=2`, to batch size 2 anyway, the only change needed was to have the input_batch hold two sequences for warmup. 

We had the exact same issue for fp16 in the past, see #347 and #312

UPDATE: upstream logic has changed since, fix is more involved. De-prioritizing this and throw an error instead (#467)


yannicks1:
ahh, I see. I believe we did not have the logits_processor logic in upstream vLLM when we solved it for fp16... too bad. 
Agree with @joerunde, not worth the additional effort. Lets merge #467 instead and circle back when torch 2.8 is supported. Thanks a lot for trying this @wallashss !

PR #470, opened 2025-09-18
try static batching tests with the same model as the others

maxdebayser:


prashantgupta24 (commented):
You might also need to change this line:
```
VLLM_SPYRE_TEST_MODEL_LIST: "${{ matrix.test_suite.name == 'static batching' && 'JackFram/llama-160m' || '' }}"
```

prashantgupta24:
Wonder if we can also remove these lines as well - https://github.com/vllm-project/vllm-spyre/pull/470/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R183-R186

maxdebayser:
Since the tests don't pass with this model yet I've closed this PR.

PR #471, opened 2025-09-22
Disable compilation catalog

gkumbhat:
# Description

Extends https://github.com/vllm-project/vllm-spyre/pull/455 by adding support for a catalog.json file which is supposed to contain data from list of `model_compile.log.json` files. This is targeted to handle cases where we have multiple pre-compiled models available in 1 folder and this file will help in identifying which configs + models we support.


vllm_spyre/platform.py
```diff
@@ -520,3 +527,215 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+                f"DISABLE_COMPILATION=1 was set, but no pre-compiled model "
+                f"config was found in the TORCH_SENDNN_CACHE_DIR: "
+                f"{str(compilation_config_path)} or"
+                f"{str(compilation_catalog_path)} does not exist")
+
+        if not compilation_catalog_path.is_file() and \
+            not compilation_config_path.is_file():
+            raise ValueError(
+                "DISABLE_COMPILATION=1 was set, but the pre-compiled model "
+                "config is not a file")
```
tjohnson31415, line 586: These errors forces use of pre-compiled model configs, which I don't think we want (eg. so users can compile their own models by just running vllm).

vllm_spyre/platform.py
```diff
@@ -520,3 +527,215 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+        For decoder models we want to respect the `DISABLE_COMPILATION`
+        environment variable, which disallows torch_sendnn from compiling new
+        graphs and only allows it to load pre-compiled graphs.
+        In order to do this, we must load up some config from the torch_sendnn
+        cache and check to make sure that the current vllm config matches,
+        otherwise the cached artifacts cannot be used.
+
+        For encoder models, we do not allow disabling compilation.
+        """
+        disable_compilation_env_var = "DISABLE_COMPILATION"
```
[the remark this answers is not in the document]
gkumbhat: yep, I'll change this

vllm_spyre/platform.py
gkumbhat: My intention was, since they pointed to a pre-compiled model and we have something there, but its not a file (may be some copying issue), then this would allow them to catch

tjohnson31415: IMO, "since they pointed to a pre-compiled model" is not true. The configurations to the user are "disable compilation" and "use a cache", neither of which necessitate using a precompiled artifact config or catalog.

I suppose we could update the checks to say "if it exists, it must be a file" to catch the case that it ends up as a directory or something.

tjohnson31415: I think you mean the change of the env var to `REQUIRE_PRECOMPILED_DECODERS`?
But yeah, that solves my concern by making the request to use precompiled artifacts explicit

vllm_spyre/platform.py
```diff
@@ -515,3 +523,216 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+            return
+
+        # If this isn't a decoder model, re-enable compilation
+        if not is_decoder:
+            logger.info("Unsetting %s because %s is not a decoder model",
+                        DISABLE_COMPILATION_ENV_VAR,
+                        vllm_config.model_config.model)
+            os.environ[DISABLE_COMPILATION_ENV_VAR] = "false"
+            return
+
```
tjohnson31415, line 553: A little tweak to the logic here after changing to `REQUIRE_PRECOMPILED_DECODERS`

```suggestion
      if not is_decoder:
          return

      # REQUIRE_PRECOMPILED_DECODERS forces compilation to be disabled
      disable_compilation = os.getenv(DISABLE_COMPILATION_ENV_VAR, "true")
      if disable_compilation != "true"):
          raise ValueError("DISABLE_COMPILATION is set to allow compilation "
              "in conflict with REQUIRE_PRECOMPILED_DECODERS which disables it.",
          )

```

vllm_spyre/platform.py
```diff
@@ -515,3 +523,216 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+        # If the user asked to disable compilation, then we need to enforce that
+        # they setup their cache
```
tjohnson31415, line 555: With the `REQUIRE_PRECOMPILED_DECODERS` change, we could make checks on `DISABLE_COMPILATION=true` requiring cache enablement happen for both decoders and encoders (separate from the validation for `REQUIRE_PRECOMPILED_DECODERS`).

vllm_spyre/platform.py
```diff
@@ -515,3 +523,206 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+        For decoder models we want to respect the `REQUIRE_PRECOMPILED_DECODERS`
+        environment variable, which disallows torch_sendnn from compiling new
+        graphs and only allows it to load pre-compiled graphs.
+        In order to do this, we must load up some config from the torch_sendnn
+        cache and check to make sure that the current vllm config matches,
+        otherwise the cached artifacts cannot be used.
+
+        For encoder models, we do not allow disabling compilation.
+        """
+        req_precompiled_decoder_env_var = "REQUIRE_PRECOMPILED_DECODERS"
```
prashantgupta24, line 540: Would it make sense for this env var to live in `vllm_spyre/envs.py`?

vllm_spyre/platform.py
```diff
@@ -515,3 +523,206 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+            os.environ[DISABLE_COMPILATION_ENV_VAR] = "false"
+            return
+
+        # If the user asked to disable compilation, then we need to enforce that
+        # they setup their cache
+        torch_cache_dir = os.getenv("TORCH_SENDNN_CACHE_DIR", None)
+        torch_cache_enabled = bool(
+            int(os.getenv("TORCH_SENDNN_CACHE_ENABLE", "0")))
+
+        if not torch_cache_dir or not torch_cache_enabled:
```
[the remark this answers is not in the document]
prashantgupta24: (Or pathlib)
```
from pathlib import Path
Path(torch_cache_dir).is_dir()
```

tests/unit/test_disable_compilation.py
```diff
@@ -0,0 +1,140 @@
+import pytest
+from spyre_util import write_sample_model_config
+from vllm.config import (ModelConfig, ParallelConfig, SchedulerConfig,
+                         VllmConfig)
+
+from vllm_spyre.platform import (PRE_COMPILE_MODEL_CATALOG_FILENAME,
+                                 SpyrePlatform)
+
+
+@pytest.mark.parametrize("batch_type", ["sb", "cb"])
```
prashantgupta24, line 10: We usually have sb/cb parameterization like this in other tests:
```
parametrize(
            "cb", [pytest.param(1, marks=pytest.mark.cb, id="cb"), 0])
```
where cb ==1 means cb else sb.

Would it help if we follow the same convention? If so, the `tests/conftest.py` file automatically adds this parameterization if we just add an argument called `cb` within the test function name and remove the existing parameterization.

tests/unit/test_disable_compilation.py
```diff
@@ -0,0 +1,140 @@
...
+    monkeypatch.setenv("TORCH_SENDNN_CACHE_ENABLE", "1")
+
+    vllm_config = VllmConfig(
+        model_config=ModelConfig(
+            model="ibm-ai-platform/micro-g3.3-8b-instruct-1b",
+            max_model_len=256),
+        parallel_config=ParallelConfig(tensor_parallel_size=2),
+        scheduler_config=SchedulerConfig(max_num_seqs=2))
+
+    SpyrePlatform._handle_disable_compilation(vllm_config, is_decoder=True)
```
prashantgupta24, line 62: Going off of Travis' finding that we call `check_and_update_config` anyway in the post init for `VllmConfig`, can we not omit this explicit call?

vllm_spyre/platform.py
prashantgupta24, line 1: Would it make sense to add all compilation logic within its own package too instead of `platform.py`? We could reuse that name in the package for the unit test too

tests/unit/test_disable_compilation.py
tjohnson31415: I think we could, but then we should add a way to assert that the `_handle_disable_compilation` function is actually called in each test where we expect it to be.

I also looked in to creating an actual dummy VllmConfig that didn't do all of the `__post_init__` that we could pass to `_handle_disable_compilation`, but hit a Pydantic error if the ModelConfig wasn't post-init'd...

tests/unit/test_disable_compilation.py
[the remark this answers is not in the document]
gkumbhat: Yep, am open to suggestion on the location. Let me know.

tests/unit/test_disable_compilation.py
gkumbhat: surely. I'll try that out. I took this convention from the base PR of this change #455

gkumbhat:
Thanks @prashantgupta24 @tjohnson31415 for the review, we have updated with all the changes.

prashantgupta24:
> Thanks @prashantgupta24 @tjohnson31415 for the review, we have updated with all the changes.

Looking very nice! Do we plan on running this test with any of our GH test suites? If so then we need to add some markers to this test. Depends on whether we want to run only on CPU or spyre or both.

tests/unit/test_disable_compilation.py
prashantgupta24: Also I was able to get warnings asserted using custom caplog! That way we can test the actual warning message and omit the duplicate call to disable compilation

prashantgupta24: Also in my testing I had to monkeypatch `REQUIRE_PRECOMPILED_DECODERS` to True otherwise the code returns immediately after this line without actually running through the entire code:
```
if not envs_spyre.VLLM_SPYRE_REQUIRE_PRECOMPILED_DECODERS:
        return
```

gkumbhat:
@prashantgupta24 good catch, didn't know about that. I added the marker as well

tests/precompilation/test_disable_compilation.py
```diff
@@ -0,0 +1,156 @@
...
+        model_config=ModelConfig(
+            model="ibm-ai-platform/micro-g3.3-8b-instruct-1b",
+            max_model_len=256),
+        parallel_config=ParallelConfig(tensor_parallel_size=2),
+        scheduler_config=SchedulerConfig(max_num_seqs=2))
+
+    handle_disable_compilation(vllm_config, is_decoder=True)
+
+    assert "DISABLE_COMPILATION" in os.environ
+    assert os.getenv("DISABLE_COMPILATION") == "true"
```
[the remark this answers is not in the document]
prashantgupta24: Also maybe an overkill but one (or more) tests where we can check the edge cases like no matching config found or matching config found but vllm version doesn't match, etc.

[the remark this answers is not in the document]
gkumbhat: Done. Added a test for catalog mismatch as error message checking as well

vllm_spyre/compilation_utils.py
```diff
@@ -0,0 +1,225 @@
...
+    if not envs_spyre.VLLM_SPYRE_REQUIRE_PRECOMPILED_DECODERS:
+        return
+
+    if not is_decoder:
+        return
+
+    # If this isn't a decoder model, re-enable compilation
+    logger.info("Setting %s because %s is not a decoder model",
+                DISABLE_COMPILATION_ENV_VAR, vllm_config.model_config.model)
+    os.environ[DISABLE_COMPILATION_ENV_VAR] = "true"
```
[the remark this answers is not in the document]
gkumbhat: fixed the message

tests/precompilation/test_disable_compilation.py
```diff
@@ -0,0 +1,249 @@
...
+    monkeypatch.setenv("DISABLE_COMPILATION", "")
+
+    vllm_config = VllmConfig(
+        model_config=ModelConfig(
+            model="ibm-ai-platform/micro-g3.3-8b-instruct-1b",
+            max_model_len=256),
+        parallel_config=ParallelConfig(tensor_parallel_size=2),
+        scheduler_config=SchedulerConfig(max_num_seqs=2))
+
+    handle_disable_compilation(vllm_config, is_decoder=True)
```
prashantgupta24, line 68: if we're asseting warning messages, we don't really need to call this explicitly now, since `VllmConfig` post init will ensure we call this function. If that ever changes, the warning message assert will fail

tests/precompilation/test_disable_compilation.py
```diff
@@ -0,0 +1,249 @@
...
+    # monkeypath so that it resets the value to its previous state after the
+    # test as a cleanup
+    monkeypatch.setenv("DISABLE_COMPILATION", "")
+
+    vllm_config = VllmConfig(
+        model_config=ModelConfig(
+            model="ibm-ai-platform/micro-g3.3-8b-instruct-1b",
+            max_model_len=256),
+        parallel_config=ParallelConfig(tensor_parallel_size=2),
+        scheduler_config=SchedulerConfig(max_num_seqs=2))
```
prashantgupta24, line 66: small nit - we can use another context manager here - 

```
 with caplog_vllm_spyre.at_level(logging.WARNING):
```
this will ensure that the caplog.txt will have only warning messages.

tests/precompilation/test_disable_compilation.py
```diff
@@ -0,0 +1,249 @@
...
+
+    handle_disable_compilation(vllm_config, is_decoder=True)
+
+    assert "DISABLE_COMPILATION" in os.environ
+    assert os.getenv("DISABLE_COMPILATION") == "true"
+    assert "[PRECOMPILED_WARN]" in caplog_vllm_spyre.text
```
prashantgupta24, line 72: we should also check the warning text imo, since a lot of messages are pre-appended with [PRECOMPILED_WARN] in the compilation package?

tests/precompilation/test_disable_compilation.py
```diff
@@ -0,0 +1,246 @@
...
+    # test as a cleanup
+    monkeypatch.setenv("DISABLE_COMPILATION", "")
+
+    with caplog_vllm_spyre.at_level(logging.INFO):
+        _ = VllmConfig(model_config=ModelConfig(
+            model="ibm-ai-platform/micro-g3.3-8b-instruct-1b",
+            max_model_len=256),
+                       parallel_config=ParallelConfig(tensor_parallel_size=2),
+                       scheduler_config=SchedulerConfig(max_num_seqs=2))
+        assert "[PRECOMPILED_WARN]-Setting DISABLE_COMPILATION" \
```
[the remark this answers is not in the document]
gkumbhat: model name mismatch can actually happen and that is okay, so these are somewhat equivalent where Setting DISABLE_COMPILATION is more important for functionality perspective.

vllm_spyre/compilation_utils.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    For decoder models we want to respect the `REQUIRE_PRECOMPILED_DECODERS`
+    environment variable, which disallows torch_sendnn from compiling new
+    graphs and only allows it to load pre-compiled graphs.
+    In order to do this, we must load up some config from the torch_sendnn
+    cache and check to make sure that the current vllm config matches,
+    otherwise the cached artifacts cannot be used.
+
+    For encoder models, we do not allow disabling compilation.
+    """
```
tjohnson31415, line 35:
```suggestion
    """
    The `DISABLE_COMPILATION` environment variable disallows torch_sendnn from
    compiling new graphs forcing it to load from the cache. Enabling
    `VLLM_SPYRE_REQUIRE_PRECOMPILED_DECODERS` will force DISABLE_COMPILATION
    for decoder models and require pre-compiled models.

    In order to do this, we must load up some config from the torch_sendnn
    cache and check to make sure that the current vllm config matches,
    otherwise the cached artifacts cannot be used.
    """
```

vllm_spyre/compilation_utils.py
```diff
@@ -0,0 +1,228 @@
...
+
+    if not envs_spyre.VLLM_SPYRE_REQUIRE_PRECOMPILED_DECODERS:
+        return
+
+    if not is_decoder:
+        return
+
+    # If this is a decoder model, disable compilation
+    logger.info(
+        "[PRECOMPILED_WARN]-"
```
tjohnson31415, line 47: NITs:
`[PRECOMPILED_WARN]` is an awkward tag for an `info` level log... Since the tag is meant for log messages that have their own levels, using `[PRECOMPILED]` as the tag would be cleaner IMO.
Even NIT-ier: the `-` after the tag on the log messages seems strange to me; it can just be a whitespace.

vllm_spyre/compilation_utils.py
```diff
@@ -0,0 +1,228 @@
...
+        else:
+            matching_config = compilation_config
+
+    if matching_config:
+        # Check vllm_spyre version
+        try:
+            from vllm_spyre._version import version as vllm_spyre_version
+            if matching_config["vllm_spyre_version"] != vllm_spyre_version:
+                # Can be converted to ValueError if we want to be strict
+                # with checking
```
tjohnson31415, line 140: Now that the configuration is expcility `VLLM_SPYRE_REQUIRE_PRECOMPILED_DECODERS` we could be stricter on the checking here for cases where we know that no entry in the catalog can work (version mistmatches).

vllm_spyre/compilation_utils.py
```diff
@@ -0,0 +1,228 @@
...
+    if compilation_catalog_path.is_file():
+        with open(compilation_catalog_path) as f:
+            try:
+                pre_compile_catalog = json.load(f)
+            except json.JSONDecodeError as e:
+                raise ValueError(
+                    f"Precompiled catalog {str(compilation_catalog_path)}"
+                    " is not a valid JSON file") from e
+        match_result = match_from_pre_compile_catalog(pre_compile_catalog,
+                                                      vllm_config)
```
tjohnson31415, line 98: In a follow up PR, it would be better to surface all potential matches from the catalog. If there is no exact match on the model name, we could still check for "any" matches for certain configs (eg. vllm_spyre_version).

tjohnson31415 (commented):
Some NITs and comment update suggestions

vllm_spyre/compilation_utils.py
tjohnson31415: NVM, the `vllm_spyre_version` doesn't necessarily invalidate the cache...

PR #473, opened 2025-09-22
feat: enable custom logits processors

wallashss:
# Description

This PR enable to add custom logits processors.

### DIFF
--- a/tests/v1/worker/test_spyre_input_batch.py
+++ b/tests/v1/worker/test_spyre_input_batch.py
@@ -7,11 +7,11 @@ import pytest
 import torch
 from vllm.sampling_params import SamplingParams
 from vllm.utils import is_pin_memory_available, make_tensor_with_pad
-from vllm.v1.sample.logits_processor import BatchUpdateBuilder
+from vllm.v1.sample.logits_processor import LogitsProcessors
 from vllm.v1.sample.metadata import SamplingMetadata
 
-from vllm_spyre.v1.worker.spyre_input_batch import (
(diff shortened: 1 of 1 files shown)


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -305,9 +306,17 @@ def load_model(self, prompt_lens: Iterable[int],
...
-        # constructor argument; for now the list is fixed to builtin processors
-        logits_processors = get_builtin_logits_processors(self.vllm_config)
+
+        # TODO: compatibility code, remove it when it's not needed anymore
+        if hasattr(vllm.v1.sample.logits_processor, "LogitsProcessors"):
+            from vllm.v1.sample.logits_processor import build_logitsprocs
+
+            logits_processors = build_logitsprocs(
+                self.vllm_config, self.device, self.pin_memory, False,
+                self.vllm_config.model_config.logits_processors)
```
maxdebayser, line 316: Nice, but can you update `get_builtin_logits_processors` to use `build_logitsprocs` instead? Maybe we should also rename the method to  `get_logits_processors`.

maxdebayser (changes requested):

vllm_spyre/v1/worker/spyre_model_runner.py
wallashss: Thanks for the suggestion! Makes sense. Also, there was a PR that we removed some compatibility code, then we could simplify a lot this implementation.

wallashss:
bot:test

PR #477, opened 2025-09-24
[high prio] enable VLLM_SPYRE_ENABLE_PREFILL_OPTIMIZATION by default

yannicks1:
Our bench marking has shown that from both a user experience and a system throughput perspective this feature brings great benefits across all traces we evaluated on. 

this PR turns the feature on by default.


yannicks1:
Changing the default value for this changes the processing in /test_spyre_cb_scheduler_steps.py. I can adapt those later when I have some slack time. If we wanna get this in asap, feel free to force merge @tdoublep 
cc @joerunde .

tests/e2e/test_spyre_cb_scheduler_steps.py
```diff
@@ -686,8 +686,8 @@ def test_prompt_too_long_for_current_tkv(model: ModelInfo, backend: str,
             * 1: len = 70, max tokens = 4, step joining = 0
     """
 
-    if prefill_optimization:
-        monkeypatch.setenv('VLLM_SPYRE_ENABLE_PREFILL_OPTIMIZATION', '1')
+    if not prefill_optimization:
+        monkeypatch.setenv('VLLM_SPYRE_ENABLE_PREFILL_OPTIMIZATION', '0')
```
sducouedic, line 690: Maybe it would be better to not assume any default value and set it anyway. It makes things more explicit.

```suggestion
    monkeypatch.setenv('VLLM_SPYRE_ENABLE_PREFILL_OPTIMIZATION', '1' if prefill_optimization else '0')
```

tests/e2e/test_spyre_cb_scheduler_steps.py
```diff
@@ -913,8 +913,6 @@ def test_prefill_optimization_tkv_too_big(model: ModelInfo, backend: str,
             * 1: len = 70, max tokens = 50, step joining = 0
     """
 
-    monkeypatch.setenv('VLLM_SPYRE_ENABLE_PREFILL_OPTIMIZATION', '1')
-
```
sducouedic, line 917: Same, why not keeping it to make it explicit

tests/e2e/test_spyre_cb_scheduler_steps.py
```diff
@@ -1084,8 +1082,6 @@ def test_prefill_optimization_use_more_than_available_blocks(
         * available_blocks: 4
     """
 
-    monkeypatch.setenv('VLLM_SPYRE_ENABLE_PREFILL_OPTIMIZATION', '1')
```
sducouedic, line 1087: same

tests/e2e/test_spyre_cb_scheduler_steps.py
```diff
@@ -1942,14 +1940,15 @@ def test_requests_exceed_batch_tkv_limit(model: ModelInfo, backend: str,
 @pytest.mark.parametrize("max_num_seqs", [2])
 @pytest.mark.parametrize("max_model_len", [192])
 @pytest.mark.parametrize("available_blocks", [None])
-def test_requests_use_full_batch_tkv_limit_prefill_opt(
-        model: ModelInfo, backend: str, monkeypatch: pytest.MonkeyPatch,
-        set_random_seed, max_num_seqs: int, max_model_len: int,
-        available_blocks: int):
+def test_requests_use_full_batch_tkv_limit(model: ModelInfo, backend: str,
```
sducouedic, line 1943: Some other tests above are prefixed with "test_prefill_optimization_"

sducouedic (commented):
In my opinion it is better not to assume any default value for the prefill optimization. All the tests that have different behaviour with and without would be named "{test_name}_prefill_opt", "{test_name}_no_prefill_opt" and environment variable set accordingly

yannicks1:
Hi @sducouedic I addressed your feedback: 
All the tests that have different behaviour with and without the optimization are now named "{test_name}_prefill_opt" and "{test_name}_no_prefill_opt" and the environment variable set explicitly. The rest of the test are untouched. 

Note that tests starting with "test_prefill_optimization_" do explicitly test the prefill optimization (I also explicitly set the env there).

PR #478, opened 2025-09-24
feat: golden token injector logits processor

wallashss:
# Description

This PR introduces the golden token injector as logits processors as helper in our test to guide the generation based on expectation. It will choose the right token if during sampling it has probability below a given threshold.

At first, I just updated a set of test to illustrate the usage and validate the results. We'll need a follow up PR to update the remaining tests.

### DIFF
--- a/tests/golden_token_injector.py
+++ b/tests/golden_token_injector.py
@@ -10,24 +10,30 @@ from vllm.v1.sample.logits_processor import (BatchUpdate, LogitsProcessor,
 
 
 class ExpectationState:
-    '''This class controls the state of the generation.'''
+    '''
+    This class controls the state of the generation.
+    Args:
+        expected_token_ids: Expected tokens ids
+        expected_logprobs: Expected logprobs
+        error_threshold: Acceptable threshold to keep the injection. If it is 
+            over the threshold, we stop the injection and give feedback at the
+            end of the generation that this token is diverging too much.
+        label: Used to identify the request, ideally it would be the request 
+            id. However we might not have that yet, therefore we have the
+            opportunity to add a more human friendly label. It is used to log 
+            which requests are being injected with the golden token.
+    '''
 
-    def __init__(self, input_dict: dict):
+    def __init__(self,
+                 expected_token_ids: list[int],
+                 expected_logprobs: list[float],
+                 error_threshold: float,
+                 label: Optional[str] = None):
 
-        # Expected token id
-        self.token_ids: list[int] = input_dict["expected_token_ids"]
-        # Expected logprob
-        self.logprobs: list[float] = input_dict["expected_logprobs"]
-        # Acceptable threshold to keep the injection. If it is over the
-        # threshold, we stop the injection to give feedback at the
-        # end of the generation that this token is diverge too much.
-        self.threshold: float = input_dict["error_threshold"]
-        # Used to identify the request, ideally it would be the request id.
-        # However we might not have that yet, therefore we have the
-        # opportunity to add a more human friendly label.
-        # It is used to log which requests are being injected with
-        # the golden token.
-        self.label: Optional[str] = input_dict.get("label")
+        self.token_ids: list[int] = expected_token_ids
+        self.logprobs: list[float] = expected_logprobs
+        self.threshold: float = error_threshold
+        self.label: Optional[str] = label
 
         self.current_token_idx = 0
         self.has_error = False
@@ -61,7 +67,7 @@ class GoldenTokenInjector(LogitsProcessor):
             if params.extra_args and (
                     injector_dict :=
                     params.extra_args.get("golden_token_injector")):
-                self.req_states[index] = ExpectationState(injector_dict)
+                self.req_states[index] = ExpectationState(**injector_dict)
 
         if not self.req_states:
             return


tests/e2e/test_spyre_cb_scheduler_steps.py
prashantgupta24, line 2377: if we're comparing against HF within the `check_scheduler_inference_steps` function itself then we don't need these 2 return values

prashantgupta24, line 1: `cb_outputs, prompts = check_scheduler_inference_steps`

if we're comparing against HF within the check_scheduler_inference_steps function itself then we don't need these 2 return values

tests/golden_token_injector.py
```diff
@@ -0,0 +1,144 @@
...
+                    params.extra_args.get("golden_token_injector")):
+                self.injectors[index] = GoldenTokenInjectorState(injector_dict)
+
+        if self.injectors:
+            # Process removed requests.
+            for index in batch_update.removed:
+                self.injectors.pop(index, None)
+
+            # Process moved requests, unidirectional move (a->b) and swap
+            # (a<->b)
```
prashantgupta24, line 57: expanding on this comment would help a lot

tests/scheduling_utils.py
```diff
@@ -235,12 +274,18 @@ def check_scheduler_inference_steps(
...
+        vllm_results.append(output)
 
-    return collected_outputs_new, generated_prompts
+    compare_results(model=model,
+                    tensor_parallel_size=1,
+                    backend=backend,
+                    vllm_results=vllm_results,
+                    hf_results=hf_results,
+                    prompts=prompts)
+    return None, None
```
prashantgupta24, line 291: we don't really need to return anything now?
```suggestion

```

wallashss:
bot:test
TEST_FILE=tests/e2e/test_spyre_cb_scheduler_steps.py MARKERS=quantized

wallashss:
bot:test
TEST_FILE=tests/e2e/test_spyre_cb_scheduler_steps.py MARKERS="spyre and quantized"

tests/golden_token_injector.py
```diff
@@ -0,0 +1,169 @@
...
+            logits[req_idx][expected_token_id] = lp
+
+            # Decode the tokens for better human readability
+            token = self.tokenizer.decode([token_id])
+            expected_token = self.tokenizer.decode([expected_token_id])
+
+            print(f"Golden token injection for request {label}"\
+                  f" at token index '{expectation.current_token_idx}':")
+            print(f"'{token}' {prob.item() * 100:.2f}% -> "
+                  f"'{expected_token}' {expected_prob * 100:.2f}%")
```
prashantgupta24, line 166: Can we add a bit more info to this? Right now it shows as
```
' used' 14.66% -> ' done' 15.25%
```
maybe we can indicate which was the expected token etc.

prashantgupta24: Also not sure if we want this to be a log or a print statement?

prashantgupta24: But either way they're very helpful

wallashss: I changed the log to: 
```
Golden token injection for request '#2' at token index '11':
' used' (16.64%) replaced by ' done' (13.16%), baseline: (15.25%)
```

I realized that I was logging a misleading message. The first two probabilities are from the current decoding. The last one is used to guide the decoding.

wallashss:
bot:test
TEST_FILE=tests/e2e/test_spyre_cb_scheduler_steps.py MARKERS="spyre and quantized"

prashantgupta24:
@wallashss do you plan on adding a unit test for the `tests/golden_token_injector.py` file? There are some edge cases which IMO should be checked through unit tests to make sure we're hitting them all?

tests/golden_token_injector.py
```diff
@@ -0,0 +1,171 @@
...
+                      "Token injection will be skipped")
+                expectation.has_error = True
+                continue
+
+            full_prob = torch.ones(1, dtype=logprobs.dtype)  # 100%
+
+            # Keep the same logprob for the expected token and
+            # redistribute evenly the probability among the other
+            # token ids.
+            # NOTE: se are setting logprobs to the logits, if we recalculate
```
prashantgupta24, line 138: Is this a typo?

wallashss:
> @wallashss do you plan on adding a unit test for the `tests/golden_token_injector.py` file? There are some edge cases which IMO should be checked through unit tests to make sure we're hitting them all?

Can you share with the edge cases please?

prashantgupta24:
I'm assuming the edge cases become all the `if` conditions that we have within the `apply` function of the custom logits processor, right?

tests/golden_token_injector.py
```diff
@@ -0,0 +1,171 @@
...
+        if not batch_update:
+            return
+
+        # Process added requests.
+        for index, params, _, _ in batch_update.added:
+            assert params is not None
+            if params.extra_args and (
+                    injector_dict :=
+                    params.extra_args.get("golden_token_injector")):
+                self.req_states[index] = ExpectationState(injector_dict)
```
maxdebayser, line 64: Suggestion: `**injector_dict`

maxdebayser (changes requested):
In general this LGTM, but in one place the math sees off to me. I've also left some typing suggestions.

tests/golden_token_injector.py
```diff
@@ -0,0 +1,171 @@
...
+from vllm.config import VllmConfig
+from vllm.transformers_utils.tokenizer import get_tokenizer
+from vllm.v1.sample.logits_processor import (BatchUpdate, LogitsProcessor,
+                                             MoveDirectionality)
+
+
+class ExpectationState:
+    '''This class controls the state of the generation.'''
+
+    def __init__(self, input_dict: dict):
```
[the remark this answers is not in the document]
wallashss: Thanks @maxdebayser that's a good idea, I changed to that.

tests/golden_token_injector.py
```diff
@@ -0,0 +1,171 @@
...
+            # NOTE: we are setting logprobs to the logits, if we recalculate
+            # the softmax again over this distribution we shall find the same
+            # values, but with some minimal difference. The intention is
+            # inject the golden token but preserving the original logprob.
+
+            other_token_ids_count = logits.shape[1] - 1
+            other_logprobs = torch.log(
+                (full_prob - prob) / other_token_ids_count)
+
+            if lp < other_logprobs:
```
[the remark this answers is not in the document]
maxdebayser: I see, you're comparing with the average of all other token probs.

PR #482, opened 2025-09-25
🔥 rip out VLLM_SPYRE_TEST_BACKEND_LIST

prashantgupta24:
# Description

We anyway use markers for selecting the tests so this env var wasn't useful


prashantgupta24:
Hmm embedding and quantization tests are failing... Checking

tests/spyre_util.py
```diff
@@ -197,14 +197,12 @@ def get_spyre_model_dir_path() -> Path:
     return Path(model_dir_path)
 
 
-# get model backends from env or default to all and add pytest markers
+# add pytest markers to supported different backends
 def get_spyre_backend_list():
-    user_backend_list = os.environ.get("VLLM_SPYRE_TEST_BACKEND_LIST",
-                                       "eager,inductor,sendnn")
+    backend_list = ["eager", "inductor", "sendnn"]
```
prashantgupta24, line 202: I guess I could simplify this function more, but it does feel more readable this way 🤷

joerunde:
Low risk test-only changes, let's get it in

PR #483, opened 2025-09-26
Only set threading env vars if not already defined

AnishPahilajani:
# Description

This PR ensures that threading-related environment variables (e.g. OPENBLAS_NUM_THREADS, MKL_NUM_THREADS, etc.) are only set if they are not already defined by the user.

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -456,7 +456,8 @@ class SpyrePlatform(Platform):
                 )
 
             for env in THREADING_ENVS:
-                os.environ[env] = str(cpus_per_worker)
+                if not os.environ.get(env):
+                    os.environ[env] = str(cpus_per_worker)
 
             logger.info(
                 "%s for %d workers. Since VLLM_SPYRE_UPDATE_THREAD_CONFIG is "


AnishPahilajani:
I understand the idea behind using only the `VLLM_SPYRE_UPDATE_THREAD_CONFIG` flag to control whether threading environment variables are overridden.

On our Power 11 systems, some threading env vars (like OMP_NUM_THREADS) are set by scripts in the image, while others are unset. If we override all vars unconditionally, it could break existing configurations. That’s why I added the conditional check (`if not os.environ.get(env):`) it ensures we only set vars that aren’t already defined, while still respecting the feature flag.

tjohnson31415:
On the other hand, the "existing configuration" can be sub-optimal for or break vllm-spyre. What if you _want_ to override the system `OMP_NUM_THREADS` setting based on vLLM's configuration?
If `VLLM_SPYRE_UPDATE_THREAD_CONFIG=1` only sets unset env vars, it is not possible to have vllm-spyre override the system settings (which I presume are static and not based on the launched worker count that vLLM knows about).

My thought with `VLLM_SPYRE_UPDATE_THREAD_CONFIG` is that, if you know what you are doing with threading and want to control it, you'd have full control, including optimizing for the specific vLLM worker configuration you are using.

joerunde:
closing as we have an alternative implemented in #487

PR #462, opened 2025-09-12
fix: tests for graph comparison with FP8

wallashss:
# Description

fix: tests for graph comparison with FP8

PR #490, opened 2025-09-29
🔥 remove auto-marked xfail for fp8, include fp8 tests by default, add xfail manually

prashantgupta24:
# Description

Fp8 should now be fully supported (almost). Some tests are failing due to output mismatch, those are marked `xfail` manually.

## Related PRs

fix: static batching with FP8 - https://github.com/vllm-project/vllm-spyre/pull/457
and 
fix: tests for graph comparison with FP8 - https://github.com/vllm-project/vllm-spyre/pull/462


joerunde:
@prashantgupta24 I see we also still have `_skip_quantized_by_default`- do you think it's time to remove that too?

If we're confident in the fp8 tests (except the ones manually marked xfail here) then it may be less confusing to have them all run by default. We would have to update our `pytest` invocations though to e.g. skip fp8 on most of the test matrix here in GHA since the cpus are too slow to run the fp8 models

prashantgupta24:
> @prashantgupta24 I see we also still have `_skip_quantized_by_default`- do you think it's time to remove that too?
> 
> If we're confident in the fp8 tests (except the ones manually marked xfail here) then it may be less confusing to have them all run by default. We would have to update our `pytest` invocations though to e.g. skip fp8 on most of the test matrix here in GHA since the cpus are too slow to run the fp8 models

Could be a follow up PR?

wallashss (approved):
LPGTM thanks!

prashantgupta24:
bot:test

prashantgupta24:
bot:test

PR #491, opened 2025-09-29
[fp8] fix cb scheduler step tests

yannicks1:
this change does always compile for the next power of 2 for the batch size.


yannicks1:
bot:test

joerunde:
> Just for the record, can you add an explanation either here in the PR or in the code as to why this is required?

+1, can we extend the 🌶️🌶️🌶️ comment right above these changes?

wallashss (approved):

PR #494, opened 2025-10-01
[s390x] Update s390x depencies

nikheal2:
# Description

Few pakcages are not available for s390x  and usually built by source, skipping those packages from pyproject and executed `uv lock`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,6 @@ override-dependencies = [
     # Skip packages only on s390x, expected to be pre-installed
     "vllm ; platform_machine != 's390x'",
     "ray; platform_machine != 's390x'",
-    "ray[default]; platform_machine != 's390x'",
     "llvmlite; platform_machine != 's390x'",
     "pyarrow; platform_machine != 's390x'",
 ]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -50,6 +50,13 @@ override-dependencies = [
     "torchaudio; sys_platform == 'never'",
     "torchvision; sys_platform == 'never'",
     "triton; sys_platform == 'never'",
+    
+    # Skip packages only on s390x, expected to be pre-installed
+    "vllm ; platform_machine != 's390x'",
+    "ray; platform_machine != 's390x'",
+    "ray[default]; platform_machine != 's390x'",
```
joerunde, line 57: I think this is causing all the `ray[default]` deps to be pulled in for x86 (see all the uv.lock additions) which is not what we want

uv.lock
```diff
@@ -526,26 +530,6 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/d2/81/e3073017a8f5c75169e79108eda209e6089e3f96c9f197d307cbda7df71c/compressed_tensors-0.11.0-py3-none-any.whl", hash = "sha256:e1cbc46e1ae032b7ceea915fe18c8d2de5a54d3a50a607969b6bdfe703b6cb83", size = 179951, upload-time = "2025-08-19T18:59:29.308Z" },
 ]
 
-[[package]]
-name = "cupy-cuda12x"
```
joerunde, line 530: nice!

Not sure how this was pulled in anyway- good to see it removed. Maybe this was from `ray[cgraph]`?

joerunde (approved):
nice

PR #495, opened 2025-10-01
[CB] 🐛  fix padding of position ids

yannicks1:
When padding the position ids, we should always use 0 (not padding token id !)

As the mask disregards the padded tokens anyway, that does not have an influence on the generation results. It can break the implementation though when the padding token is bigger than the max model length (that is also how I found the bug). 

**_For the models we support so far this was never a problem, but for future ones we should fix it!_**

Note: We did not find that bug earlier as for our models of interest, the padding token id happens to be 0 too, but that does not have to hold true in general.


sducouedic (approved):
LGTM 🚀

PR #496, opened 2025-10-01
Granite4 2b & 3b support

yannicks1:
Supporting granite4 2b & 3b (attention only). 

fms support has been merged into main here: https://github.com/foundation-model-stack/foundation-model-stack/pull/472 

implemented it in a backward compatible way, such that we can get it into vllm-spyre main without relying on a fms version post above PR.

### DIFF
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -341,11 +341,9 @@ class ContinuousBatchingFmsModel(FmsModelBase):
 
         if self.config.model_type in {'llama', 'granite', 'granitemoehybrid'}:
             self.kv_cache_specs['num_layers'] = self.config.num_hidden_layers
-            if 'granite' in self.config.model_type:
-                self.kv_cache_specs['head_dim'] = self.model.config.head_dim
-            else:
-                self.kv_cache_specs['head_dim'] = self.config.hidden_size // \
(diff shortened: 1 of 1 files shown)


vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -353,10 +353,16 @@ def __init__(
         self.kv_cache_specs['num_kv_heads'] = model_config.get_num_kv_heads(
             parallel_config)
 
-        if self.config.model_type in {'llama', 'granite'}:
+        if self.config.model_type in {'llama', 'granite', 'granitemoehybrid'}:
             self.kv_cache_specs['num_layers'] = self.config.num_hidden_layers
             self.kv_cache_specs['head_dim'] = self.config.hidden_size // \
                 self.config.num_attention_heads
+
+            # *** ALERT *** Granite 2b hack for AIU Compiler
```
maxdebayser, line 361: Do we know if this is a temporary hack? Then perhaps we can add a compatibility test

yannicks1: will see how they implement it properly in fms. I do just mirror their strategy here... (very much work in progress)

maxdebayser (approved):
LGTM. I suppose we can add the model to the know supported configurations once we have PELE test coverage.

cc: @ckadner , @joerunde , @jfialho

PR #497, opened 2025-10-01
⏪ revert back to 3-layer micro model

prashantgupta24:
# Description

Revert back to 3-layer micro model

### DIFF
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -265,7 +265,7 @@ def _default_test_models(isEmbeddings=False, isScoring=False):
     # the test command includes `-m quantized`.
     tinygranite = ModelInfo(
         name="ibm-ai-platform/micro-g3.3-8b-instruct-1b",
-        revision="6e9c6465a9d7e5e9fa35004a29f0c90befa7d23f")
+        revision="2714578f54cfb744ece40df9326ee0b47e879e03")
     tinygranite_fp8 = ModelInfo(
         name="ibm-ai-platform/micro-g3.3-8b-instruct-1b-FP8",
         revision="0dff8bacb968836dbbc7c2895c6d9ead0a05dc9e",


prashantgupta24:
FYI bot test seems to be fine with the 3 layer model!

.github/workflows/test.yml
```diff
@@ -174,15 +174,15 @@ jobs:
           key: ${{ runner.os }}-hf-model-${{ env.model_key }}
 
       - name: "Download HF models"
-        if: ( steps.changed-src-files.outputs.any_changed == 'true' && steps.cache_restore.outputs.cache-hit != 'true' )
+        if: ( steps.changed-src-files.outputs.any_changed == 'true' && steps.cache_restore.outputs.cache-hit == 'true' )
```
ckadner, line 177: This means you will have both revisions:
- 4 layer model from the cache
- 3 layer model downloaded with revision

Making all the tests work. If you bypass the cache, I reckon some tests will break like in PR #499

.github/workflows/test.yml
```diff
@@ -174,15 +174,15 @@ jobs:
...
         run: |
           # We are caching HF models (HF_HUB_CACHE) for reliability rather than speed, since HF downloads are flaky for concurrent jobs.
           # Be careful when adding models to the cache here, as the GHA cache is limited to 10 GB.
           # If a new model is added here, a new hash key is generated. The previous cache blob can then
           # be removed by an admin or can be left to expire after 7 days.
 
           download_tinygranite() {
-            python -c "from transformers import pipeline, AutoTokenizer; pipeline('text-generation', model='$1'); tokenizer=AutoTokenizer.from_pretrained('$1')"
+            python -c "from transformers import pipeline; pipeline('text-generation', model='$1', revision='2714578f54cfb744ece40df9326ee0b47e879e03');"
```
ckadner, line 185: was the AutoTokenizer only needed to the 4-layer model (or the FP8 version)?

ckadner (commented):
If it is just tiny granite that we need to change revisions for (occasionally), then your quick fix here is probably sufficient -- after merging your PR and deleting the old cache blobs

.github/workflows/test.yml
prashantgupta24: I'm not sure, if I run the `pipeline` function now it downloads the tokenizer for both revisions for me, so I think it was never needed in the first place

.github/workflows/test.yml
prashantgupta24: Once the 3 layer model is verified, I don't think we need the 4 layer anymore, so we should only be using one revision of the model at all places

ckadner: I think it needs more code/test changes to make sure we don't use the latest/main revision of the tiny granite model anymore/anywhere.

Kicking of a test run here: #502

ckadner: #503

prashantgupta24:
Don't think we need this anymore

PR #500, opened 2025-10-02
:bug: implement better checking for granite

### LINKED ISSUE #498: Name mismatch for granite model name
### Description

It seems vllm-spyre has a hardcoded name check [here](https://github.com/vllm-project/vllm-spyre/blob/88350f83496fcb01cbb98a12b96c84171dcaa871/vllm_spyre/model_executor/model_loader/spyre.py#L391), which assumes the mode name to be called in certain way.  In real scenarios, it is possible that the model is not following that pattern and hence this will cause following problems:
1. `NUM_BLOCKS_SPYRE` will not get set correctly 
2. If it is pre-compiled model, then there will be a mismatch

Another place with similar issue: https://github.com/vllm-project/vllm-spyre/blob/main/vllm_spyre/platform.py#L213

joerunde:
# Fix granite-3.3-8b detection

Fixes #498 

This PR updates the logic that checks for granite-3.3-8b-instruct. Instead of matching on the model name, it checks the hf config type and checks for specific values related to the 8b granite variant.

This also consolidates all of the granite-specific overrides to one method inside `platform.py`. This should make it much easier to both:
- Update these overrides for granite 3.3 8b 
- Add overrides for other models

in the future.

The number of blocks override for the spyre kv cache size was particularly interesting. We were in a halfway state where we took overriddes from `--num-gpu-blocks-override` and saved them off to platform.py and then used the override to change the scheduler's behavior. I've updated things so that the scheduler no longer needs that extra override so that we can simply use the configured override everywhere. This is more robust since the config value is serialized and given to subprocesses, and more maintainable than managing two separate numbers for this purpose.

### DIFF
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -57,8 +57,6 @@ class SpyreCausalLM(nn.Module):
         rank: int,
     ) -> None:
         super().__init__()
-        self.vllm_config = vllm_config
-
         self.logits_processor = LogitsProcessor(
             vllm_config.model_config.hf_config.vocab_size,
             logits_as_input=True)
@@ -163,7 +161,6 @@ class FmsModelBase(nn.Module):
 
         # Actual FMS model
         self.model: nn.Module
-        self.vllm_config = vllm_config
         self.model_config = vllm_config.model_config
         self.parallel_config = vllm_config.parallel_config
         self.cache_config = vllm_config.cache_config


vllm_spyre/platform.py
```diff
@@ -518,4 +505,52 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+        # If no HDMA p2psize override was specified, set 256MB
+        if not os.getenv("FLEX_HDMA_P2PSIZE", None):
+            p2psize_256m = 256 * 1024 * 1024
+            os.environ["FLEX_HDMA_P2PSIZE"] = str(p2psize_256m)
+            logger.info(
+                "Model granite-3.3-8b-instruct and tensor parallel size 4 "
+                "detected. Using FLEX_HDMA_P2PSIZE = %d", p2psize_256m)
+
+    @classmethod
+    def is_granite_33_8b(cls, model_config: ModelConfig):
```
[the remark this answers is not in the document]
joerunde: sure- might even go just `granite_3_8b` since I think these are all the same with 3.1 and 3.2 too

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -388,7 +388,7 @@ def get_num_blocks_available(self) -> int:
 
         # TODO: replace the hard coded NUM_BLOCKS_SPYRE by calling a function
         # in torch_sendnn which returns the value set by the Spyre compiler.
-        if ('granite-3.3-8b-instruct' in self.model_config.model
+        if (SpyrePlatform.is_granite_33_8b(self.model_config)
```
prashantgupta24, line 391: I wonder if we have to account for other granite models down the road

vllm_spyre/platform.py
```diff
@@ -209,28 +210,14 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         os.environ["VLLM_DT_MAX_BATCH_SIZE"] = str(
             max(vllm_config.scheduler_config.max_num_seqs, 2))
 
+        # Hardcode some things for granite-3.3-8b-instruct
+        cls.configure_granite_33_8b(vllm_config)
```
tjohnson31415, line 214: Very much a style NIT and totally optional, but I think it reads better if the `if not cls.is_granite_33_8b(vllm_config.model_config):` check is moved here instead of embedded in `configure_granite_33_8b`.

vllm_spyre/model_executor/model_loader/spyre.py
joerunde: Very likely, given current timeframes for model support

joerunde:
bot:test
MARKERS="spyre and cb and not quantized"

vllm_spyre/platform.py
```diff
@@ -540,3 +513,72 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+            # only override configs for TP=4
+            return
+
+        tkv_128k = 128 * 1024
+        if not os.getenv("VLLM_DT_MAX_BATCH_TKV_LIMIT", None):
+            os.environ["VLLM_DT_MAX_BATCH_TKV_LIMIT"] = str(tkv_128k)
+            logger.info("Model granite-3.3-8b-instruct and tensor parallel " \
+            "size 4 detected. Using VLLM_DT_MAX_BATCH_TKV_LIMIT = %d",
+            tkv_128k)
+        else:
```
tjohnson31415, line 535: NIT: This `else` doesn't check the value of the env var, so a warning will be produced even if set to the expected value.

Could make it an
```
elif os.getenv("VLLM_DT_MAX_BATCH_TKV_LIMIT") != str(tkv_128k):
```

vllm_spyre/platform.py
```diff
@@ -540,3 +513,72 @@ def get_max_output_tokens(self, prompt_len: int) -> int:
...
+        # If no HDMA p2psize override was specified, set 256MB
+        p2psize_256m = 256 * 1024 * 1024
+        if not os.getenv("FLEX_HDMA_P2PSIZE", None):
+            os.environ["FLEX_HDMA_P2PSIZE"] = str(p2psize_256m)
+            logger.info(
+                "Model granite-3.3-8b-instruct and tensor parallel size 4 "
+                "detected. Using FLEX_HDMA_P2PSIZE = %d", p2psize_256m)
+        else:
+            logger.warning(
+                "FLEX_HDMA_P2PSIZE was set to %d, not using the "
```
tjohnson31415, line 550:
```suggestion
                "FLEX_HDMA_P2PSIZE was set to %s, not using the "
```

vllm_spyre/platform.py
```diff
@@ -209,28 +198,16 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
-                    "Model granite-3.3-8b-instruct and tensor parallel size 4 "
-                    "detected. Using FLEX_HDMA_P2PSIZE = %d",
-                    1024 * 1024 * 256)
-        else:
-            # default value for any other model/ tensor parallel size
+        # Hardcode some things for granite-3.3-8b-instruct
+        if cls.is_granite_3_8b(vllm_config.model_config):
+            cls.configure_granite_3_8b(vllm_config)
+
+        if not os.getenv("VLLM_DT_MAX_BATCH_TKV_LIMIT", None):
```
tjohnson31415, line 205: NIT:
```suggestion
        if not os.getenv("VLLM_DT_MAX_BATCH_TKV_LIMIT"):
```
The default `default` for `os.getenv` is `None`. Maybe it is just me but adding the explicit default makes it harder for me to read.

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -51,17 +51,17 @@ class SpyreCausalLM(nn.Module):
...
-        model_config: ModelConfig,
-        parallel_config: ParallelConfig,
-        scheduler_config: SchedulerConfig,
+        vllm_config: VllmConfig,
         max_prompt_length: int,
         max_decode_length: int,
         rank: int,
     ) -> None:
         super().__init__()
+        self.vllm_config = vllm_config
```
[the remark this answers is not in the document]
joerunde: Ah, yeah I had used this when the kv block stuff was still in this file. I still like the change to only pass `vllm_config` to the initializers, but I can remove the unused `self.vllm_config`s  too

joerunde:
bot:test
MARKERS="spyre and cb and not quantized"

PR #502, opened 2025-10-02
[DON'T MERGE] <[CI Test]> Don't use cache for 4-layer granite model (PR #497)

ckadner:
# Description

Draft PR #497 has all tests pass because it uses both tiny granite models

- the 4-layer model from the cache
- the 3-layer model downloaded with revision

This PR checks the test results for bypassing the cache (not using 4-layer model)

## Related Issues

#497


prashantgupta24:
Going to push commits to this PR to test the model.revision bug fixes!

prashantgupta24:
So my plan at this time is to open a new PR with the latest commits from this PR which fix the model revision used in the tests. Once we get that in then we can discuss about whether we even want to revert to the 3 layer model or not

PR #507, opened 2025-10-07
:bug: fixup full_model marker

joerunde:
# Description

This PR fixes a bug where the test collection failed on the full-sized models because we injected a string instead of a `ModelInfo` for the `model` param. This both:
1. Updates the full-sized model parameterization to use a ModelInfo with a pinned revision
2. Updates the test collection code to not fail if a string is passed anyway


maxdebayser (approved):
LGTM

PR #510, opened 2025-10-07
test: update tests to use golden token injection

wallashss:
# Description

This PR is a follow up for #478 

It updates the remaining tests with the function `check_output_against_hf` and replace by the new `validate_vllm_vs_hf_output` that does "everything": it generates for vllm and hf, where the results of hf are used as baselines for the golden token injection, afterwards it compares the results of both systems. 

I also included these envs, to get more flexibility to setup the tests:
- `VLLM_SPYRE_TEST_DISABLE_GOLDEN_TOKEN` - to disable the golden token injection
- `VLLM_SPYRE_TEST_ABS_TOL` - override the constant `ISCLOSE_ABS_TOL` 
- `VLLM_SPYRE_TEST_QUANTIZED_ABS_TOL` override the constant `ISCLOSE_ABS_TOL_QUANTIZATION`


wallashss:
bot:test

PR #519, opened 2025-10-10
[profiler] fix multi-aiu profiling and add setable options

mcalman:
# Description

1. Switches to an alternative method for enabling AIU event profiling when profiling is enabled with `VLLM_TORCH_PROFILER_DIR`. This resolves the following error that occurs when trying to profile a multi-AIU workload: `RuntimeError: Please register PrivateUse1HooksInterface by 'RegisterPrivateUse1HooksInterface' first. `
2. Adds support for the following profiling options that were added to vLLM: https://github.com/vllm-project/vllm/blob/30f78af147cb9eac0a5c643a69882a3b0e45f986/docs/contributing/profiling.md?plain=1#L10-L13

## Related Issues

n/a


maxdebayser (approved):
LGTM

PR #532, opened 2025-10-15
:art: Swap to ruff and prek

joerunde:
# Description

Popping in vllm's current pre-config setup, seeing how easy it'd be to replace our current setup with this

---

**Update:**

Using [`prek`](https://prek.j178.dev/diff/) instead of pre-commit (`prek` is a much faster Rust based swap-in for pre-commit)


pyproject.toml
```diff
@@ -63,8 +63,8 @@ environments = [
 ]
 
 [tool.ruff]
-# Allow lines to be as long as 80.
-line-length = 80
+# Use widescreen monitors 😎😎😎
+line-length = 120
```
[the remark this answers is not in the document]
joerunde: lol yeah I wanted to see what the negative diff here is and it's like 600 lines of savings

It seems pretty fine on my laptop screen, but let me know what you think!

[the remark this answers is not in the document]
ckadner: I like 120 for my local dev, but when looking at GitHub PR diffs side-by-side, the 80 character line limit seems to work better. With 120 character, Github needs to wrap those lines and then Python's semantic indentation gets lost and it's harder to track context when reading the code changes.

tests/aftu/test_compare_graphs.py
```diff
@@ -26,42 +30,53 @@
...
+        "--compile_dynamic",
+        "--min_pad_length",
+        "64",
+        "--max_new_tokens",
+        "5",
+        "--batch_size",
+        str(max_num_seqs),
+        "--compile_dynamic_sendnn",
+        "--attention_type",
+        attn_type,
```
ckadner: This does not occur often, but in cases like this one I would favor readability over consistency and compactness (or line width) in cases like this one to group parameter names and values on one line (if the code formatting police allows it)

```Python
    inference_py_args = [
        sys.executable, "scripts/inference.py",
        "--architecture", "hf_pretrained",
        "--model_path", model_path,
        "--tokenizer", model_path,
        "--unfuse_weights", 
        "--device_type", "aiu",
        "--compile",
        "--cast_bf16_to_fp16", 
        "--compile_dynamic",
        "--min_pad_length", "64",
        "--max_new_tokens", "5",
        "--batch_size", str(max_num_seqs),
        "--compile_dynamic_sendnn",
        "--attention_type", attn_type
```

tests/e2e/test_logits_processors.py
```diff
@@ -125,24 +119,17 @@ def apply(self, logits: torch.Tensor) -> torch.Tensor:
...
-                             temperature=0,
-                             logprobs=0,
-                             ignore_eos=True)
-    params2 = SamplingParams(max_tokens=7,
-                             temperature=0,
-                             logprobs=0,
-                             ignore_eos=True)
+    params0 = SamplingParams(max_tokens=5, temperature=0, logprobs=0, ignore_eos=True)
+    params1 = SamplingParams(max_tokens=10, temperature=0, logprobs=0, ignore_eos=True)
+    params2 = SamplingParams(max_tokens=7, temperature=0, logprobs=0, ignore_eos=True)
```
ckadner: 👍🏻 this is better than before

tests/e2e/test_spyre_cb.py
```diff
@@ -237,35 +243,42 @@ def test_swap_decode_programs_for_cb(
     sampling_params_16k = [create_sampling_params(p16k) for _ in range(1)]
     sampling_params_32k = [create_sampling_params(p32k) for _ in range(1)]
 
-    sampling_params = sampling_params_1k + sampling_params_2k + \
-        sampling_params_4k + sampling_params_8k + sampling_params_16k + \
-            sampling_params_32k
+    sampling_params = (
```
ckadner: nice

pyproject.toml
ckadner: Here is an example of readability getting worse at least when looking at Github diffs:

image

or here

image

ckadner (commented):
Overall this seems to be a positive change. Maybe we can tweak some of our code conventions a little bit and allow a bit more freedom in favor of code readability in some places.

maxdebayser:
I agree with @ckadner's  assessment. Overall it's positive, but 120 character lines seem to make diffs more difficult to read.

ckadner:
@joerunde -- I swapped out pre-commit in favor of [`prek`](https://prek.j178.dev/diff/). And I had to make a few extra changes to get the markdown linter to collaborate.

PR #535, opened 2025-10-16
:bug: Fix fp8 model name check with quantization check

gkumbhat:
# Description

- Replace FP8 model name check with quantization check since user could choose to change their model name and then things can go into weird state, and if they are trying to use pre-compiled model, then it will go into re-compilation

### DIFF
--- a/vllm_spyre/v1/worker/spyre_worker.py
+++ b/vllm_spyre/v1/worker/spyre_worker.py
@@ -444,7 +444,8 @@ class SpyreWorker(WorkerBaseV1):
             0, len(valid_token_ids_tensor), (3, prompt_len))]
 
         # TODO: we need 2 requests for warmup on FP8+CB
-        is_fp8_plus_cb = 'FP8' in self.model_config.model and \
+        # Check if model is quantized
+        is_fp8_plus_cb = self.model_config.quantization is not None and \
             envs_spyre.VLLM_SPYRE_USE_CB
         req_count = 3 if is_fp8_plus_cb else 2
         requests = [


vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -444,7 +444,8 @@ def _warmup_spyre_dynamic_size(self, special_token_ids):
             0, len(valid_token_ids_tensor), (3, prompt_len))]
 
         # TODO: we need 2 requests for warmup on FP8+CB
-        is_fp8_plus_cb = 'FP8' in self.model_config.model and \
+        # Check if model is quantized
```
ckadner: What about other quantizations like 4 bit or 8 bit `int`?

There is some more code from @joerunde and some I recently wrote to do this kind of check, which could be adapted and used here:

- https://github.com/vllm-project/vllm-spyre/blob/de0e3d2c6e18ad21baca4833d947e1f9f5df82f8/vllm_spyre/platform.py#L597
- https://github.com/vllm-project/vllm-spyre/blob/de0e3d2c6e18ad21baca4833d947e1f9f5df82f8/vllm_spyre/config/runtime_config_validator.py#L139

joerunde (approved):
Lgtm too

joerunde:
bot:test
MARKERS="spyre and quantized and not multi"

joerunde:
I don't believe in the bot tests any more but we might as well try to kick one off to make sure nothing barfs in a new and unusual way

vllm_spyre/v1/worker/spyre_worker.py
ckadner: Created a follow-up issue: https://github.com/vllm-project/vllm-spyre/issues/537

PR #536, opened 2025-10-17
feat: improve golden token injection

wallashss:
# Description

PR #540, opened 2025-10-21
feat: improve golden token injection

maxdebayser:
A rebase of @wallashss's PR https://github.com/vllm-project/vllm-spyre/pull/536 without the LogitsProcessor refactors.

Improvements to the GoldenTokenInjector (GTI):
- update GTI args to support parsing from JSON string, eg. passed through the vLLM OpenAI `vllm_xargs` extension
- support always injecting the golden token regardless of expected logprobs / divergence checks
- GTI is now part of the vllm_spyre package; it can be enabled with: `--logits-processors vllm_spyre.v1.sample.golden_token_injector:GoldenTokenInjector`

Example usage:
```
VLLM_SPYRE_USE_CB=1 vllm serve ibm-ai-platform/micro-g3.3-8b-instruct-1b --max-num-seqs 16 --max-model-len 1024 --logits-processors vllm_spyre.v1.sample.golden_token_injector:GoldenTokenInjector
```

```
curl http://localhost:8000/v1/completions \
    -d '{
  "model": "ibm-ai-platform/micro-g3.3-8b-instruct-1b",
  "prompt": "a squared plus b squared equals ",
  "vllm_xargs": {
     "golden_token_injector": "{\"expected_token_ids\": [85, 39560]}"
  },
  "logprobs": 0,
  "return_tokens_as_token_ids": true,
  "max_tokens": 2,
  "temperature": 0
}'
```
to generate "c squared" instead of "16".

### DIFF
--- a/vllm_spyre/v1/sample/golden_token_injector.py
+++ b/vllm_spyre/v1/sample/golden_token_injector.py
@@ -121,7 +121,6 @@ class GoldenTokenInjector(LogitsProcessor):
 
         current_token_idx = len(expectation.output_token_ids)
 
-
         if not current_token_idx < len(expectation.token_ids):
             logger.warning_once(
                 "Request %s does not have enough expected tokens "


tjohnson31415:
bot:test

tests/e2e/test_spyre_max_new_tokens.py
```diff
@@ -38,7 +38,9 @@ def test_output(model: ModelInfo, stop_last: bool, max_model_len: int,
         logprobs=0,  # return logprobs of generated tokens only
         ignore_eos=False)
 
-    vllm_sampling_params = [vllm_sampling_params_normal] * 3
+    vllm_sampling_params = [
+        vllm_sampling_params_normal.clone() for _ in range(3)
+    ]
```
tjohnson31415: This change is needed now that all tests use GTI by default (before this PR, only tests using `get_engine` would use GTI). Haivng a copied reference instead of a `.clone()` meant that all sequences had the same GTI config (even with different prompts).

waleedqk:
bot:test
MARKERS="spyre and cb and not quantized and not multi"

wallashss:
bot:test

PR #545, opened 2025-10-28
fix: min_tokens > 1 causes long generation with continuous batching

tjohnson31415:
# Description

The MinTokensLogitsProcessor needs to get a `batch_update` at each step to detect when enough tokens have been generated. The `LogitProcessorWrapper` copied the typical logic of skipping updates when batch_update is None, but this meant that min tokens would not get the needed call to `update_state`.

The fix here is to always call `update_state` on each of the wrapped logitsprocs in the batch, with a some extra code to not call `update_state` for a particular index more than once.

## Related Issues

Follow up to https://github.com/vllm-project/vllm-spyre/pull/544 which fixed the behavior for static batching
Cherry-picked improvement to test_sampling_params.py from https://github.com/vllm-project/vllm-spyre/pull/536


wallashss (approved):
LPGTM, tks!

gkumbhat:
Tested this PR and it seems to be resolving the problem with min_token

Issue #550, opened 2025-10-31
[Bug] Delay on first response when requesting `logprobs`


tjohnson31415:
I have found the first request to a fresh server that includes `logprobs` to have an increased response time compared to subsequent requests.

Repro with a simple server invocation:
```
VLLM_SPYRE_USE_CB=1 vllm serve ibm-ai-platform/micro-g3.3-8b-instruct-1b --max-model-len 1024 --max-num-seqs 1
```
Sending a request like:
```
time curl http://localhost:8000/v1/completions \
    -H "Content-Type: application/json" \
    -d '{
        "model": "ibm-ai-platform/micro-g3.3-8b-instruct-1b",
        "prompt": "1 2 3",
        "max_tokens": 1,
        "logprobs": 0
    }'
```

In my test environment, the first requet takes ~4.7s and subsequent requests take ~0.3-0.4s.

In my investigation, I found the time increase to be from a `torch.compile` that happens for [vLLM's `batched_count_greater_than`](https://github.com/vllm-project/vllm/blob/01efc7ef781391e744ed08c3292817a773d654e6/vllm/v1/sample/ops/logprobs.py#L10) function hit only when requesting logprobs.

PR #551, opened 2025-11-05
fix: use sampling during warmup and disable backed_size_oblivious after model compilation

tjohnson31415:
# Description

A couple of changes related to compilation of operations during sampling.
- the [`batched_count_greater_than` function](https://github.com/vllm-project/vllm/blob/b8b302cde434df8c9289a2b465406b47ebab1c2d/vllm/v1/sample/ops/logprobs.py#L11) requires compilation and is used to compute logprobs.
- we also found that this function will fail to compile in Pytorch 2.8.0 and 2.9.0 when `backed_size_oblivious` is enabled, so this PR disables `backed_size_oblivious` after model compilation.

## Related Issues

Fixes https://github.com/vllm-project/vllm-spyre/issues/550

### DIFF
--- a/vllm_spyre/v1/worker/spyre_worker.py
+++ b/vllm_spyre/v1/worker/spyre_worker.py
@@ -103,6 +103,17 @@ def _maybe_warmup_context(limit: int, world_size: int, rank: int):
         _inside_warmup_mode = False
 
 
+@contextlib.contextmanager
+def use_torch_fx_backed_size_oblivious():
+    # this setting is required to mark a dimension of size 1 as dynamic
+    # for pytorch >= 2.7.1 (needed to support batch size 1 for decodes)
+    # NB: this setting is disabled at the end of this function
+    from torch.fx.experimental import _config as config
+    config.backed_size_oblivious = True
+    yield
+    config.backed_size_oblivious = False
+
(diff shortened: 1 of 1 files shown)


vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -657,6 +666,11 @@ def _dynamic_warmup(
         prompt_len: int,
         valid_token_ids_tensor: torch.Tensor,
     ) -> None:
+        # this setting is required to mark a dimension of size 1 as dynamic
+        # for pytorch >= 2.7.1 (needed to support batch size 1 for decodes)
+        # NB: this setting is disabled at the end of this function
```
joerunde: It'd probably be best to toss this in a context manager to ensure it gets unset, since we don't currently have a test matrix set up to test different versions of torch for compatibility.

```
with torch_backed_size_oblivious():
    ...
```

joerunde: We also likely have a new chunked prefill warmup coming very soon so this would be able to be reused there

joerunde:
bot:test
MARKERS="cb and spyre and not quantized and not multi"

PR #552, opened 2025-11-06
feat: chunked prefill spyre model runner

wallashss:
# Description

New model runner with chunked prefill support 

- The implementation is behind a feature flag: VLLM_SPYRE_USE_CHUNKED_PREFILL
- I changed few code of static batching to have more realistic scheduler output that would work better with chunked prefill.
- Disabled code related to pipeline parallel which was messing up some outputs and it is not yet used with spyre.

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -186,6 +186,8 @@ class SpyrePlatform(Platform):
                     f"`{scheduler_config.max_num_batched_tokens}`. Please "
                     "set `--max-num-batched-tokens` to a number that satisfy "
                     "this constraint.")
+                os.environ["VLLM_DT_CHUNK_LEN"] = \
+                    str(scheduler_config.max_num_batched_tokens)
 
         logger.info(
             "Overriding configurations based on warmup shapes. "


sducouedic:
I had a quick look and it seems very very nice and clean :)

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+
+        # Reserve the number of blocks that this new sequence requires in the
+        # worst case (it might always stop early by producing the EOS token)
+        new_tokens = (sampling_params.max_tokens
+                      if sampling_params is not None else 0)
+        total_tokens = prompt_len + new_tokens - 1
+        # subtract the padding blocks from the reserved blocks
+        n_reserved_blocks = math.ceil(total_tokens / self.block_size)
+
+        self.req_ids2reserved_blocks[req_id] = n_reserved_blocks
```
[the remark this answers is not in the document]
wallashss: IIUC you're suggesting to prefill some chunks ahead of time and wait blocks get free to finish their prefill/decode later.

That's sounds dangerous, but interesting and tempting 😄 . After seriously thinking about it, the issue that we'd having is regarding the input_batch. First you'd need to track these cases in the scheduler side but scheduler would also need to inform model runner that this request won't be processed, like a preemption, so we'd need to rebuild the input batch removing these partial prefills from the step. Therefore, I think it is too complex atm, and we need first validate that CP is working.

I think it is an interesting idea that we should document it in an issue, but maybe we should try to run few benchmarks and try to identify cases that this optimization could bring benefits.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+        self.input_batch.add_request(req_state)
+        self.prefill_batch.add_request(req_state)
+
+        # Refresh sampling metadata after all request are added to the batch
+        self.input_batch.refresh_metadata()
+        self.prefill_batch.refresh_metadata()
+
+    def prepare_model_input(self, scheduler_output):
+
+        is_prefill = False
```
sducouedic: what are the possible number of `scheduler_output.scheduled_new_reqs` and `scheduler_output.scheduled_cached_reqs` here? Is is max 1 for each of them, and at least one of the two should be non-empty? maybe we can assert that ?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+            if left_padding == 0:
+                # Case III
+                # No left padding, it will start with full chunks
+                token_idx = chunk_i * chunk_size
+            else:
+                # Case II - remaining chunks
+                # The `left_pad_blocks_offset` is the piece of the first
+                # chunk that needed to be incomplete to compensate left
+                # padding. Sum it with the remaining chunks count to get
+                # where to slice the prompt
```
sducouedic: it is a bit difficult to phrase clearly, but suggestion: 

`left_pad_blocks_offset` is the number of prompt tokens used in the first chunk, which is not a multiple of the chunk size due to left padding. Sum this value with the offset of the current chunk to know where to slice the prompt.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+        # Case I
+
+        Prompt fits in the chunk with left padding
+
+        3 tokens
+        1 block
+        1 chunk
+        4 left padding
+
+        X X X X | T T T O || O O O O | O O O O || O O O O | O O O O || ...
```
joerunde: Nice drawings!

I would probably omit the trailing `0 0 0 0` blocks to remove any confusion that we're pre-allocating a long sequence of right padding

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+
+        request = self.requests[req_id]
+        prompt_token_ids = request.prompt_token_ids
+
+        chunk_size = self.scheduler_config.max_num_batched_tokens
+        chunk_blocks = chunk_size // self.block_size
+        num_computed_tokens = request.num_computed_tokens
+        # basic check, if the computed tokens is divisible by the
+        # chunk size
+        assert num_computed_tokens % chunk_size == 0
```
joerunde: This is checking that the number of previously computed tokens is divisible by the chunk size, right?

If so I don't think that's true- when we do left padding with full blocks for the first chunk then we will compute a number of tokens smaller than the chunk size

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+                # chunk that needed to be incomplete to compensate left
+                # padding. Sum it with the remaining chunks count to get
+                # where to slice the prompt
+                token_idx = left_pad_blocks_offset + (chunk_i - 1) * chunk_size
+
+            input_tokens_np[:chunk_end - token_idx] = (
+                prompt_token_ids[token_idx:min(token_idx +
+                                               chunk_size, prompt_len)])
+            input_positions_np[:chunk_end - token_idx] = range(
+                token_idx, chunk_end)
```
sducouedic: there is something wrong with `chunk_end`, I just tried with a long prompt that spans 3 chunks: `prompt_len` = 260, `chunk_size` = 128. 
For the second chunk, we have `chunk_end = 256`, and this raises error because now we go over the size of `input_tokens` which is 192 so `range(token_idx, chunk_end)` produces more values than the tensor can accept. Also `chunk_end - token_idx` produces value that doesn't make sense imo.

uv.lock
```diff
@@ -940,13 +940,13 @@ wheels = [
 
 [[package]]
 name = "ibm-fms"
-version = "1.4.0"
+version = "1.5.0"
```
joerunde: We'll need a corresponding bump to the lower bound in pyproject.toml as well, right?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+        # Get request id from new request or cached request
+        req_id = scheduler_output.scheduled_new_reqs[0].req_id if \
+            len(scheduler_output.scheduled_new_reqs) == 1 \
+                else scheduler_output.scheduled_cached_reqs.req_ids[0]
+
+        request = self.requests[req_id]
+        prompt_token_ids = request.prompt_token_ids
+
+        chunk_size = self.scheduler_config.max_num_batched_tokens
+        chunk_blocks = chunk_size // self.block_size
```
joerunde: These should be constant so they could be computed once in the initializer and held as class attributes

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+
+            return CBSpyreModelRunnerOutput(
+                req_ids=list(req_id_to_index.keys()),
+                req_id_to_index=req_id_to_index,
+                sampled_token_ids=[],
+                logprobs=None,
+                # TODO: probably it does not makes sense
+                prompt_logprobs_dict=prompt_logprobs_dicts,
+                pooler_output=[],
+                tkv=self.tkv)
```
sducouedic: Need to set `n_free_blocks`

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,565 @@ def execute_model(
...
+        code to create the chunks, not necessarily enumerate the possible 
+        scenario. Of course there are interpretations where these cases 
+        overlaps. 
+        
+        '''
+
+        # Get request id from new request or cached request
+        req_id = scheduler_output.scheduled_new_reqs[0].req_id if \
+            len(scheduler_output.scheduled_new_reqs) == 1 \
+                else scheduler_output.scheduled_cached_reqs.req_ids[0]
```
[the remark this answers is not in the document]
wallashss: Change the signature of this method to receives the request id already, the assertions are before this call.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,561 @@ def execute_model(
...
+                                        pooler_output=[],
+                                        num_nans_in_logits=None,
+                                        tkv=self.tkv)
+
+    @SpyrePlatform.inference_mode()
+    def execute_model(
+        self,
+        scheduler_output: SchedulerOutput,
+        **kwargs,
+    ) -> ModelRunnerOutput:
```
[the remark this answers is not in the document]
wallashss: It makes sense for sure, but I'd prefer in this PR to not touch in original implementation, I think also we'll pollute it with rules of the child class, sound not so good for me. IMO it is not significant the code duplication there. 
I propose to review this after we have the full implementation because more changes would be needed and the isolation and flexibility is better for now.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,624 @@ def execute_model(
...
+        model_output = CBSpyreModelRunnerOutput(
+            req_ids=list(req_id_to_index.keys()),
+            req_id_to_index=req_id_to_index,
+            sampled_token_ids=output.sampled_token_ids.tolist(),
+            logprobs=(output.logprobs_tensors.tolists()
+                      if output.logprobs_tensors else None),
+            prompt_logprobs_dict=prompt_logprobs_dicts,
+            pooler_output=[],
+            tkv=self.tkv,
+            n_free_blocks=self.get_n_free_blocks())
```
sducouedic: Would it be possible still in this PR to send the correct number of `num_computed_tokens` for the prefilling request back to the scheduler? From the discussions it seems like the best way would be through `CBSpyreModelRunnerOutput`. See [my comment in the scheduler changes](https://github.com/vllm-project/vllm-spyre/blob/085068eb2907893210c49faef541ce9620c154dd/vllm_spyre/v1/core/scheduler.py#L436-L439).

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,624 @@ def execute_model(
...
+        10 tokens
+        4 blocks
+        2 chunks
+        4 left padding
+
+        X X X X | T T T T || T T T T | T T O O || 
+        
+        # Case III 
+
+        No left paddings and more than one chunk
```
yannicks1: *padding

yannicks1 (commented):
great stuff! I read most of it and put some minor comments which you can address in your next PR. Will read the rest later and might provide some more. 

Sorry for being late to the party:)

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,624 @@ def execute_model(
...
+        # set the new tkv to the prompt length if starting a new decode batch
+        if is_new_batch:
+            self.tkv = prompt_len
+
+        # Reserve the number of blocks that this new sequence requires in the
+        # worst case (it might always stop early by producing the EOS token)
+        new_tokens = (sampling_params.max_tokens
+                      if sampling_params is not None else 0)
+        total_tokens = prompt_len + new_tokens - 1
+        # subtract the padding blocks from the reserved blocks
```
yannicks1: please update this comment as you do not subtract any padding blocks here..

vllm_spyre/v1/worker/spyre_model_runner.py
yannicks1: just for completeness (and my understanding;)

> what are the possible number of scheduler_output.scheduled_new_reqs and scheduler_output.scheduled_cached_reqs here? Is is max 1 for each of them

`scheduler_output.scheduled_cached_reqs` can be > 1 for decodes because it holds all the requests of the current decode batch, right ? So its max is the max batch size. 

For chunked prefills however `scheduler_output.scheduled_cached_reqs` is 0 for the first chunk (because it is in `scheduler_output.scheduled_new_reqs`) and 1 for all the remaining chunks (for these `scheduler_output.scheduled_new_reqs` is 0). Can you confirm @wallashss?
cc @sducouedic

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,624 @@ def execute_model(
...
+                    else scheduler_output.scheduled_cached_reqs.req_ids[0]
+            model_inputs = self._prepare_chunked_prefill(req_id)
+
+            self._maybe_prepare_last_prefill(req_id, scheduler_output)
+
+            return model_inputs
+        else:
+            return self._prepare_decode(scheduler_output.scheduled_cached_reqs)
+
+    def get_empty_output(self):
```
yannicks1: maybe you can put a comment here why we cannot use `EMPTY_MODEL_RUNNER_OUTPUT`, as in the static continuous batching case.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1740,3 +1741,624 @@ def execute_model(
...
+                pooler_output=[],
+                tkv=self.tkv,
+                n_free_blocks=self.get_n_free_blocks())
+
+        # Sample the next token.
+        output: SamplerOutput = self.model.sample(
+            logits=logits,
+            sampling_metadata=self.get_sampling_metadata(is_prefill),
+        )
+        t1 = time.time() - t0
```
yannicks1: while the timing makes sense for decodes I believe it will only print the time of the last chunk when doing prefill. we should also add a debugg log into the above if statement. and possibly print whether it is a timing for a decode or chunked prefill step.

PR #553, opened 2025-11-07
Model runner tests for chunked prefill

joerunde:
# Description

Tests the new chunking and batching logic at the model runner level. This should be almost ready to go- I've pointed it into @wallashss's branch so it can merge into his open PR #552 

If we decide to pass back the number of computed tokens from the model runner, these tests can be updated to take advantage of that. Otherwise they should be good to review

### DIFF
--- a/tests/e2e/test_chunked_prefill_tkv_steps.py
+++ b/tests/e2e/test_chunked_prefill_tkv_steps.py
@@ -18,6 +18,7 @@ from vllm.v1.engine.core import EngineCore
 from vllm.v1.outputs import ModelRunnerOutput
 from vllm.v1.request import Request, SamplingParams
 
+from vllm_spyre.platform import SpyrePlatform
 from vllm_spyre.v1.worker.spyre_model_runner import SpyreModelRunner
 
 
@@ -146,7 +147,8 @@ def test_single_block_chunked_prefill(model: ModelInfo, max_model_len: int,
         monkeypatch=monkeypatch)
 
     req_id = "1"
-    prompt_len = 48
+    # Sub-block prompt
+    prompt_len = SpyrePlatform.get_block_size() - 12
     new_req_data = make_new_request_data(req_id, prompt_len)
 
     scheduler_output = make_scheduler_output(
@@ -160,6 +162,17 @@ def test_single_block_chunked_prefill(model: ModelInfo, max_model_len: int,
     # tkv is prompt_len
     assert output.tkv == prompt_len
 
+    # One decode pass to ensure no extra padding shenanigans
+    cached_req_data = make_cached_request_data({req_id: prompt_len})
+    scheduler_output = make_scheduler_output(
+        scheduled_new_reqs=[],
+        scheduled_cached_reqs=cached_req_data,
+        num_scheduled_tokens={req_id: 1})
+    output = runner.execute_model(scheduler_output)
+    assert len(output.sampled_token_ids[0]) == 1
+    # No extra block or chunk padding
+    assert output.tkv == prompt_len + 1
+
 
 @pytest.mark.cpu
 @pytest.mark.chunked_prefill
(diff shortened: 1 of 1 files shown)


joerunde:
@wallashss I'm getting these failures atm:
```
====================================================================================================================== short test summary info ======================================================================================================================
FAILED tests/e2e/test_chunked_prefill_tkv_steps.py::test_multi_chunk_padded_prefill[ibm-ai-platform/micro-g3.3-8b-instruct-1b-max_model_len(512)-max_num_seqs(2)-chunk_size(128)] - AssertionError
FAILED tests/e2e/test_chunked_prefill_tkv_steps.py::test_multi_chunk_unpadded_prefill[ibm-ai-platform/micro-g3.3-8b-instruct-1b-max_model_len(512)-max_num_seqs(2)-chunk_size(128)] - AssertionError: assert 200 == (200 + 1)
FAILED tests/e2e/test_chunked_prefill_tkv_steps.py::test_decode_padding_to_same_block[ibm-ai-platform/micro-g3.3-8b-instruct-1b-max_model_len(512)-max_num_seqs(2)-chunk_size(128)] - RuntimeError: stack expects a non-empty TensorList
======================================================================================================= 3 failed, 1 passed, 4 deselected, 6 warnings in 6.03s =======================================================================================================
```

I believe the assertions for `test_multi_chunk_padded_prefill` and `test_multi_chunk_unpadded_prefill` are correctly catching errors with the model runner's logic, but I don't know why the `RuntimeError: stack expects a non-empty TensorList` error is happening- that may be a bug in this test setup

tests/e2e/test_chunked_prefill_tkv_steps.py
```diff
@@ -0,0 +1,350 @@
...
+    # Scheduler will give first 128 token chunk
+    scheduler_output = make_scheduler_output(
+        scheduled_new_reqs=[new_req_data], num_scheduled_tokens={req_id: 128})
+    output = runner.execute_model(scheduler_output)
+
+    # no output tokens
+    assert len(output.sampled_token_ids) == 0
+    # No left padding block so all 128 tokens are computed
+    # assert output.computed_tokens[req_id] == 128
+    computed_tokens = {req_id: 128}
```
yannicks1: same comment here: assert that `max_num_batched_tokens` is 128.

tests/e2e/test_chunked_prefill_tkv_steps.py
```diff
@@ -0,0 +1,350 @@
...
+    step and should not be padded out to the chunk boundary on decode"""
+    runner: SpyreModelRunner = get_cpu_model_runner(
+        model=model,
+        max_model_len=max_model_len,
+        max_num_seqs=max_num_seqs,
+        max_num_batched_tokens=max_num_batched_tokens,
+        monkeypatch=monkeypatch)
+
+    req_id = "1"
+    prompt_len = 48
```
yannicks1: comment that `prompt_len < chunksize = 128`

tests/e2e/test_chunked_prefill_tkv_steps.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+@pytest.mark.cpu
+@pytest.mark.chunked_prefill
+@default_test_params
+def test_multi_chunk_unpadded_prefill(model: ModelInfo, max_model_len: int,
+                                      max_num_seqs: int,
+                                      max_num_batched_tokens: int,
+                                      monkeypatch: pytest.MonkeyPatch):
+    """A request that's longer than a chunk is split into multiple chunks"""
```
yannicks1: would be nice to mention the (non existing left) padding in the docstring here:  e.g need two chunks 2*128 = 256, 256 - 200 = 56 -> falls within last block of last chunk, hence no left padding needed...

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2064,9 +2064,9 @@ def add_new_request(self, request: NewRequestData):
 
         blocks_count = math.ceil(prompt_len / self.block_size)
 
-        # set the new tkv  to the block padding if starting a new decode batch
+        # If starting a new batch, the initial TKV will be prompt_len + 1
         if is_new_batch:
```
yannicks1: hmm, what is the motivation behind changing this? 

We previously had tkv pointing to the position (0-indexed) we currently decode/ sample the token. 
e.g with prompt length 64, tkv was 64 after the prefill (which is pointing to the 65th token slot, the token we get from prefill). 
After doing the first decode tkv was 65 (we sample the token in the 66th token slot, and fill the KV cache of the 65th token slot). that would require a adding a new block to write the KV cache into. 

when having tkv = 65 after prefill already, that would at least conceptually already require a new block... 

PS: I have not read @wallashss PR yet, maybe it becomes clear then...

joerunde: yeah, I think the root cause here was confusion due to the fact that we set the tkv based on the input from the scheduler and not based on internal state from the model runner. Since I originally had the tests input the current sequence length as `num_computed_tokens`, that caused the tkv to jump to prompt_len + 2 at the first decode pass which looked very strange. It seems like the correct behavior is to pass `num_computed_tokens=sequence_length - 1` in from the scheduler (at least that's what vllm does today), I guess a token isn't "computed" until it has been passed back into the model as input.

joerunde:
Thanks for the review @yannicks1 !

Very good points about all the constants here, I've refactored so that the test inputs are all derived from the parameters where necessary. If the defaults are updated then the tests should still work

PR #554, opened 2025-11-11
ChunkedPrefillSpyreScheduler: No Interleaving

sducouedic:
Need to merge #552 First ✅

# Description

Introduce the new scheduler class for chunked prefill, which handles ongoing chunked prefills that were previously considered as running decodes by previous implementation. Specifically, the new scheduler does the following:
* Differentiate between chunked prefills and decodes in `self.running`
* Prioritize chunked prefill over decode, and prevent scheduling them together
* Prevents new requests to be scheduled when a chunked prefill is still ongoing

Note that for now we don't interleave prefills with decode steps, this allow the two following simplifications:
1. Make sure that chunked prefills per se is correctly implemented and consecutive chunked prefills produce exactly the same output as before.
2. We can re-use the save conditions verifications used for `ContinuousBatchingSpyreScheduler` exactly as before, at the time of prefilling the first chunk, because chunked prefill are consecutive and won't change blocks allocation or tkv value.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,8 +7,7 @@ on:
   pull_request:
     # add labeled and unlabeled to the default types (runs when label is added)
     types: [opened, synchronize, reopened, labeled, unlabeled, auto_merge_enabled]
-    # temporary hack: run tests into wallas-cp-new
-    branches: [main, wallas-cp-new]
+    branches: [main]
 
   push:
     branches: [main]


vllm_spyre/v1/core/scheduler.py
```diff
@@ -411,3 +411,145 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
+                    "requests", len(self.running), len(holdback_queue))
+
+        # Check new requests to prefill
+        elif len(self.waiting) > 0:
+            self.prefilling.extend(self.waiting)
+            # Hide current decodes from the scheduler
+            running_holdback = self.running
+            self.running = []
+            logger.debug(
+                "Scheduling a prefill step of %d requests, holding back %d "
```
yannicks1: this is technically also a chunked prefill step (just the first one). we might wanna update the logging message?

vllm_spyre/platform.py
```diff
@@ -128,8 +128,12 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
             os.environ["COMPILATION_MODE"] = "offline"
 
         if envs_spyre.VLLM_SPYRE_USE_CB and is_decoder:
-            scheduler_config.scheduler_cls = "vllm_spyre.v1.core."\
-                "scheduler.ContinuousBatchingSpyreScheduler"
+            if envs_spyre.VLLM_SPYRE_USE_CHUNKED_PREFILL:
```
joerunde: If we don't already, we should probably either:
1. Set `VLLM_SPYRE_USE_CB=1` if `VLLM_SPYRE_USE_CHUNKED_PREFILL` is set, or
2. Raise an error if `VLLM_SPYRE_USE_CHUNKED_PREFILL` is set and `VLLM_SPYRE_USE_CB` is not

vllm_spyre/platform.py
```diff
@@ -179,6 +183,9 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 scheduler_config.max_num_batched_tokens = (
                     model_config.max_model_len * scheduler_config.max_num_seqs)
             else:
+                # TODO remove before merging: temporary hack to see passing CP
+                # tests without actually implement more tests
+                scheduler_config.max_num_batched_tokens = 128
```
joerunde: are these hacks still needed? The tests that we have should set max_num_batched_tokens appropriately

vllm_spyre/v1/core/scheduler.py
```diff
@@ -411,3 +411,140 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
+            # TODO agree on interface with CBSpyreModelRunnerOutput
+            # req.num_computed_tokens = \
+            # model_runner_output.req_num_computed_tokens[req.request_id]
+            pass
+
+        # Remove completed prefills
+        self.prefilling = [
+            req for req in self.prefilling
+            if req.num_computed_tokens < req.num_prompt_tokens
+        ]
```
joerunde: Not sure if we have a better suggestion but @tjohnson31415 and I were both briefly confused because `self.prefilling` reads like it's a dequeue like `self.waiting` and `self.running` but it's not.

Maybe having the name just not end with `-ing` would help?

sducouedic: what about `ongoing_prefills`?

PR #556, opened 2025-11-13
Increase default KV cache size for granite 8b

yannicks1:
changes:
- increasing number of blocks from 2080 to 8192


yannicks1:
yeah, this makes sense! did not mean to merge this as is, but rather to have a branch to test this. 
we can add the checks here once we know how to do it.

tests/models/test_granite.py
```diff
@@ -50,7 +50,7 @@ def test_granite_3_8b_overrides():
             cache_config=NO_SWAP_CONFIG,
         )
 
-        assert granite_3_8b_config.cache_config.num_gpu_blocks_override == 2080
+        assert granite_3_8b_config.cache_config.num_gpu_blocks_override == 8192
```
maxdebayser: As travis mentioned, you could check the torch_sendnn version to see if it already has support 8K. The version can be found: `torch_sendnn._version.__version__`. But I think the code should use 8192 as the default and downgrade to 2080 if an old version of torch sendnn is found instead of the contrary because if someone is hacking on a local environment where they have editable installs directly from local git repos the version information might be wrong. Also `0.0.0` is the same thing as "version not found": https://github.ibm.com/ai-chip-toolchain/torch_sendnn/blob/546c73752389137f3ef18ffa7f18e99476bc87e4/torch_sendnn/_version.py#L19C17-L19C22

vllm_spyre/platform.py
```diff
@@ -618,7 +618,22 @@ def configure_granite_3_8b(cls, vllm_config: VllmConfig):
...
         # TODO: remove this once we have correct free memory info available
-        blocks_override = 2080
+
+        if envs_spyre.VLLM_SPYRE_DYNAMO_BACKEND == "sendnn":
+            try:
+                from torch_sendnn import torch_sendnn  # noqa: F401
+                version_str = torch_sendnn._version.__version__
+                version = tuple(map(int, version_str.split(".")))
+            except ImportError:
+                print("WARNING: Disabled: torch_sendnn")
```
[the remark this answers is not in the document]
yannicks1: I did copy paste form [here](https://github.com/vllm-project/vllm-spyre/blob/223269873ba9689915c1d37d4f4df27e29ce8fbb/vllm_spyre/model_executor/model_loader/spyre.py#L269) and [here](https://github.com/vllm-project/vllm-spyre/blob/223269873ba9689915c1d37d4f4df27e29ce8fbb/vllm_spyre/v1/worker/spyre_model_runner.py#L1578). I am wondering what the original motivation was? I see these snippets were added by @wallashss and @maxdebayser. Can you give more insight why we should handle it like this, or do you agree with Joe?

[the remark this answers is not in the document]
wallashss: IIRC This was just moved from a global import to this snippet for a lazy importing. Before, this conditional importing makes sense. But Joe is right, we don't have to catch an import error, I think we could just remove the try catch and let it raises the import error.

tests/models/test_granite.py
```diff
@@ -36,21 +38,31 @@ def test_granite_3_8b_detection():
 
 
 @pytest.mark.cpu
-def test_granite_3_8b_overrides():
+@pytest.mark.parametrize("sendnn_configured, sendnn_version, expected_blocks",
```
yannicks1: so clean 😍 !

vllm_spyre/platform.py
```diff
@@ -656,3 +669,23 @@ def is_granite_3_8b(cls, model_config: ModelConfig):
                 and model_config.hf_config.vocab_size == 49159
                 and model_config.hf_config.num_key_value_heads == 8
                 and model_config.hf_config.num_attention_heads == 32)
+
+    @classmethod
+    def sendnn_configured(cls) -> bool:
```
yannicks1: makes sense to have it as a classmethod. I actually also had it like that, but then took your [comment](https://github.com/vllm-project/vllm-spyre/pull/556#discussion_r2623906191) too literally (`...with mock.patch("vllm_spyre.platform.sendnn_configured") as ...`) 😄

yannicks1 (commented):
Looks great. thanks for helping out @joerunde !

I guess the next step is to run this with torch_sendnn installed and check that the version is picked up correctly. And verify that the newest stack works with 8k blocks.

yannicks1:
bot:test
MARKERS="spyre and (cb or chunked_prefill) and full_model and not quantized"

Abhishek-TAMU:
bot:test
MARKERS="full_model and prefix_caching and not quantized"

PR #557, opened 2025-11-13
fix: left padding of prompts less than chunk size

wallashss:
# Description

fix: left padding of prompts less than chunk size


yannicks1 (approved):
makes sense. lgtm

wallashss:
That's depends on what you call recently. The bug was introduced when I did a refactoring and fixed the bug of more than 2 chunks you reported it. In a quick debug I could confirm that the model inputs were clearly wrong. Thinking better now, I think makes sense to review tests to cover this case, I'll do that next.

PR #558, opened 2025-11-13
Interleave chunked prefills with single decoding steps

sducouedic:
# Description

* Deprioritize chunked prefill when previous step was already a prefill and requests are decoding
* Evaluate blocks condition at the time of first chunked prefill
* Evaluate remaining conditions at the time of last chunked prefill

* Tests will be implemented other PRs:
    * interleaving logic correctness: e.g. that after step X which was a prefill, step X+1 is a decode
    * ✅ DONE constraint correctness: verify that we cannot schedule first chunked prefill when there isn't enough blocks

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -689,10 +689,11 @@ class ChunkedPrefillSpyreScheduler(ContinuousBatchingSpyreScheduler):
         if request in self.ongoing_prefills:
             return True
 
-        # Check that there is space in the prefill batch
-        max_prefill_batch_size = 1
-        num_scheduled_prefills = len(self.waiting) + len(self.ongoing_prefills)
-        return num_scheduled_prefills < max_prefill_batch_size
+        # We can start prefilling a new requests if we satisfy the maximum
+        # number of concurrent prefills
+        max_concurrent_prefills = 1
+        num_prefills = len(self.waiting) + len(self.ongoing_prefills)
+        return num_prefills < max_concurrent_prefills
 
     def check_batch_tkv_limit_cp(self, request, new_req_tkv, n_blocks, running,
                                  max_batch_tkv_limit) -> bool:


vllm_spyre/v1/core/scheduler.py
```diff
@@ -523,28 +594,126 @@ def schedule(self) -> "SchedulerOutput":
...
+
+        if not self._has_scheduling_priority(request):
+            return False
+
+        return self._satisfies_constraints(request)
+
+    def _satisfies_constraints(self, request: Request) -> bool:
+        is_first_chunk = request.num_computed_tokens == 0
+        is_last_chunk = (request.num_prompt_tokens -
+                         request.num_computed_tokens) <= self.chunk_size
```
joerunde: general comment here is that it looks like we have separate sets of constraints for
1. accepting a new prefill (is_first_chunk)
2. starting decodes for the last prefill (is_last_chunk)

Can we separate these into different methods to make it clear which is which?

```
can_schedule = True
if is_first_chunk:
    can_schedule = _satisfies_prefill_constraints(req)
if is_last_chunk and can_schedule:
    can_schedule = _satisfies_decode_constraints(req)

return can_schedule
```

vllm_spyre/v1/core/scheduler.py
```diff
@@ -414,29 +415,91 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
+        - Prefill batch size: prefill batch size if of 1, only one request's
+            chunk prefill can be scheduled at a time
+            
+        - Decode batch size: cannot have more than batch-size number of 
+            running requests, including prefill and decode
+
+        - Prompt fits the TKV (this constraint can be relaxed by setting
+            `VLLM_SPYRE_N_TOKENS_PREFILL_PRIO` variable to 1): the prompt
+            length should not exceed the current maximum TKV of all the
+            running requests
```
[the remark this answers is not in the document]
sducouedic: ah yes right! I'll remove the doc description for this constraint, but it is probably better to always do the optimization in the another PR I guess

vllm_spyre/v1/core/scheduler.py
```diff
@@ -392,29 +393,82 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
-    """ Support of chunked prefill """
+    """ 
+    # TODO also add all this to vllm-spyre documentation
+    Chunked-Prefill Scheduling policy
+
+    The prefill vs. decode priority policy is the following:
+        - Current prefill request priority: No new request's chunked prefill
+            can start as long as there another request's prefill is on-going
+            
+        - Prefills step interleaving: The prefill steps are interleaved with
```
tjohnson31415:
```suggestion
        - Prefill step interleaving: The prefill steps are interleaved with
```

vllm_spyre/v1/core/scheduler.py
```diff
@@ -392,29 +393,82 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
+            one decode steps: as long as there are decoding requests, two
+            prefill steps cannot be consecutive
+
+        - General prefill priority: conditioned on interleaving constraint,
+            prefill has priority over decode
+
+        - No empty step: if a prefill step is prevented because it doesn't
+            satisfy Spyre's specific constraints, a decode step is scheduled
+
+    Additional Spyre's specific constraints:
```
tjohnson31415:
```suggestion
    Spyre scheduling constraints:
```

vllm_spyre/v1/core/scheduler.py
```diff
@@ -392,29 +393,82 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
+        - Blocks constraint: all the blocks necessary to serve a request are
+            allocated at the time of scheduling the first chunk of a chunked
+            prefill. For the first chunked prefill, there must be enough
+            blocks to accommodate the prompt and the maximum number of output
+            tokens
```
tjohnson31415:
```suggestion
            prefill. There must be enough blocks to accommodate the prompt and
            the maximum number of output tokens
```

vllm_spyre/v1/core/scheduler.py
```diff
@@ -392,29 +393,82 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
+
+        Note: all the remaining constraints need to be satisfied at the time
+        of scheduling the last chunk of a chunked prefill
+        
+        - Prefill batch size: prefill batch size if of 1, only one request's
+            chunk prefill can be scheduled at a time
+            
+        - Decode batch size: cannot have more than batch-size number of 
+            running requests, including prefill and decode
```
tjohnson31415:
```suggestion
        - Decode batch size: cannot have more than max_num_seqs
            running requests, including prefill and decode
```

vllm_spyre/envs.py
```diff
@@ -172,6 +173,12 @@ def _backend_backwards_compat() -> str:
     # single prefill is used.
     "VLLM_SPYRE_USE_CHUNKED_PREFILL":
     lambda: bool(int(os.getenv("VLLM_SPYRE_USE_CHUNKED_PREFILL", "0"))),
+
+    # Feature Flag
+    # Works only with chunked prefill enabled. If set, prefill steps are
+    # interleaved with a decode step
+    "VLLM_SPYRE_CP_USE_INTERLEAVING":
```
yannicks1: nit: does `VLLM_SPYRE_CP_INTERLEAVE_STEPS` sound better? 
I cannot come up with the perfect name for this.. 😄

vllm_spyre/v1/core/scheduler.py
```diff
@@ -376,29 +378,83 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
+
+        - Remaining space after TKV: the number of requested tokens must fit
```
yannicks1: Remaining space after TKV -> Max model length constraint

vllm_spyre/v1/core/scheduler.py
```diff
@@ -543,30 +657,44 @@ def can_schedule(self, request):
                                          req.num_prompt_tokens)
             max_tokens_remaining = req.max_tokens - n_generated_output_tokens
             # note that the -1 comes from the token we generate during prefill
-            cond3_current = max_tokens_remaining - 1 <= (max_context_len -
+            cond2_current = max_tokens_remaining - 1 <= (max_context_len -
                                                          dec_req_tkv)
-            cond3 = cond3 and cond3_current
+            cond2 = cond2 and cond2_current
             # early exiting loop if violated 3rd condition
-            if not cond3:
```
yannicks1: also update 3rd -> 2nd

vllm_spyre/v1/core/scheduler.py
```diff
@@ -543,30 +657,44 @@ def can_schedule(self, request):
...
+
+    def _has_scheduling_priority(self, request):
+        decoding_requests = [
+            r for r in self.running if r not in self.ongoing_prefills
+        ]
+
+        # If we do interleaving, then two consecutive prefill steps are
+        # forbidden when there are decoding requests
+        if self.do_interleaving and self.previous_step_was_prefill \
+            and decoding_requests:
```
yannicks1: very small nit: I would personally prefer `len(decoding_requests) > 0` for better readability...

vllm_spyre/v1/core/scheduler.py
```diff
@@ -376,29 +378,83 @@ def check_batch_tkv_limit(self, request, tkv, running,
...
+
         # We want to keep track of requests for which the prefill is ongoing.
         # Theoretically, only one request can be prefilled at a time, but we
         # keep a list to be able to batch prefills in the future.
         self.ongoing_prefills: list[Request] = []
 
+        # Prefills interleaving: if the feature flag is set, prefill operations
+        # are interleaved with a decode step. This allows to minimize currently
+        # decoding requests
+        self.do_interleaving: bool = envs_spyre.VLLM_SPYRE_CP_USE_INTERLEAVING
```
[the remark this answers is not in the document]
sducouedic: now that VLLM_SPYRE_CP_USE_INTERLEAVING has been renamed to VLLM_SPYRE_CP_INTERLEAVE_STEPS, I think "do" sounds nicer actually. okay with that?

vllm_spyre/v1/core/scheduler.py
```diff
@@ -612,7 +740,9 @@ def check_batch_tkv_limit_cp(self, request, new_req_tkv, n_blocks, running,
         decode_req_max_tkvs.sort()
 
         # Initialize values
-        batch_size = len(running) + 1
+        batch_size = len(running)
+        if request not in running:
+            batch_size += 1
```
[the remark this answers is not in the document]
sducouedic: that's right

vllm_spyre/v1/core/scheduler.py
```diff
@@ -447,32 +511,33 @@ def schedule(self) -> "SchedulerOutput":
             # them separately. Either we schedule a chunked prefill step, or a
             # decoding step
 
-            schedule_prefill = self.can_schedule(self.ongoing_prefills[0])
```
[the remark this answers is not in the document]
yannicks1: it is tiny and not part of this PR but the previous one.

```
        assert len(self.ongoing_prefills) <= 1, \
            "Only one request can be prefilled at a time, but got %d" \
                % len(self.ongoing_prefills)
        assert len(self.waiting) == 0 or len(self.ongoing_prefills) == 0, \
        "Cannot schedule new requests while another request prefill is ongoing."
        assert all(r in self.running for r in self.ongoing_prefills), \
        "Ongoing prefill requests must be in the running queue."
```
to 
```
        assert len(self.ongoing_prefills) <= 1, \
            "Only one request can be prefilled at a time, but got %d" \
                % len(self.ongoing_prefills)
        assert len(self.waiting) == 0 or len(self.ongoing_prefills) == 0, \
            "Cannot schedule new requests while another request prefill is ongoing."
        assert all(r in self.running for r in self.ongoing_prefills), \
            "Ongoing prefill requests must be in the running queue."
```
note: you might need more lines to silence the linter... 

we do have intends (4 spaces) for all other assert messages, just looks odd to not follow the same convention here...

PR #559, opened 2025-11-13
feat: left padding from model runner to scheduler

wallashss:
# Description

This PRs adds the `CPSpyreModelRunnerOutput` that includes a dictionary that maps the left padding to help scheduler be aware if all scheduled tokens were actually computed in a prefill. 

Extra:
- Expanded a test to include one more case: prompt that fits in a single chunk.

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -2275,6 +2275,11 @@ class ChunkedPrefillModelRunner(ContinuousBatchingSpyreModelRunner):
             for req_id, idx in req_id_to_index.items()
         }
 
+        # TODO: dead code, this only works for SB with bs=1, either
+        # fix it or remove it.
+        prompt_logprobs_dicts = self._get_prompt_logprobs_dict(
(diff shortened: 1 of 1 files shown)


tests/e2e/test_chunked_prefill.py
```diff
@@ -37,7 +40,9 @@ def get_model_runner(cp_model: LLM):
 
 USE_CASES = {
     # Case I - Prompt fits in a single chunk
-    "case_I": (prompt_95, 128, 1, 0),
+    "case_Ia": (prompt_95, 128, 1, 0),
+    # Case I - Prompt fits in a single chunk with left padding
+    "case_Ib": (prompt_51, 128, 1, 64),
     # Case II - Has left padding
     "case_II": (prompt_260, 128, 3, 64),
     # Case III again - no padding
```
sducouedic: why "again" here?

btw not for this PR, but I find a bit odd that single chunk with or without left padding are both case I, but case II when having left padding becomes case III

sducouedic: yes I see, cases are aligned with the implementation. Thanks a lot for detailed explanation

PR #564, opened 2025-11-18
[CP] optimal chunked prefill scheduler constraints

yannicks1:
replaces the conservative upper bounds for `cond3` and `cond5` with the per sequence optimal values for chunked prefill.  

changes:
- implements optimal conditions for max model len and batch x tkv constraints (`cond3` and `cond5`). the latter needed to re-implement the `check_batch_tkv_limit()` function for chunked prefill. 
- scheduler step level test for above conditions (+ max number of blocks constraint)
- expand engine caching to capture chunk size and add hf cache for new tests
- variable renaming for CB scheduler to have consistent var names that are intuitive

Note that the implemented test are working for both scheduler policies with and without interleaving prefills and decodes.

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -316,30 +316,32 @@ class ContinuousBatchingSpyreScheduler(SpyreScheduler):
                 return tkv <= lower
 
         # Compute the effective token length of the new request
-        new_req_tkv = tkv + request.max_tokens - 1
+        new_req_max_tkv = tkv + request.max_tokens - 1
 
         # Compute token lengths for all running requests (decode batch)
-        decode_req_tkvs = [
+        decode_req_max_tkvs = [
             tkv + req.max_tokens - 1 -
             (req.num_computed_tokens - req.num_prompt_tokens)
             for req in running
         ]
         # Sort decode requests token lengths in ascending order
-        decode_req_tkvs.sort()
+        decode_req_max_tkvs.sort()
 
         # Initialize values
         batch_size = len(running) + 1
         max_batch_tkv = 0
 
         # Try adding the new request to the batch and check the max volume
-        for decode_req_tkv in decode_req_tkvs:
-            if new_req_tkv <= decode_req_tkv:
+        for decode_req_max_tkv in decode_req_max_tkvs:
+            if new_req_max_tkv <= decode_req_max_tkv:
                 # If the new request is shorter, it limits the batch volume
-                max_batch_tkv = max(max_batch_tkv, batch_size * new_req_tkv)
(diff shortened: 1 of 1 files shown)


maxdebayser (approved):
LGTM

vllm_spyre/v1/core/scheduler.py
```diff
@@ -544,18 +544,22 @@ def can_schedule(self, request):
         # - prompt_len < self.tkv and fall into different blocks
         prompt_len = request.num_prompt_tokens
         n_blocks = math.floor(max(self.tkv, prompt_len) / self.block_size)
-        max_tkv = n_blocks * self.block_size + max(
-            self.tkv % self.block_size, prompt_len % self.block_size)
         new_tkv = n_blocks * self.block_size + prompt_len % self.block_size
```
sducouedic: If I understand correctly, this tkv won't be the "new tkv" (ie, the next self.tkv value) but the tkv of the new joining prompt specifically. Since we don't have a single global and absolute tkv value anymore, and this is quite a major concept change, I would try to be very specific in the naming of the tkv values, to help and assimilate the new understanding, that makes sense.

```suggestion
        new_req_tkv = n_blocks * self.block_size + prompt_len % self.block_size
```

vllm_spyre/v1/core/scheduler.py
```diff
@@ -544,18 +544,22 @@ def can_schedule(self, request):
...
         for req in self.running:
-            cond3_current = req.max_tokens <= (max_context_len - max_tkv)
+            # current tkv of the (left aligned) decode sequence
+            tkv = n_blocks * self.block_size + \
+                req.num_computed_tokens % self.block_size
+            n_generated_output_tokens = (req.num_computed_tokens -
+                                         req.num_prompt_tokens)
+            max_tokens_remaining = req.max_tokens - n_generated_output_tokens
+            # note that the -1 comes from the token we generate during prefill
+            cond3_current = max_tokens_remaining - 1 <= (max_context_len - tkv)
```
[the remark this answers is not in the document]
yannicks1: yeah, that is why we need the -1. We had no -1 in the continuous batching scheduler.

vllm_spyre/v1/core/scheduler.py
yannicks1: if I change it here, the variables mean different things in continuous batching vs chunked prefill... which is bad. 
changing the CB variable names as part of this PR might also be a bit confusing as this is a pr about CP. 
I would probably keep CP var names consistent with CB var names here, and if needed address with a variable change in a follow up PR for both CP and CB. WDYT @sducouedic ?

sducouedic:
NP, I just had the tests remaining to review, had a quick view, looks pretty good 👍

PR #575, opened 2025-11-21
Add step tests for chunked prefill 

maxdebayser:
# Description

This PR add step by step execution tests for the chunked prefill scheduler. It tests several scenarios:

1) Single sequence with several chunks
2) Two sequences scheduled at the same time, one with 1 chunk and the other with 4
3) Same as 2 but with interleaving disabled to test this debugging feature
4) Same as 2 but the second requests arrives later.

The tests show that the interleaving of requests is working as expected.

### DIFF
--- a/tests/aftu/test_compare_graphs.py
+++ b/tests/aftu/test_compare_graphs.py
@@ -170,7 +170,7 @@ def test_compare_graphs_static_batching(model: ModelInfo,
 
 
 @pytest.mark.spyre
-@pytest.mark.cb
+@pytest.mark.chunked_prefill
 def test_compare_graphs_chunked_prefill(model: ModelInfo, max_num_seqs: int,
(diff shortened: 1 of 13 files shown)


wallashss (approved):
LGTM

tests/llm_cache.py
```diff
@@ -33,6 +35,8 @@ def __init__(self, teardown_method: Callable[[T], None] | None = None):
         else:
             self._teardown = teardown_method
 
+        self._preexisting_max_tkv = os.getenv("VLLM_DT_MAX_BATCH_TKV_LIMIT")
```
tjohnson31415: Tests that modify `VLLM_DT_MAX_BATCH_TKV_LIMIT` should be using `monkeypatch` which would automatically restore the value after the test. Can we use that instead of having a special case reset here?

PR #577, opened 2025-11-21
feat/fix: add finish_requests to handle removal from ongoing_prefills

tjohnson31415:
# Description

Implements `finish_requests()` for `ChunkedPrefillSpyreScheduler` to handle removing requests that have been cancelled from ongoing_prefills.

Without this, cancelling a large prefill during chunked prefill would result in a crash:
```
  File "/home/senuser/vllm-spyre/vllm_spyre/v1/core/scheduler.py", line 478, in schedule
    assert all(r in self.running for r in self.ongoing_prefills), \
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Ongoing prefill requests must be in the running queue.
```

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -3,7 +3,7 @@
 import math
 import os
 from collections import deque
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Iterable, Union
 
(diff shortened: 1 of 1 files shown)


maxdebayser (approved):
LGTM

PR #585, opened 2025-11-30
[PC] Refactor CB model runner to use vLLMs block pool

maxdebayser:
# Description

This PR paves the way to better integration with vLLMs KV Cache manager and also for the reuse of the prefix caching algorithms and data structures that are already integrated in the BlockPool and the KV cache managers.

In this PR, the padding blocks are managed outside of the KV cache manager which will make it easier to take the block mapping directly from the scheduler in the future.

For now this PR uses the FullAttentionManager which is a SingleTypeKVCacheManager for simplicity. The KV cache coordinators and the KVCacheManager are layers of indirection around this which deal with multiple types of attention and multiple KV Cache Groups. For now, since we don't support any of this yet I prefer to keep this PR simple.

In general this integration is pretty straigthforward. The only impedance mismatch that I've found is that the kv cache manager interface deals with numbers of tokens, while in the spyre model runner, we sometimes need to work directly with block allocation.

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -867,7 +867,12 @@ class ContinuousBatchingSpyreModelRunner(SpyreModelRunner):
         self.block_pool = BlockPool(num_gpu_blocks=self.n_blocks + 1,
                                     enable_caching=False,
                                     enable_kv_cache_events=False)
-        attn_spec = KVCacheSpec(block_size=self.block_size)
+        attn_spec = FullAttentionSpec(
+            block_size=self.block_size,
+            # dummy values
+            num_kv_heads=1,
+            head_size=1,
+            dtype=torch.float16)
         self.kv_cache_manager = FullAttentionManager(
             kv_cache_spec=attn_spec,
             block_pool=self.block_pool,


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -812,8 +815,6 @@ def __init__(
 
         self.block_size = SpyrePlatform.get_block_size()
 
-        # TODO: move to a KV cache manager
-        self.req_ids2blocks: dict[str, deque[int]] = {}
         # max number of blocks needed (reserved) per request id
         self.req_ids2reserved_blocks: dict[str, int] = {}
```
yannicks1: it seems to me that `self.req_ids2reserved_blocks` is the last thing related to block management left in the model runner. As far as I can tell there is no concept of reserved blocks in the `FullAttentionManager`. We could potentially derive a custom class from `FullAttentionManager` that adds `req_ids2reserved_blocks`... Not sure if this is something we want to do. Advantage: all block managing is then happening in the kv cache manager (not model runner and kv cache manager), downside: we need a custom class (which only adds one field tho...). WDYT?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -992,16 +1009,17 @@ def _prepare_prompt(
         self.req_ids2reserved_blocks[req_id] = n_reserved_blocks
 
         # filling block table and slot mapping
-        blocks = []
+
+        blocks = self.kv_cache_manager.allocate_new_blocks(
+            req_id, right_padding_tkv)
+
         slots = []
         for pos_i in range(right_padding_tkv):
```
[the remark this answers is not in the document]
maxdebayser: Yes, good idea. I ended up doing it in this PR with torch operations to be faster.

vllm_spyre/v1/worker/spyre_model_runner.py
maxdebayser: Note: the volumetric constraint is probably always lower than the available number of blocks. We need to verify this information and resolve this question in a future PR. @yannicks1

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -863,7 +864,22 @@ def complete_warmup(self) -> None:
     def _set_blocks(self, num_blocks: int) -> None:
         # set number of available blocks and populate block_pool
         self.n_blocks = num_blocks - 1
-        self.block_pool = deque([i for i in range(1, self.n_blocks + 1)])
+        self.block_pool = BlockPool(num_gpu_blocks=self.n_blocks + 1,
+                                    enable_caching=False,
+                                    enable_kv_cache_events=False)
+        attn_spec = FullAttentionSpec(block_size=self.block_size,
```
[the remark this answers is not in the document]
maxdebayser: Actually, this might be tricky to do for now. We set the block size to the max model len in platform.py to disable vllm's paged attention scheduling: https://github.com/vllm-project/vllm-spyre/blob/d3c8b428b9064190aeec0f2e53f48b016d9bf7ad/vllm_spyre/platform.py#L183

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -992,16 +1014,17 @@ def _prepare_prompt(
         self.req_ids2reserved_blocks[req_id] = n_reserved_blocks
```
tdoublep: I was a bit confused reading the code because I assumed this dictionary mapped the request id to the list of block IDs that were reserved, and couldn't understand how that interacted with the KV cache manager. Now I understand it is just the *number* of reserved blocks. Maybe we could change the name to something like `req_ids2num_reserved_blocks` or something (as a follow-up).

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1094,24 +1116,31 @@ def _prepare_decode(
         for req_id in req_ids:
             # adding new blocks if needed
             if self.tkv % self.block_size == 0:
-                self.req_ids2blocks[req_id].append(self.block_pool.popleft())
-            n_blocks = max(n_blocks, len(self.req_ids2blocks[req_id]))
+                req_state: SamplingRequestState = self.requests[req_id]
+                total_tokens = req_state.left_padding % self.block_size \
+                    + req_state.num_computed_tokens + 1
```
tdoublep: Could we add a comment to explain this formula (especially the `req_state.left_padding % self.block_size` bit)?

vllm_spyre/v1/worker/spyre_model_runner.py
yannicks1: we are actively discussing this. I agree that we should not change behavior in this PR. This PR is only refactoring/integrating upstream code. 

For #586 , follow up PRs: 
the current implementation in #586 (still using the reserved block concept) does not consider blocks with reference counts > 1 (prefix hits) neither in the scheduler nor in the model runner (which actually 'reserves' the number of blocks). 

As we opted for a prefix-caching unaware scheduler, the minimal thing we should do (probably in #586) is to modify the model runner to consider prefix hits when reserving the number of blocks. The scheduler is then still unaware of prefix hits when making decision for a new sequence, but the total number of available blocks is less conservative as it is considering the duplicates in the existing decode batch (blocks with reference count > 2)

In a next step (follow up PR) we can remove the concept of reserved blocks. I believe this should be doable by proving that the volumetric constraint is always stricter than the number of blocks constraint. if that is not given, we could indeed support preemption.

PR #588, opened 2025-12-04
test: a few test configuration updates to have chunked prefill tests pass on Spyre

tjohnson31415:
# Description

In our internal CI we run unit tests on Spyre devices. This PR fixes some of the new Chunked Prefill tests to be able to pass on Spyre.

The two issues causing tests to fail:
- chunk size greater than max model length (due to vLLM default of 2048 or granite TP4 detection setting it to 4096)
- some edge case I don't fully understand in `test_single_cp_prefill` causing failure during inference with `DtException: No matching compiler iter found`

Also found that having `@pytest.mark.cpu` on all tests in `test_spyre_cp_scheduler_steps.py` is incorrect (the mark is automatically applied from the backend parameterization).

### DIFF
--- a/tests/e2e/test_spyre_cp_scheduler_steps.py
+++ b/tests/e2e/test_spyre_cp_scheduler_steps.py
@@ -11,7 +11,6 @@ from scheduling_utils import check_scheduler_inference_steps
 from spyre_util import ModelInfo
 
 
-@pytest.mark.cpu
 @pytest.mark.chunked_prefill
 @pytest.mark.full_model
 # These values are all parameterized for test sorting
@@ -152,7 +151,6 @@ def test_prefill_tkv_too_big(model: ModelInfo, backend: str,
     )
 
 
-@pytest.mark.cpu
 @pytest.mark.chunked_prefill
 @pytest.mark.full_model
 @pytest.mark.parametrize("max_num_seqs", [2])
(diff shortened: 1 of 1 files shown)


tests/e2e/test_spyre_cp_scheduler_steps.py
```diff
@@ -429,6 +429,13 @@ def test_single_cp_prefill(model: ModelInfo, backend: str,
         * number of prompts: 1
             * 0: len = 512, max tokens = 1, step joining = 0
     """
+    # max_model_len=514 tests an edge case in the scheduler, but does not work
```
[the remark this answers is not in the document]
tjohnson31415: I was wondering something like that as well, but values like 576, 640, 768 also didn't work in my testing.

tests/spyre_util.py
```diff
@@ -40,6 +41,10 @@ def patch_environment(use_cb: bool,
     monkeypatch.setenv("VLLM_SPYRE_DYNAMO_BACKEND", backend)
     monkeypatch.setenv("VLLM_SPYRE_USE_CHUNKED_PREFILL",
                        "1" if use_chunked_prefill else "0")
+    # NB: setting this env var explicitly is needed to set the desired value for
+    # the chunk size in the case that granite 8b TP4 is detected
```
joerunde: In that case wouldn't we also need to re-override the internal config for `max_num_batched_tokens` as well?

joerunde: very hot 🌶️

PR #591, opened 2025-12-08
[PC][test] add additional test cases

yannicks1:
additional scenarios:

- [x] double prefix hit: three sequences in the decode batch share the same KV cache block.
- [x] only part of the prompt is identical (no full prefix) -> needs additional functionality in testing utils (see `prefix_lens`)
- [x] test prefix match of blocks that were filled identically during decode -> needs additional functionality in testing utils (see `n_lookaheads`)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
     # add labeled and unlabeled to the default types (runs when label is added)
     types: [opened, synchronize, reopened, labeled, unlabeled, auto_merge_enabled]
-    branches: [main, pc-additional-tests]
+    branches: [main]
 
   push:
     branches: [main]


tests/spyre_util.py
```diff
@@ -363,12 +363,16 @@ def create_seq_prompt(model: ModelInfo, token_length: int) -> str:
...
-                          seed: int = None) -> Request:
+def create_random_request(
+    request_id: int,
+    num_tokens: int,
+    sampling_params: SamplingParams,
+    from_model_vocab: bool = False,
+    model: Optional[ModelInfo] = None,
+    seed: int = None,
+    prefix_len: int = None,
+    n_lookahead: int = None,
```
sducouedic: A small comment / docstring would be useful to explain these two parameters, this functions starts to become complex

tests/spyre_util.py
```diff
@@ -383,6 +387,24 @@ def create_random_request(request_id: int,
...
+
+        if prefix_len is not None and prefix_len < num_tokens:
+            token_id_to_replace = prompt_token_ids[prefix_len]
+            # resample a token to diverge the prompt after prefix_len
+            while True:
+                token_id_replacement = random.choices(valid_token_ids, k=1)[0]
+                if token_id_replacement != token_id_to_replace:
+                    prompt_token_ids[prefix_len] = token_id_replacement
+                    break
+        if n_lookahead is not None and n_lookahead > 0:
```
sducouedic: I think we need an assert here that `prefix_len` is also not None and > 0 to avoid error in line 407

tests/spyre_util.py
```diff
@@ -383,6 +387,24 @@ def create_random_request(request_id: int,
...
+            # import here to prevent circular import
+            from output_util import generate_hf_output
+            generated_token_ids = generate_hf_output(
+                model=model,
+                prompts=[prompt_token_ids[0:prefix_len]],
+                max_new_tokens=n_lookahead)[0]["token_ids"]
+            for i in range(n_lookahead):
+                prompt_token_ids[prefix_len + i] = generated_token_ids[i]
```
sducouedic: if `n_lookahead + prefix_len > num_tokens`, we get an error. We need another assert

tests/spyre_util.py
sducouedic: oh you checked both of those above, good job

sducouedic (approved):
I really like the simple changes for partial prefix caching and decode prefix caching. I added suggestions of changes, but those are nits

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -146,6 +146,170 @@ def test_prefix_hit_within_batch(model: ModelInfo, backend: str,
...
+            * 0: len = 126,  max tokens = 68, step joining = 0
+            * 1: len = 193, max tokens = 2, step joining = 67
+    """
+    monkeypatch.setenv("VLLM_SPYRE_CP_INTERLEAVE_STEPS", "0")
+
+    seqs_max_tokens = [68, 2]
+    prompts_lengths = [126, 193]
+    steps_add_reqs = [0, 67]
+    seeds = [0, 0]  # twice the same sequence
+    prefix_lens = [126, 126]
```
joerunde: "prefix_lens" here is quite confusing, 126 is the entire length of sequence 0?

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -465,6 +629,179 @@ def test_limit_blocks_no_prefix_hit(model: ModelInfo, backend: str,
...
+                                        monkeypatch: pytest.MonkeyPatch,
+                                        set_random_seed, max_num_seqs: int,
+                                        max_model_len: int,
+                                        max_num_batched_tokens: int,
+                                        available_blocks: int):
+    """ Scenario where three equal and one different sequences are scheduled.
+    While prefilling the second and fourth sequence we have a prefix cache
+    hit and can reuse the first chunk. Note that the fetched prefix blocks 
+    are still part of the existing decode batch. Hence we have duplicated 
+    blocks in the block table for this example. More specifically, three
```
joerunde: >Note that the fetched prefix blocks 
    are still part of the existing decode batch. Hence we have duplicated 
    blocks in the block table for this example.
    
I'm a bit confused here, so we're identifying prefix cache hits but then we can't actually re-use the kv cache blocks at all, because the original sequence that filled those blocks is still running? So we're not actually caching anything in this case and we're recomputing?

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -146,6 +146,170 @@ def test_prefix_hit_within_batch(model: ModelInfo, backend: str,
...
+        monkeypatch=monkeypatch,
+        seqs_max_tokens=seqs_max_tokens,
+        prompts_lengths=prompts_lengths,
+        steps_add_reqs=steps_add_reqs,
+        checked_steps=checked_steps,
+        max_num_seqs=max_num_seqs,
+        max_model_len=max_model_len,
+        available_blocks=available_blocks,
+        use_cb=False,
+        random_prompts=True,
```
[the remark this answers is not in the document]
yannicks1: makes sense, its gotten really ugly.

the question is, do we want to merge these to have the test in place and do the refactoring as a follow up, or is it that ugly that we want to refactor first? 
Reason I am asking is that I wont get to this until next Tuesday. If someone has capacity to address it feel free to do it.

tests/spyre_util.py
```diff
@@ -383,6 +387,24 @@ def create_random_request(request_id: int,
...
+                if token_id_replacement != token_id_to_replace:
+                    prompt_token_ids[prefix_len] = token_id_replacement
+                    break
+        if n_lookahead is not None and n_lookahead > 0:
+            # import here to prevent circular import
+            from output_util import generate_hf_output
+            generated_token_ids = generate_hf_output(
+                model=model,
+                prompts=[prompt_token_ids[0:prefix_len]],
+                max_new_tokens=n_lookahead)[0]["token_ids"]
```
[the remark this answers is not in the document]
yannicks1: yes that is the intent. We want to check that we also match kv cache blocks that were filled during decodes, not just prefills. 

To test this I need an oracle that tells me the continuation of a sequence that the model will produce and opted for hf.

if it is numerically to unstable, we can:
- make this a soft fail
- only execute this test on cpu (it is working with the current configuration)
- or just don't test this case. 

If you see another option please let me know.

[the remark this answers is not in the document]
joerunde: See #606 

We do use GTI everywhere anyway, so yeah it seems like it's actually not an issue to use the hf outputs here. The refactored changes there make all of this much more explicit though

joerunde (approved):
ship it!

PR #595, opened 2025-12-09
bump: aftu to v0.5.0

tjohnson31415:


joerunde (commented):
Isn't this required to properly test chunked prefill? We might want to update the lower bound in pyproject.toml too

PR #597, opened 2025-12-09
Revert ":fire: Limit CI usage"

joerunde:
Reverts vllm-project/vllm-spyre#596

PR #598, opened 2025-12-09
:bug: try again

joerunde:
# Description

Instead of https://github.com/vllm-project/vllm-spyre/pull/597, this might fix the failure on main. This wraps an `||` so that we don't short-circuit to `true` and instead print some json


tjohnson31415 (approved):
🤞

PR #603, opened 2025-12-10
Deduplicate prefill

maxdebayser:
# Description

When we load blocks from prefix caching, there are two situations where we need to recompute blocks that are present in cache:
1) The sequences of blocks loaded does not entirely fill a number of blocks
2) The rightmost blocks are part of the last chunk of the prompt and must be recomputed regardless of filling this chunk entirely or not

This means that we create duplicates in the block table which unnecessarily reduce the number of available blocks. This PR "deduplicates" these blocks by actually running the recomputation on the padding block 0, by means of pointing the slot mappings of recomputed blocks to the address range of block 0, but keeping the block table as if we're loading from cahe.

This PR should also fix a corner case in which the loaded prefix fills the first K chunks completely, where K is not the last chunk of the prompt, and where this Kth chunk would be recomputed anyway. (TODO: add a test for that)

### DIFF
--- a/vllm_spyre/v1/worker/spyre_model_runner.py
+++ b/vllm_spyre/v1/worker/spyre_model_runner.py
@@ -2012,16 +2012,9 @@ class ChunkedPrefillModelRunner(ContinuousBatchingSpyreModelRunner):
 
         prompt_token_ids = request.prompt_token_ids
         prompt_len = len(prompt_token_ids)
-        if prompt_len <= chunk_size:
-            # Case I - Prompt fits in a single chunk
+        if num_computed_tokens == 0:
             chunk_start = 0
-            chunk_end = prompt_len
-            chunk_left_offset = left_padding
-        elif left_padding > 0 and num_computed_tokens == 0:
-            # Case II - First chunk, but it contains some padding blocks at
-            # the left
-            chunk_start = 0
-            chunk_end = chunk_size - left_padding
+            chunk_end = min(chunk_size - left_padding, prompt_len)
             chunk_left_offset = left_padding
         else:
             chunk_start = chunk_i * chunk_size - left_padding
--- a/tests/spyre_util.py
+++ b/tests/spyre_util.py
@@ -218,7 +218,7 @@ def get_spyre_model_dir_path() -> Path:
 
 # add pytest markers to supported different backends
 def get_spyre_backend_list():
-    backend_list = ["eager"]  #, "inductor", "sendnn"]
+    backend_list = ["eager", "inductor", "sendnn"]
 
     backends = []
     for backend in backend_list:


tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -12,6 +12,41 @@
 from spyre_util import ModelInfo
 
 
+def get_block_tables(engine_core):
```
yannicks1: Can you add types here for better readability: 
```
def get_block_tables(engine_core: EngineCore) -> Tuple[Dict[str, List[int]], Dict[str, int]]:
```

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -12,6 +12,41 @@
...
+    block_ref_count = {}
+
+    for blocks in req_to_blocks.values():
+        for block in blocks:
+            block_ref_count[block.block_id] = block.ref_cnt
+
+    return block_tables, block_ref_count
+
+
+def verify_block_tables(engine_core, step_num, step_ref, disable_asserts):
```
yannicks1: also typing would be nice, you did type the dummy function in scheduling_utils.py:

```
def verify_block_tables(engine_core: EngineCore, step_num: int,
                      step_ref: dict[str, Any], disable_asserts: bool) -> None:
```

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -765,7 +987,7 @@ def test_double_prefix_hit_within_batch(model: ModelInfo, backend: str,
             "running": ["1", "0"],
             "request_outputs": [],
             "n_reserved_blocks": 8,
-            "n_used_blocks": 5,
+            "n_used_blocks": 3,
```
yannicks1: it would be nice to add the block table /ref count assertion to this test. as it is testing the special case of a double prefix hit (ref count 3)

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -1180,8 +1411,14 @@ def test_multi_chunk_full_match(model: ModelInfo, backend: str,
...
-            "n_used_blocks": 10,
-            "n_cached_blocks": 4
+            "n_used_blocks": 8,
+            "n_cached_blocks": 4,
+            # when decode starts, we see the tables diverge
+            "block_tables": {
+                '0': [1, 2, 3, 4, 5, 6, 7],
+                '1': [1, 2, 3, 4, 5, 6, 8]
+            },
+            "block_ref_count": {1: 2, 2: 2, 3: 2, 4: 2, 5: 2, 6: 2, 7: 1, 8: 1}
```
yannicks1: multi line again

tests/scheduling_utils.py
```diff
@@ -388,16 +397,14 @@ def validate_scheduler_steps(
                     for r_id in req_ids2blocks
                 ]
                 n_cached_blocks = sum(cached_blocks)
-                for r_id in req_ids2blocks:
-                    print(f"{reqs[r_id].num_cached_tokens=}")
```
yannicks1: thanks, was about to rip that out in a separate commit

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -12,6 +12,41 @@
...
+def verify_block_tables(engine_core, step_num, step_ref, disable_asserts):
+
+    block_tables, block_ref_count = get_block_tables(engine_core)
+
+    if not disable_asserts:
+        if "block_tables" in step_ref:
+            assert step_ref["block_tables"] == block_tables
+
+        if "block_ref_count" in step_ref:
+            assert step_ref["block_ref_count"] == block_ref_count
```
yannicks1: isn't checking the ref count completely redundant if the block tables already match?

tests/e2e/test_spyre_pc_scheduler_steps.py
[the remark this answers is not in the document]
maxdebayser: Right, the step number is already part of the step_ref dict, so it's redundant.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1979,14 +1983,33 @@ def _prepare_chunked_prefill(self, req_id: str):
...
+                num_cached_block.get(req_id, 0)
+
+            total_blocks = left_padding_blocks + num_cached_blocks
+
+            # full match
+            if total_blocks == chunk_count * self.chunk_blocks_count:
+                blocks_to_recompute = self.chunk_blocks_count
+            else:
+                blocks_to_recompute = total_blocks % \
+                    self.chunk_blocks_count
```
yannicks1: This counts/de-duplicates only blocks in the last chunk. But we could also have blocks to be recomputed a non-last chunk.

maxdebayser: I think I've addressed the problem. The test case `test_multi_chunk_partial_match_misaligned` exercises this but there is not direct way to verify it in this test suite. In the model runner unit tests I hope to verify these things directly.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2180,43 +2214,52 @@ def _maybe_load_prefix_from_cache(self, scheduler_request: Request) -> int:
...
-            usable_blocks = (((left_blocks + n_hit) // self.chunk_blocks_count)\
-                * self.chunk_blocks_count) - left_blocks
-            usable_blocks = max(usable_blocks, 0)
-            logger.debug("Found: %d usable blocks in cache", usable_blocks)
-            computed_blocks = computed_blocks[:usable_blocks]
+            full_chunks_with_cached_blocks = ((left_blocks + n_hit) //
+                                              self.chunk_blocks_count)
+
+            # the last chunk of the prompt must always be
+            # recomputed
```
yannicks1: no newline needed

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1979,14 +1983,44 @@ def _prepare_chunked_prefill(self, req_id: str):
...
+
+            # When the current chunk has passed the number
+            # of chunk loaded entirely from cache, the difference
+            # between the blocks from cache and the allocated
+            # blocks will the the number of blocks to recompute.
+            # We could use more direct ways such as getting
+            # all blocks with a ref count > 1, but then it will
+            # make it more difficult to migrate to the engine core
+            # KV cache manager.
+            if chunk_i == chunks_from_cache:
```
yannicks1: nice logic here!

yannicks1 (commented):
LGTM !

tests/e2e/test_spyre_pc_scheduler_steps.py
yannicks1: oh, yes. good point!

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -124,8 +157,19 @@ def test_prefix_hit_within_batch(model: ModelInfo, backend: str,
...
+            "block_tables": {
+                '0': [1, 2, 3, 4],
+                '1': [1, 2, 3, 5]
+            },
+            "block_ref_count": {
+                1: 2,
+                2: 2,
+                3: 2,
+                4: 1,
+                5: 1
```
joerunde: Nice, this makes the "duplicated block" bit clear that it's the concurrent decoding of identical requests producing duplicate entries

tests/scheduling_utils.py
```diff
@@ -176,7 +181,8 @@ def check_scheduler_inference_steps(
     random_prompts: bool = False,
     prefix_caching: bool = False,
     seeds: list[int] = None,
-):
+    extra_assert_func: Callable[[EngineCore, dict[str, Any], bool],
+                                Any] = dummy_assert_func):
```
joerunde: it looks like we only use this to pass `verify_block_tables`. Shouldn't we always be validating block tables if one is given in a step ref, or is there a reason that we wouldn't want to?

tests/llm_cache.py
```diff
@@ -267,25 +269,10 @@ def get_engine(
...
-                # there is only one group.
-                kv_cache_group_id=0,
-                # We don't support DCP
-                # https://docs.vllm.ai/en/latest/serving/context_parallel_deployment/#decode-context-parallel
-                dcp_world_size=1,
-            )
+            worker.model_runner.block_pool = (
+                worker.model_runner._make_block_pool())
+            worker.model_runner.kv_cache_manager = (
+                worker.model_runner._make_kv_cache_manager())
```
joerunde: hell yeah 👍 

I wouldn't mind a `worker.model_runner.reset_kv_cache()` but this is waaaaay better already

tests/scheduling_utils.py
maxdebayser: Well, the idea was to move things out of the `check_scheduler_inference_steps()` function which was becoming way to big. (The motivation is similar to the refactoring you did to move the prompt generation out of the function). By passing an extra assert function, we can do test-specific verifications without polluting the function's code. 

Maybe validating block tables is generic enough to be applicable to other tests, but then I think we need another refactoring to make these verifications more modular.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2180,43 +2212,50 @@ def _maybe_load_prefix_from_cache(self, scheduler_request: Request) -> int:
...
+                "Found: %d reusable blocks in cache. "
+                "%d blocks will be (re)computed", usable_blocks,
+                blocks_to_compute)
             num_cached_tokens = usable_blocks * self.block_size
 
+            # Save all of the computed blocks and not only the usable
+            # ones because we will make a dummy recomputation of computed
+            # blocks in the last chunk to deduplicate the used blocks. So
+            # although we will recompute, we'll still point the block table
+            # to the cached blocks.
```
joerunde: Since we already know `computed_blocks`, `usable_blocks`, and `blocks_to_compute` here, could we save some of this info in the prefill batch for this request instead of requiring the logic to recalculate the number of recomputed blocks in the next pass?

joerunde (approved):
Took a bit to figure out how it all works but it looks 🌶️ !

If it is possible to save some of this info about the blocks that we're recomputing and not recompute the recomputed blocks (I heard you like recomputing 😉 )  later I think there's room for improving readability that way. Otherwise the tests look ✅ and this seems correct to me

maxdebayser:
Thanks for the reviews. Please don't merge yet as I think that I've caught a bug with the unit tests PR (unpushed yet).

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1979,14 +1983,42 @@ def _prepare_chunked_prefill(self, req_id: str):
...
+            # all blocks with a ref count > 1, but then it will
+            # make it more difficult to migrate to the engine core
+            # KV cache manager.
+            if chunk_i == chunks_from_cache:
+                num_cached_blocks = num_cached_tokens // self.block_size
+                # num_allocated_blocks will be > 0 after the first
+                # chunk is computed because as we compute, we commit
+                # to cache
+                num_allocated_blocks = self.kv_cache_manager.\
+                    num_cached_block.get(req_id, 0)
```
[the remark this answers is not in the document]
maxdebayser: Not exactly, let's say that there is no prefix hit, then initially num_allocated_blocks will be 0 but will increase chunk size tokens at each iteration.

vllm_spyre/v1/worker/spyre_model_runner.py
maxdebayser: That's not a bad idea. We can subclass the SamplingInputBatch to add these extra attributes and add builder method to the cb model runner that we can override in the cp model runner. What if we do this as a follow-up so that we can get the linter migration PR in?

PR #604, opened 2025-12-11
:arrow_up: Add support for vllm 0.12.0

joerunde:
# Description

So far it looks like the churn for 0.12.0 is actually not nearly as bad as 0.11.1. This bumps the upper bound and pins the lock to 0.12.0


vllm_spyre/compat_utils.py
```diff
@@ -1,5 +1,20 @@
...
 
 
 def dataclass_fields(cls: type) -> list[str]:
     return [f.name for f in fields(cls)]
+
+
+def has_argument(func: Callable, param_name: str) -> bool:
+    # Checks the signature of a method and returns true iff the method accepts
+    # a parameter named `$param_name`. This is slow and should not be used in
+    # the critical path.
```
tjohnson31415: Reading this gave me a thought: can we track this and check to make sure we don't put this in a critical path? Maybe we track the rate of calls to this function or something and WARN if it exceeds a threshold.

Or probably better: use an `@lru_cache` on this and other compatibility functions that use `inspect`.

.github/workflows/test.yml
```diff
@@ -82,7 +79,7 @@ jobs:
               repo: "git+https://github.com/vllm-project/vllm --tag v0.10.2"
             test_suite:
               name: "backward compat"
-              markers: "compat or (cpu and basic)"
+              markers: "compat or (cpu and basic) and not quantized and not static_batching"
```
tjohnson31415: Is this relying on operator precedence?
I presume we want all `compat` tests to run so we may want to be explicit:
```
compat or (cpu and basic and not quantized and not static_batching)"
```

joerunde:
bot:test
MARKERS="spyre and (cb or chunked_prefill) and not multi and not quantized"

PR #606, opened 2025-12-15
Refactor scheduler tests: split out prompt generation

joerunde:
# Description

Refactors for tkv step tests. This splits out the prompt creation from the engine validation, so that we can be very explicit about what prompts we are constructing for the tests.

IMHO, it makes the tests for these prefix edge cases more understandable, though it is a bit more verbose for the simple case where the prompts are just identical.

### DIFF
--- a/tests/e2e/test_spyre_pc_scheduler_steps.py
+++ b/tests/e2e/test_spyre_pc_scheduler_steps.py
@@ -39,12 +39,14 @@ def test_prefix_hit_within_batch(model: ModelInfo, backend: str,
     """
     monkeypatch.setenv("VLLM_SPYRE_CP_INTERLEAVE_STEPS", "0")
 
+    prompt = random_prompt(model=model, seed=0, length=192)
+
     request1 = create_request_for_scheduler_test(
         model=model,
         request_id=0,
         add_step=0,
         max_tokens=2,
-        prompt=random_prompt(model=model, seed=0, length=192),
+        prompt=prompt,
         use_golden_token_injection=True)
 
     request2 = create_request_for_scheduler_test(
@@ -52,7 +54,7 @@ def test_prefix_hit_within_batch(model: ModelInfo, backend: str,
         request_id=1,
         add_step=0,
         max_tokens=2,
-        prompt=random_prompt(model=model, seed=0, length=192),
+        prompt=prompt,
         use_golden_token_injection=True)
 
     checked_steps = [
@@ -172,8 +174,7 @@ def test_prefix_hit_decoded_block_within_batch(model: ModelInfo, backend: str,
     block is entirely prompt while the second block is a mix of prompt and
     decoded tokens. Note that the fetched prefix blocks are still part of the 
     existing decode batch. Hence we have duplicated blocks in the block table 
-    for this example. Also note that we cannot fetch a prefix and this test 
-    fails when we set n_lookaheads = [0, 0].
+    for this example.
 
     Configuration:
         * max_num_seqs: 2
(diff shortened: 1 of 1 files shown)


tests/scheduling_utils.py
```diff
@@ -38,6 +40,92 @@ def augment_checked_steps(
...
+        seed=seed,
+        num_tokens=length
+    ).prompt_token_ids
+
+def create_request_for_scheduler_test(
+    model: ModelInfo,
+    request_id: int,
+    add_step: int,
+    max_tokens: int,
+    prompt: list[int],
```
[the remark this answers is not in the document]
prashantgupta24: oh is that why you created the `random_prompt` function?

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -289,24 +289,36 @@ def test_prefix_hit_decoded_block_within_batch(
...
         monkeypatch=monkeypatch,
-        seqs_max_tokens=seqs_max_tokens,
-        prompts_lengths=prompts_lengths,
-        steps_add_reqs=steps_add_reqs,
+        requests=[request1, request2],
         checked_steps=checked_steps,
         max_num_seqs=max_num_seqs,
         max_model_len=max_model_len,
         available_blocks=available_blocks,
+        max_batch_tkv_limit=-1,
```
prashantgupta24: not sure about this

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -1161,14 +1229,27 @@ def test_multi_chunk_partial_match(model: ModelInfo, backend: str,
...
+        use_golden_token_injection=True)
+
+    request2 = create_request_for_scheduler_test(
+        model=model,
+        request_id=1,
+        add_step=0,
+        max_tokens=2,
+        prompt=request1.request.prompt_token_ids[0:254] +
+        random_prompt(model=model, seed=0, length=384 - 254),
+        use_golden_token_injection=True)
```
sducouedic: oh this is makes things actually so much simpler, no need to have the dedicated logic to create partial prefix prompt in the functions, just for a simple test. 

But I would put prompt 2 in a dedicated variable, for readability and highlighting that it is partial prefix prompt

```suggestion
    prompt2= request1.request.prompt_token_ids[0:254] +
        random_prompt(model=model, seed=0, length=384 - 254)
    request2 = create_request_for_scheduler_test(
        model=model,
        request_id=1,
        add_step=0,
        max_tokens=2,
        prompt=prompt2,
        use_golden_token_injection=True)
```

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -177,15 +183,24 @@ def test_prefix_hit_decoded_block_within_batch(
...
+        prompt=random_prompt(model=model, seed=0, length=126),
+        use_golden_token_injection=True)
+
+    request2 = create_request_for_scheduler_test(
+        model=model,
+        request_id=1,
+        add_step=67,
+        max_tokens=2,
+        prompt=request1.request.prompt_token_ids +
+        request1.hf_output["token_ids"][:2] +
```
sducouedic: here also I would put in `prompt2` variable

tests/e2e/test_spyre_pc_scheduler_steps.py
joerunde: Thanks @sducouedic! 

I went ahead and did this for all the test cases, they're all more readable now 🚀🚀🚀

tests/scheduling_utils.py
```diff
@@ -39,14 +42,81 @@ def augment_checked_steps(
...
+    # Can add golden token injection, which will ensure that the vllm output
+    # matches the hf output so that we can do logits comparisons on identical
+    # token outputs.
+
+    sampling_params = SamplingParams(max_tokens=max_tokens,
+                                     temperature=0.0,
+                                     logprobs=0,
+                                     ignore_eos=True)
+
+    hf_results = generate_hf_output(
```
yannicks1: very nice how this is afterwards used in the hf vllm comparison!

yannicks1 (approved):
This is amazing work! much more readable now. Thanks a lot @joerunde and @prashantgupta24 !

PR #608, opened 2025-12-16
[CP] Fix TKV constraint bug

sducouedic:
The tests here show the tkv and values after each step in the current implementation, third request gets scheduled when it shouldn't and the output is wrong (there is no inference error raised on cpu)

### DIFF
--- a/tests/e2e/test_logits_processors.py
+++ b/tests/e2e/test_logits_processors.py
@@ -41,19 +41,21 @@ def test_custom_logits_processor(model: ModelInfo, backend, monkeypatch,
             return logits
 
     patch_environment(
-        use_cb=mode in ["cb", "cp"],
+        use_cb=mode in ["cb", "cp", "pc"],
         warmup_shapes=warmup_shapes if mode == "sb" else None,
         backend=backend,
-        use_chunked_prefill=mode == "cp",
+        use_chunked_prefill=mode in ["cp", "pc"],
         monkeypatch=monkeypatch,
     )
 
-    spyre_model = LLM(model=model.name,
-                      revision=model.revision,
-                      max_model_len=max_model_len,
-                      max_num_seqs=max_num_seqs,
-                      max_num_batched_tokens=128 if mode == "cp" else None,
-                      logits_processors=[DummyLogitsProcessor])
+    spyre_model = LLM(
+        model=model.name,
+        revision=model.revision,
+        max_model_len=max_model_len,
+        max_num_seqs=max_num_seqs,
+        max_num_batched_tokens=128 if mode in ["cp", "pc"] else None,
+        enable_prefix_caching=mode == "pc",
+        logits_processors=[DummyLogitsProcessor])
     prompt = "Hello Logits Processors"
     params = SamplingParams(max_tokens=5, temperature=0, logprobs=0)
 


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2302,6 +2302,14 @@ def _maybe_prepare_last_prefill(self, req_id: str,
         if num_computed_tokens + num_scheduled_tokens < prompt_len:
             return
 
+        # Last prefill: we might need to update the tkv
+        req_n_blocks = len(self._get_blocks(req_id))
+        max_n_blocks = max(req_n_blocks,
+                           max(0, self.tkv - 1) // self.block_size)
```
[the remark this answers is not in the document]
joerunde: Also does this need to be `ceil((self.tkv - 1) / self.block_size)`? Because e.g. `24 // 64 == 0` but we'd want to count one block

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2302,6 +2302,14 @@ def _maybe_prepare_last_prefill(self, req_id: str,
...
+        n_left_padding = (max_n_blocks - req_n_blocks) * self.block_size
+        req_tkv = n_left_padding + prompt_len
```
joerunde: I'm on my first coffee so I might be wrong but I think there could be a bug here because it might not be true that it's the current prompt that has to get padded. Consider:

```
self.tkv = 60
prompt_len = 80
```

In this case, the sequence that's at 60 tokens will be padded, so our tkv should be 124. But I think the logic here will set it to 144. Do we need to account for all sequences in the batch here to determine the new correctly padded tkv?

vllm_spyre/v1/worker/spyre_model_runner.py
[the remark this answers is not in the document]
tjohnson31415: `self.tkv == 0` would be the case that we don't have a current decode batch?
Maybe it would be clearer to have an `if` condition for that case and just set `self.tkv = prompt_len`.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2302,6 +2302,14 @@ def _maybe_prepare_last_prefill(self, req_id: str,
         if num_computed_tokens + num_scheduled_tokens < prompt_len:
             return
 
+        # Last prefill: we might need to update the tkv
+        req_n_blocks = len(self._get_blocks(req_id))
+        max_n_blocks = max(req_n_blocks,
+                           max(0, self.tkv - 1) // self.block_size)
+        n_left_padding = (max_n_blocks - req_n_blocks) * self.block_size
```
maxdebayser: `self._get_blocks(req_id)` doesn't include padding blocks, so this expression won't correctly result in the number of padding blocks, I think.

tests/e2e/test_spyre_cp_scheduler_steps.py
```diff
@@ -151,6 +151,161 @@ def test_prefill_tkv_too_big(model: ModelInfo, backend: str,
...
+            "finished_requests": ["1"],
+            "n_reserved_blocks": 3,
+            "n_used_blocks": 3
+        },
+        {
+            # Prefill sequence 2
+            # Left padding due to long prompt 1 can be removed, tkv is shifted
+            # back to the first block
+            #
+            # TODO: This step is failing because the tkv is still 86 right now even after the long prompt request 1 finishes
```
sducouedic: @yannicks1 opinion here?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2302,6 +2302,16 @@ def _maybe_prepare_last_prefill(self, req_id: str,
...
             return
 
+        # Last prefill: we might need to update the tkv
+        req_n_blocks = len(self._get_blocks(req_id))
+        cur_n_blocks = math.ceil(self.tkv / self.block_size)
+        new_n_blocks = max(req_n_blocks, cur_n_blocks)
+        assert new_n_blocks > 0
+        base_n_tokens = (new_n_blocks - 1) * self.block_size
+        req_tkv_new_block = base_n_tokens + prompt_len % self.block_size
+        cur_tkv_new_block = base_n_tokens + self.tkv % self.block_size
```
yannicks1: this is treating the boundary case wrong when `prompt_len` is a multiple of the block size it should add 64 (block_size) but it adds 0.

tests/e2e/test_spyre_cp_scheduler_steps.py
sducouedic: thanks so much for investigating. 

> accept that we are suboptimal because the tkv our scheduler sees does not consider sequences that just left the batch in the last step.

this first solution makes sense to me, it is not dramatic to wait for one decode step to get the tkv updated. seems low priority to address this issue indeed. I'll update the tkv steps for that case

vllm_spyre/v1/worker/spyre_model_runner.py
sducouedic: oh good point

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -262,7 +262,8 @@ def test_prefix_hit_decoded_block_within_batch(model: ModelInfo, backend: str,
         {   # prefill chunk 1 seq 1
             # prefix hit of the two blocks in the first chunk
             "step": 68,
-            "tkv": 192, # why not 193
+            # use unchanged tkv of seq 0, since seq 1 is still prefilling
```
yannicks1: nit: as mentioned above, I would add comments like these to other non-last chunk prefills ...

yannicks1 (commented):
I think it looks reasonable @sducouedic ! Want to go over it one last time on Monday before approving as this touches the core of the code. 

If we can get some more eyes on it, that would be great @joerunde @tjohnson31415 @maxdebayser

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2316,6 +2316,18 @@ def _maybe_prepare_last_prefill(self, req_id: str,
         if num_computed_tokens + num_scheduled_tokens < prompt_len:
             return
 
+        # Last prefill: we might need to update the tkv
+        req_n_blocks = len(self._get_blocks(req_id))
```
yannicks1: is there scenarios where 
```
req_n_blocks = len(self._get_blocks(req_id))
```
is not equivalent to 
```
req_n_blocks = math.ceil(prompt_len / self.block_size)
```
?
if not so, it might be easier to understand...

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2316,6 +2316,18 @@ def _maybe_prepare_last_prefill(self, req_id: str,
...
+        cur_n_blocks = math.ceil(self.tkv / self.block_size)
+        new_n_blocks = max(req_n_blocks, cur_n_blocks)
+        assert new_n_blocks > 0
+        base_n_tokens = (new_n_blocks - 1) * self.block_size
+        req_tkv_new_block = base_n_tokens + (prompt_len -
+                                             1) % self.block_size + 1
+        cur_tkv_new_block = base_n_tokens + (self.tkv -
+                                             1) % self.block_size + 1
```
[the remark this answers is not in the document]
joerunde: I think this makes sense to me too. In my head I think of it like:
```python
tkv_modulo = max(self.tkv % 64, prompt_len % 64)
self.tkv = new_n_blocks * 64 + tkv_modulo
```
and this looks like what's happening here.

Maybe in the future we can refactor things a bit so that there's only one place where tkv is calculated. Currently it's a side effect of `_prepare_decode`, which we _could_ call after adding this request to `self.input_batch` a few lines down here, but that would be fairly awkward.

A future:
```python
self.input_batch.add_request(request)

self.tkv = self.input_batch.calculate_tkv()
```
would be nice!

joerunde (approved):
lgtm, 🤞 we can get tests passing

PR #607, opened 2025-12-15
Prefill when full decode batch

sducouedic:
This PR allows to start a new prefill when a decode batch is full. The prefill completes when there is room in the batch

PR #609, opened 2025-12-17
:alembic: try to correct the KV cache stats

joerunde:
# Description

This PR attempts to correct the KV cache stats. This shuttles kv cache metadata back from the model runner to the scheduler, so that the scheduler can edit the cache stats before they're sent to the statloggers. This corrects the stats in the logs and in the /metrics output.

It would be cool to also patch the logging statlogger so that it says "Spyer KV cache usage" instead:
```
(APIServer pid=8247) INFO 12-17 13:06:55 [v1/metrics/loggers.py:236] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 0.0 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 72.6%
```
but that might require plugging in our own entire logging stat logger :(.

This _does not_ address the issue that the prefix times will be skewed when a request is pre-chunked and then queues again waiting on space in the batch. That optimization is still WIP here: https://github.com/vllm-project/vllm-spyre/pull/607

This could use some polishing, I'm not sold on reusing the PrefixCacheStats in the model runner, and there may be a better way to update these stats than sending more data back from the model runner that I didn't find. Looking for some feedback before deciding if we should polish this up and merge or look for a different option.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
             repo: "git+https://github.com/vllm-project/vllm --branch main"
         test_suite:
           - name: "chunked prefill"
-            markers: "cpu and chunked_prefill and not quantized"
+            markers: "cpu and (chunked_prefill or prefix_caching) and not quantized"
           - name: "static batching"
             markers: "cpu and decoder and not cb and not other_e2e and not chunked_prefill and not prefix_caching and not quantized"
             flags: "--timeout=300"


maxdebayser:
Thanks, @joerunde , LGTM. If needed we add the support for kv events later.

tests/e2e/test_chunked_prefill.py
```diff
@@ -124,3 +125,61 @@ def wrapper(self, *args, **kwargs):
                     vllm_results=vllm_results,
                     hf_results=hf_outputs,
                     prompts=[prompt])
+
+
+# chunked prefill too? Or just prefix cachhing?
```
sducouedic: good question, if `mode` is set to `cp` only then we shouldn't have any cache hit right (`hit_tokens / total_tokens == 0.0`)? we could test that also to ensure it is the case

joerunde:
Thanks for the review @sducouedic!

The test for kv cache usage was a little awkward, but in the end not quite as bad as I expected. 🤞 that it works on GHA as well as on my machine

PR #613, opened 2025-12-22
:zap: Reduce idle prefills for better throughput

joerunde:
# Description

Currently prefix caching is implemented entirely in the model runner with no changes in the scheduler. This means that when we hit large prefixes in cache, the scheduler still thinks it has to schedule many prefill chunks that have already been loaded into cache, resulting in extra idle prefill iterations. This causes slow performance as those idle prefill iterations are interleaved with decode iterations, resulting in multiple seconds of extra queuing for requests in some cases.

This PR simply passes back the prefix cache hit length to the scheduler on the first chunk of a prefill so that the scheduler can account for the cache hit and advance the request to the next chunk that requires prefilling. This change is reflected in the existing multi-chunk prefix match test case.

Benchmarks show an appreciable speedup. Using:
```
vllm bench serve --dataset-name prefix_repetition --model ibm-granite/granite-3.3-8b-instruct --num-prompts 64 --max-concurrency 32 --prefix-repetition-prefix-len 8192 --percentile-metrics="ttft,tpot,itl,e2el" --metric-percentiles="95,99" 
```

We can see multi-second speedups in TTFT and end-to-end latency ⚡⚡⚡

### DIFF
--- a/vllm_spyre/v1/core/scheduler.py
+++ b/vllm_spyre/v1/core/scheduler.py
@@ -455,16 +455,14 @@ class ChunkedPrefillSpyreScheduler(ContinuousBatchingSpyreScheduler):
             is_first_chunk = req.num_computed_tokens <= self.chunk_size
             is_last_chunk = req.num_computed_tokens == req.num_prompt_tokens
             if is_first_chunk and not is_last_chunk:
-
-                left_padding = model_runner_output.left_padding.get(
-                    req.request_id, 0)
-                prefix_cache_len = model_runner_output.prefix_cache_hit_len.get(
-                    req.request_id, 0)
+                left_padding = model_runner_output.left_padding.get(req.request_id, 0)
+                prefix_cache_len = model_runner_output.prefix_cache_hit_len.get(req.request_id, 0)
 
                 req.num_computed_tokens = self.adjust_computed_tokens(
                     computed_tokens=req.num_computed_tokens,
                     left_padding=left_padding,
-                    prefix_cache_len=prefix_cache_len)
+                    prefix_cache_len=prefix_cache_len,
+                )
 
         # Remove completed prefills
         self.ongoing_prefills = [
@@ -484,8 +482,9 @@ class ChunkedPrefillSpyreScheduler(ContinuousBatchingSpyreScheduler):
 
         return super().update_from_output(scheduler_output, model_runner_output)
 
-    def adjust_computed_tokens(self, computed_tokens: int, left_padding: int,
-                               prefix_cache_len: int) -> int:
+    def adjust_computed_tokens(
+        self, computed_tokens: int, left_padding: int, prefix_cache_len: int
+    ) -> int:
         """
         Returns an adjusted `num_computed_tokens` given left padding and prefix
         cache hit info.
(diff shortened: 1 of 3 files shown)


maxdebayser:
Yeah, sorry. Now you get the same information but in blocks in the  `usable_blocks` attribute

vllm_spyre/v1/core/scheduler.py
```diff
@@ -454,19 +454,23 @@ def update_from_output(self, scheduler_output, model_runner_output):
             "Expecting an instance of CPSpyreModelRunnerOutput when doing "
             "chunked prefill.")
 
-        # Update the correct num_computed_tokens value given left-padding info
+        # Update the correct num_computed_tokens value given left-padding and
+        # prefix cache hit info
         for req in self.ongoing_prefills:
-            if req.request_id not in model_runner_output.left_padding:
-                continue
```
yannicks1: did we ever run into this if statement before your changes? It seems like it was never necessary? If we hit it before, you now potentially update `req.num_computed_tokens` for a req that was left untouched before. Does that sound right?

vllm_spyre/v1/core/scheduler.py
```diff
@@ -488,6 +492,21 @@ def update_from_output(self, scheduler_output, model_runner_output):
...
+        cache hit info.
+        """
+        # The prefix cache length is already adjusted for left padding.
+        # If it's bigger than the number of computed tokens, then we hit more
+        # prefix cache than we scheduled.
+        if prefix_cache_len > computed_tokens:
+            assert (prefix_cache_len + left_padding) % self.chunk_size == 0
+            return prefix_cache_len
+        # Otherwise just account for the left padding
+        return computed_tokens - left_padding
```
yannicks1: like this logic!

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2594,6 +2600,29 @@ def execute_model(
...
         return self.kv_cache_manager.block_pool.get_usage()
 
+    def get_prefix_cache_len(self,
+                             batch: SamplingInputBatch) -> dict[str, int]:
+        """Get the prefix cache hit length for a prefilling request.
+        This is in the number of usable cache tokens: Including the left padding
+        this will always land at a chunk boundary.
+        """
+        if batch != self.prefill_batch:
+            return {}
```
yannicks1: why is that needed? you only call the function once with batch=self.prefill_batch ...

vllm_spyre/v1/core/scheduler.py
joerunde: IIUC, this guard was stopping us from accessing `model_runner_output.left_padding[req.request_id]` if it wasn't in the dict. Using `model_runner_output.left_padding.get(req.request_id, 0)` with a default value that does nothing is behaviorally the same, but with fewer if statements. This makes it much cleaner to add in the logic to also handle the prefix cache length in the same way

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2613,6 +2619,29 @@ def execute_model(
...
+        req_id = batch.requests_ids[0]
+        request = self.requests[req_id]
+        assert isinstance(request, ChunkedPrefillRequestState)
+
+        # Prefix cache hits happen on first prefill iteration
+        if request.num_computed_tokens == 0:
+            num_cached_tokens = request.usable_blocks * self.block_size
+            return {req_id: num_cached_tokens}
+
+        return {}
```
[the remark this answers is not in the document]
joerunde: True, the future proofing with the for loop is great too. I think we can also just use `self.prefill_batch` directly to keep things safe without that check and to force this to be updated if we ever refactor away `self.prefill_batch`

maxdebayser (approved):
LGTM!

PR #614, opened 2025-12-23
Multimodal / Granite Vision Support

alex-jw-brooks:
# Description

Adds support for multimodal models, starting with granite vision as a first instance of the pattern. Current state:

- Have been able to run inference on AIU with multimodal with
     - [X] continuous batching
     - [x] continuous batching w/ chunked prefill
     - [ ] ~~static batching (punting this one if needed since it's deprecated anyway)~~
- ~~_Lots_ of things are hardcoded for llava next / granite vision. Unfortunately there are quite a lot of hacks to handle some weak interfaces in FMS, would be nice to have cleaner types in the future.~~ Added a layer of indirection around most of the model specific stuff for now at least, so the llava next parts are relatively contained.
- Multimodal preprocessing is still being done through vLLM, which is ideal, but good to be aware of for the future in case we need other models running on AIU / vLLM Spyre (i.e., to get granite vision embedding running, it would be best to contribute the model to vLLM upstream first and then add support here).

Next steps:
- [x] Still need to confirm parity against CPU
    - For the example script, I've written it to compare against either transformers or FMS. It looks like things do match up in my quick checks when running with eager backend and comparing it against the FMS CPU implementation, so I think things should be pushed through correctly here, and misalignments against transformers can hopefully be attributed to differences in FMS (I'm guessing from layer norms in the vision encoder).
- [x] abstract a bunch of stuff
- [x] add working example
- [x] add a lot of tests
    - [x] unit tests for llava next utils
    - [x] end to end tests for multimodal overall - exists, but for now is comparing against FMS instead of HF :(
- [X] add some short docs for how to add support mm models in the future (if needed)

### DIFF
--- a/tests/multimodal/test_llava_next.py
+++ b/tests/multimodal/test_llava_next.py
@@ -8,8 +8,8 @@ import copy
 
 import pytest
 import torch
-from fms.models import get_model
-from fms.utils import serialization
+from fms.models.llava_next import LlavaNextConfig
+from fms.models.hf.config_utils.param_builders import build_llava_next_params
 from PIL import Image
 from transformers import AutoConfig, AutoProcessor
 from vllm.multimodal.inputs import MultiModalFeatureSpec
@@ -34,24 +34,11 @@ def hf_config():
 
 
 @pytest.fixture(scope="module")
-def fms_config():
+def fms_config(hf_config):
     """Get the FMS config corresponding to the above."""
-    # Patch the head dimension directly (needed for llava next).
-    serialization.extend_adapter("llava_next", "hf", ["weight_expansion_for_mismatched_head_dim"])
-
-    config_dict = {"head_dim": 128}
-
-    # Just grab the model and take the config off of it. This is pretty fast,
-    # and mostly done for compatibility for now, since the config mapping utils
-    # in FMS are currently being heavily refactored.
-    model = get_model(
-        "hf_pretrained",
-        GVISION_MODEL.name,
-        fused_weights=False,
-        override_hf_pretrained_config=True,
-        text_config=config_dict,
-    )
-    return model.config
+    config_params = build_llava_next_params(hf_config)
+    config_params["text_config"].head_dim = 128
+    return LlavaNextConfig(**config_params)
 
 
 @pytest.fixture(scope="module")
@@ -83,8 +70,6 @@ def test_config_validation(fms_config, hf_config):
 
 
 ### Tests for inspecting the correctness of the warmup shapes
-
-
 def test_warmup_embed_types_and_shape(llava_next_mm_utils):
     """Ensure that the types and dimensions for the embeddings are consistent
     with the input IDs. Note that currently we pass input IDs all the way


docs/contributing/multimodal/adding_new_models.md
```diff
@@ -0,0 +1,71 @@
+# Multimodal Models in vLLM Spyre
+
+In order to understand how to get multimodal models running through vLLM Spyre, it is important to understand the differences between how models are implemented in vLLM & vLLM Spyre. To illustrate this, we use `llava_next` as an example, because `granite vision` is the only multimodal model currently supported.
```
joerunde: it isn't immediately obvious to the uninitiated that granite-vision is a llava-next model

vllm_spyre/multimodal/mm_mappings/llava_next.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # NOTE: use_cache is actually not used here, but currently it's
+        # required. Also, the value of iteration for decode as long as it's > 0.
+        input_embeds, _ = fms_model.prepare_inputs_for_generation(
+            iteration=0 if not is_decode else 1,
+            input_ids=input_ids,
+            kwargs=fms_kwargs)
+        return input_embeds
+
+    def get_warmup_mm_features(self):
```
joerunde: Is it possible to re-use the same vllm code path for creating mm_features out of multimodal data instead of requiring hand-rolled warmup mm_features per model type?

It seems like there's room for bugs here if the features aren't built correctly, and the possibility of code drift between how features are created here vs. at inference time

vllm_spyre/multimodal/mm_mappings/llava_next.py
```diff
@@ -0,0 +1,158 @@
...
+
+                # Careful about this; if it's 1D, we'll a tensor of shape
+                # [x, y], which will break in a weird way in image packing,
+                # since it assumes it's 2D and will get sad about getting
+                # an int instead of an iterable
+                if image_sizes.ndim == 1:
+                    image_sizes = image_sizes.unsqueeze(0)
+                fms_kwargs["image_sizes"] = image_sizes
+
+        # NOTE: use_cache is actually not used here, but currently it's
```
joerunde: What does `not used here` mean?

I see a code path that uses it: https://github.com/foundation-model-stack/foundation-model-stack/blob/v1.5.0/fms/models/llava_next.py#L392-L394

Or do you mean that there's no caching actually happening?

vllm_spyre/multimodal/mm_mappings/llava_next.py
```diff
@@ -0,0 +1,158 @@
...
+        mm_features: list[MultiModalFeatureSpec],
+        is_decode: bool,
+    ) -> torch.Tensor:
+        """Get the text or multimodal embeddings for Llava Next using
+        the (potentially compiled) FMS model.
+        """
+        fms_kwargs = {"use_cache": True}
+        mm_spec_keys = ["pixel_values", "image_sizes"]
+
+        # Only merge multimodal features in prefill; nothing mm in decode
```
joerunde: Should this comment have an assertion tied to it?

```
if is_decode:
    assert not mm_features
```

Or does it just mean that we pass `use_cache=True` so that `fms_model.prepare_inputs_for_generation` doesn't merge them in decode?

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -329,6 +345,11 @@ class SpyreModelRunner(BaseSpyreModelRunner[SamplingInputBatch,
 
     def __init__(self, vllm_config: VllmConfig, is_driver_worker: bool,
                  rank: int):
+        # Normally we would check if the modelconfig supports multimodal inputs
+        # by comparing the model config against the registry here
+
+        # We also calculate the multimodal budget; this is probably important
+        # see scratch/projects/vllm/vllm/v1/worker/utils.py (in the budget)
```
joerunde: is this a file path on your workspace?

Could link out to a specific tag of vllm instead

docs/contributing/multimodal/adding_new_models.md
alex-jw-brooks: Yeah, I will add a point clarifying this. In case someone finds this comment later - architecturally, the only difference between normal llava next and granite vision is that:

- llava next - you throw your images through the visual encoder, take the result of one layer, and use that as image features

- granite vision - same thing, but you take the result of multiple layers and concatenate them to form the image features

since it's such a small change, it's implemented in everything (transformers, vllm, llama cpp, etc) as an instance of llava next

vllm_spyre/v1/worker/spyre_model_runner.py
alex-jw-brooks: This is actually a stale comment and was mostly relevant to static batching I think - I'll remove it!

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -60,6 +62,8 @@ def new_request_data_builder(
     }
 
     ## Temporary backwards compatibility for 0.10.2
+    # NOTE: currently multimodal only works on 0.11.0+
+    # If this is expected to be around a while, add bc
```
joerunde: If MM is only 0.11.0+ then we can rip out 0.10.2, we're not planning on any long-term support for older vllm versions atm. (We can do this in a separate PR)

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -163,6 +171,12 @@ def compile_or_warm_up_model(self) -> None:
                 "[WARMUP] (%d/%d) for prompt length %d, decoding %d tokens "
                 "with batch size %d...", i + 1, num_shape_combinations,
                 prompt_len, num_decode_tokens, batch_size)
+
+            if self.model_runner.is_multimodal():
+                logger.warning(
+                    "[WARMUP] - this model is multimodal, but mm features are "
+                    "not passed through in static batching at the moment; "
+                    "this will probably break compiled models!")
```
joerunde: I'd just raise a value error here since it's not supported

vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -442,6 +459,7 @@ def load_model(self):
             # unused for continuous batching: set here to use same API
             wup_prompt_lens, wup_new_tokens = (0, ), (0, )
         else:
+            # TODO - handle warmup for multimodal
```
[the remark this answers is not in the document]
alex-jw-brooks: 👍 good point - changed to just raise NotImplementedError in the case that we don't have CB enabled and the model is multimodal, since it's more clear (after load since we don't know until after it's loaded for now)

vllm_spyre/multimodal/mm_mappings/llava_next.py
alex-jw-brooks: Added an assertion - since MM features are computed and merged and in prefill, we should not keep passing them forward, both because we won't have placeholders to merge them into since they are only in the original prompt, and because we don't want to accidentally run redundant computations anyway.

In FMS, `prepare_inputs_for_generation` actually runs the encoder and merges; this is different in other libs, where it usually happens on the first forward()

tests/hf_result_cache.py
```diff
@@ -38,6 +41,17 @@ def __init__(self):
...
+    @staticmethod
+    def mm_data_encode(obj):
+        """Fallback for JSON serialization for cache keys from MM inputs."""
+        # Convert any multimodal data types to torch tensors and hash
+        if isinstance(obj, Image):
+            return HFResultCache.mm_data_encode(torch.as_tensor(np.array(obj)))
+
+        if isinstance(obj, torch.Tensor):
+            return hash(tuple(obj.reshape(-1).tolist()))
+        raise TypeError(f"Object of type {type(obj)} is not serializable!")
```
alex-jw-brooks: This util is currently unused atm, but will be needed when we compare against Transformers instead of FMS for future vision models, so leaving it for now.

.github/workflows/test.yml
```diff
@@ -66,14 +66,20 @@ jobs:
...
             flags: "--timeout=300"
           - name: "other e2e"
             markers: "cpu and other_e2e and not quantized"
             flags: "--timeout=300"
+          - name: "multimodal"
+            markers: "cpu and multimodal"
+            flags: "--timeout=300 -sv"
+            # hf_model: "ibm-granite/granite-vision-3.2-2b"
+            # hf_model_rev: "2818ae5b93cb750b099df1b65f7864e4a0401271"
+            env_overrides: "HF_HUB_OFFLINE=0"
```
alex-jw-brooks: E2E tests are skipped for now, so this is only downloading the processing utils from the model, which are very tiny and fast - we will reenable e2e tests once we enable an arch that has a smaller model available

PR #615, opened 2025-12-23
Unit tests for chunked prefill model runner with prefix caching

maxdebayser:
# Description

There are things where it is hard to verify that they are working correctly in the step tests. In the step tests we can only observe the model runner state between calls to execute_model. However, we can't directly test its methods with specific inputs and verify returned values and the subsequent state.

This PR adds a mock for the model so that we can run the model runner in isolation.

### DIFF
--- a/tests/v1/worker/test_prefix_caching_worker.py
+++ b/tests/v1/worker/test_prefix_caching_worker.py
@@ -11,7 +11,10 @@ def test_block_sharing_for_2_chunks(
     monkeypatch: pytest.MonkeyPatch,
 ):
     model = InstrumentedModelRunner.DEFAULT_TEST_MODEL
-    pc_model_runner = InstrumentedModelRunner.build(monkeypatch)
+    pc_model_runner = InstrumentedModelRunner.build(
+        monkeypatch,
+        max_num_batched_tokens=128,
+    )
     prompt = random_prompt(model=model, seed=0, length=192)
 
     request1 = create_request_for_scheduler_test(
@@ -83,6 +86,7 @@ def test_multi_chunk_partial_match_misaligned(
 
     pc_model_runner = InstrumentedModelRunner.build(
         monkeypatch=monkeypatch,
+        max_num_batched_tokens=128,
         available_blocks=16,
     )
 
@@ -118,7 +122,9 @@ def test_multi_chunk_partial_match_misaligned(
     )
 
     # Schedule chunk 0 of request 0
-    model_runner_output_1 = pc_model_runner.execute_new_request(request1.request, 128)
+    model_runner_output_1 = pc_model_runner.execute_new_request(
+        request=request1.request, tokens_to_schedule=128
+    )
     pc_model_runner.assert_block_tables_and_slot_mappings(
         block_tables=[[1, 2]],
         slot_mappings=[[1, 2]],
@@ -173,7 +179,9 @@ def test_multi_chunk_partial_match_misaligned(
     )
 
     # Schedule chunk 0 of request 1
-    model_runner_output_4 = pc_model_runner.execute_new_request(request2.request, 128)
+    model_runner_output_4 = pc_model_runner.execute_new_request(
+        request=request2.request, tokens_to_schedule=128
+    )
     # chunk loaded from cache
     assert pc_model_runner.model.last_attn_metadata is None
     pc_model_runner.verify_model_runner_output(
@@ -305,7 +313,9 @@ def test_first_chunk_recomputation(
     )
 
     # Schedule chunk 0 of request 0
-    model_runner_output_1 = pc_model_runner.execute_new_request(request1.request, 128)
+    model_runner_output_1 = pc_model_runner.execute_new_request(
+        request=request1.request, tokens_to_schedule=128
+    )
     pc_model_runner.assert_block_tables_and_slot_mappings(
         block_tables=[[0, 0, 1, 2]],
         slot_mappings=[[0, 0, 1, 2]],
@@ -320,7 +330,9 @@ def test_first_chunk_recomputation(
     )
 
     # Schedule chunk 0 of request 1
-    model_runner_output_2 = pc_model_runner.execute_new_request(request2.request, 128)
+    model_runner_output_2 = pc_model_runner.execute_new_request(
+        request=request2.request, tokens_to_schedule=128
+    )
     pc_model_runner.assert_block_tables_and_slot_mappings(
         block_tables=[[0, 0, 1, 3]],
         slot_mappings=[[0, 0, 0, 3]],
@@ -427,7 +439,9 @@ def test_middle_chunk_recomputation_with_padding(
     )
 
     # Schedule chunk 0 of request 0
-    model_runner_output_1 = pc_model_runner.execute_new_request(request1.request, 256)
+    model_runner_output_1 = pc_model_runner.execute_new_request(
+        request=request1.request, tokens_to_schedule=256
+    )
     pc_model_runner.assert_block_tables_and_slot_mappings(
         block_tables=[[1, 2, 3, 4]],
         slot_mappings=[[1, 2, 3, 4]],
@@ -463,8 +477,10 @@ def test_middle_chunk_recomputation_with_padding(
     )
 
     # Schedule chunk 0 of request 1
-    model_runner_output_3 = pc_model_runner.execute_new_request(request2.request, 256)
-    # chunk loaded from cache
+    model_runner_output_3 = pc_model_runner.execute_new_request(
+        request=request2.request, tokens_to_schedule=256
+    )
+    # chunk loaded from cache, therefore the attn_metadata is not set
     assert pc_model_runner.model.last_attn_metadata is None
 
     pc_model_runner.verify_model_runner_output(


tests/scheduling_utils.py
```diff
@@ -105,7 +111,8 @@ def create_request_for_scheduler_test(
                       arrival_time=0,
                       lora_request=None,
                       pooling_params=None,
-                      cache_salt=None)
+                      cache_salt=None,
+                      block_hasher=block_hasher)
```
joerunde: Why isn't this required to be set on the requests for tests using the real scheduler and model runner?

Or the other way around- why is it required here for the mock ones?

tests/v1/worker/test_prefix_caching_worker.py
```diff
@@ -0,0 +1,504 @@
...
+        self.last_input_ids = input_ids
+        self.last_positions = positions
+        self.last_masks = masks
+        self.last_is_prompt = is_prompt
+
+        forward_context = get_forward_context()
+
+        assert isinstance(forward_context.attn_metadata,
+                          SpyreAttentionMetadata)
+        self.last_attn_metadata = forward_context.attn_metadata
```
joerunde: If `get_forward_context` is all we need to call at each step to get the step's attention metadata, then I think we should be able to do that in the existing scheduler step tests. The tests are at the engine level, but turn off multiprocessing so that the model runner and model are in the same process so we should be able to fetch the real forward context and save a copy of the attn_metadata at each step

tests/v1/worker/test_prefix_caching_worker.py
```diff
@@ -0,0 +1,638 @@
...
+
+
+@pytest.mark.cpu
+@pytest.mark.worker
+@pytest.mark.prefix_caching
+@pytest.mark.parametrize("max_num_seqs", [2])
+@pytest.mark.parametrize("max_model_len", [512])
+@pytest.mark.parametrize("max_num_batched_tokens", [256])
+@pytest.mark.parametrize("available_blocks", [16])
+def test_first_chunk_recomputation(
```
maxdebayser: @joerunde , I've validated that when I comment out `blocks_to_recompute += request.padding_blocks` this test fails.

tests/v1/worker/test_prefix_caching_worker.py
```diff
@@ -0,0 +1,504 @@
...
+        model=model,
+        request_id=0,
+        add_step=0,
+        max_tokens=2,
+        prompt=prompt,
+        use_golden_token_injection=True,
+        block_hasher=pc_model_runner.request_block_hasher)
+
+    (chunk_count, left_blocks, usable_blocks,
+     n_hit) = pc_model_runner._plan_chunking(request1.request)
```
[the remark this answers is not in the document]
maxdebayser: I've implemented the dataclass suggestion to make the code cleaner. But I didn't extract the planning method because it needs to access state from the model runner.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2195,7 +2203,12 @@ def _plan_chunking(self, scheduler_request: Request) -> tuple[int, int, int, int
             usable_blocks = 0
             n_hit = 0
 
-        return chunk_count, left_blocks, usable_blocks, n_hit
+        return ChunkedPrefillPlan(
```
joerunde: 🌶️🌶️🌶️!

Much nicer than unnamed return tuples 👍 

I do think we could take this idea even further, though we should leave for a future PR since this one was aimed at getting more test coverage written. What I'm thinking is that we already know more things up-front, like exactly which blocks will be recomputed and need to be masked out at prefill time. It could be possible to encapsulate more of that logic in here so that it's both easy to test and not nested within the prefill loop in the model runner

tests/v1/worker/test_prefix_caching_worker.py
joerunde: Nice!

It would be great to also add test coverage for where this recomputation occurs in the third block or later, to capture the case when we have a long prefix hit and pick back up in the middle of a chunk. Something like:

```
 p1 = [ ABCD | EFGH ]
 p2 = [ 00AB | CDEF | GHIJ ]
```

where we would only call the model runner twice for `p2`, once for the initial prefill step to hit the cache and then once for the third chunk.

tests/v1/worker/test_prefix_caching_worker.py
```diff
@@ -0,0 +1,328 @@
...
+    pc_model_runner.assert_block_tables_and_slot_mappings(
+        block_tables=[[0, 0, 1, 2]],
+        slot_mappings=[[0, 0, 1, 2]],
+    )
+
+    assert model_runner_output_1.req_ids == ["0"]
+    assert len(model_runner_output_1.sampled_token_ids) == 1
+    assert model_runner_output_1.tkv == 128
+    assert model_runner_output_1.n_free_blocks == 12
+    assert model_runner_output_1.left_padding == {"0": 128}
```
joerunde: For all the returns from `execute_new_request` we should check the `prefix_cache_hit_len` dict as well to make sure that's correct

tests/hf_cache.json
```diff
@@ -579,6 +579,22 @@
                     "tokens": [ "cs", "ymbol" ],
                     "logprobs": [ -2.9405062198638916, -6.677685737609863 ]
                 }
+            },
+            "__tokens__41505_37255_20672_12727_25130_19903_38525_14909_23426_28674_44635_24806_13853_37149_30394_12313_44715_48305_39823_44343_15245_35872_44179_33619_23207_4950_21340_30026_44876_47510_23446_42531_12804_39568_26970_691_35375_19603_40542_32841_57_24260_42644_11989_15985_42785_9392_27894_11729_47556_39478_22019_3954_15732_24966_45850_5361_27096_34729_26908_40032_26556_47374_29648_28882_21872_29309_18919_28294_14270_9309_9178_30119_32276_23422_4415_37237_43095_45386_41408_44147_45371_26572_19233_34666_13548_39893_41754_43993_28990_46682_28493_22146_32452_48968_45069_38993_4049_30119_23910_30973_41537_11946_35954_5758_10836_39055_16345_40103_4945_7194_34292_2224_28207_44729_26257_33452_1313_31211_29803_28309_19229_18193_48194_1789_1064_47236_9092": {
```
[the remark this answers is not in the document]
maxdebayser: It's because the `create_request_for_scheduler_test` generates the hf results. But I've made that optional now.

tests/v1/worker/mock_model.py
```diff
@@ -0,0 +1,298 @@
...
+        self,
+        input_ids: torch.Tensor,
+        positions: torch.Tensor,
+        masks: torch.Tensor,
+        is_prompt: bool,
+    ) -> torch.Tensor:
+        self.last_input_ids = input_ids
+        self.last_positions = positions
+        self.last_masks = masks
+        self.last_is_prompt = is_prompt
```
yannicks1: as mentioned above: can be removed and tests still pass

tests/v1/worker/mock_model.py
```diff
@@ -0,0 +1,298 @@
...
+    def execute_model(
+        self,
+        scheduler_output: SchedulerOutput,
+        **kwargs,
+    ) -> ModelRunnerOutput:
+        self.model.last_input_ids = None
+        self.model.last_positions = None
+        self.model.last_masks = None
+        self.model.last_is_prompt = None
+        self.model.last_attn_metadata = None
```
yannicks1: as mentioned above: can be removed and tests still pass

tests/v1/worker/mock_model.py
```diff
@@ -0,0 +1,298 @@
...
+    def get_mask_dtype(self) -> torch.dtype:
+        return self._mask_dtype
+
+
+class InstrumentedModelRunner(ChunkedPrefillModelRunner):
+    ALL_SLICE = slice(None)
+    DEFAULT_TEST_MODEL = ModelInfo(
+        name="ibm-ai-platform/micro-g3.3-8b-instruct-1b",
+        revision="6e9c6465a9d7e5e9fa35004a29f0c90befa7d23f",
+    )
```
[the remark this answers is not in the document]
maxdebayser: It's just to group all these things in the same "namespace", which is the class name.

tests/v1/worker/mock_model.py
```diff
@@ -0,0 +1,298 @@
...
+
+        self.model = MockContinuousBatchingFmsModel()
+
+        self.vocab_size = vllm_config.model_config.get_vocab_size()
+
+        self.last_input_ids: torch.Tensor | None = None
+        self.last_positions: torch.Tensor | None = None
+        self.last_masks: torch.Tensor | None = None
+        self.last_is_prompt: bool | None = None
+        self.last_attn_metadata: SpyreAttentionMetadata | None = None
```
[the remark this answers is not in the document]
maxdebayser: This is just to have a clean type declaration of these attributes.

tests/v1/worker/mock_model.py
```diff
@@ -0,0 +1,298 @@
...
+        except (ImportError, ModuleNotFoundError):
+            self.sampler = Sampler()
+
+        # boolean tensor of length batch size with indices:
+        # True for unfinished sequences and
+        # False for finished or padded sequences
+        self.indices = None
+
+        # number of right pads (relevant for continuous batching only)
+        self.n_pads_right = 0
```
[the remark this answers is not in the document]
maxdebayser: It could be, but this test does not exercise enough functionality to be sure. For these specific lines, I can leave them out and if some future test requires them they can be added back.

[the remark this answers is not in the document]
maxdebayser: Ok, so these two are assigned to in the chunked prefill model runner in `_prepare_chunked_prefill` and `_prepare_decode`. Since there is only an assignment in the model runner code, the tests pass if you remove them, but in my opinion it's cleaner to have all class attributes that are assigned to declared in `__init__`.

tests/v1/worker/test_prefix_caching_worker.py
```diff
@@ -0,0 +1,485 @@
...
+        generate_hf_results=False,
+        block_hasher=pc_model_runner.request_block_hasher,
+    )
+
+    chunk_plan = pc_model_runner._plan_chunking(request1.request)
+
+    assert chunk_plan.chunk_count == 2
+    assert chunk_plan.padding_blocks == 1
+    assert chunk_plan.usable_cache_blocks == 0
+    assert chunk_plan.total_cache_blocks == 0
```
yannicks1: maybe an overkill for now, but if we add more of these tests asserting variables in `chunk_plan`, we might also want to refactor this out into a function similar as I suggested as `verify_model_runner_output`

tests/v1/worker/mock_model.py
```diff
@@ -0,0 +1,298 @@
...
+    def sample(
+        self,
+        logits: torch.Tensor,
+        sampling_metadata: SamplingMetadata,
+    ) -> SamplerOutput | None:
+        next_tokens = self.sampler(logits, sampling_metadata)
+        return next_tokens
+
+    def get_mask_dtype(self) -> torch.dtype:
+        return self._mask_dtype
```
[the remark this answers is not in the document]
maxdebayser: I've removed it.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1805,6 +1805,14 @@ def execute_model(
         return model_output
 
 
+@dataclass
+class ChunkedPrefillPlan:
```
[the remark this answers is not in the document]
yannicks1: yeah sure, lets keep it like that for now

tests/v1/worker/mock_model.py
yannicks1: `last_attn_metadata` is only asserted to be None, right? it is not used in any other way.. ? 

If we want all these variables
```
        self.model.last_input_ids = None
        self.model.last_positions = None
        self.model.last_masks = None
        self.model.last_is_prompt = None
        self.model.last_attn_metadata = None
```
already here for future test we should probably comment that somewhere. It was confusing me a lot as they are unused right now.

maxdebayser: No, it's also used in `assert_block_tables_and_slot_mappings`

tests/v1/worker/test_prefix_caching_worker.py
```diff
@@ -0,0 +1,485 @@
...
+        add_step=0,
+        max_tokens=2,
+        prompt=prompt2,
+        use_golden_token_injection=False,
+        generate_hf_results=False,
+        block_hasher=pc_model_runner.request_block_hasher,
+    )
+
+    # Schedule chunk 0 of request 0
+    model_runner_output_1 = pc_model_runner.execute_new_request(request1.request, 128)
```
yannicks1: nit: for readability add arg names here (and to other 5 occurrences in this file): 

```
model_runner_output_1 = pc_model_runner.execute_new_request(request=request1.request, tokens_to_schedule=128)
```

like you do when calling `execute_running_requests`

PR #616, opened 2026-01-02
:bug: fix critical prefix cache bug

joerunde:
# Description

This PR fixes a critical bug causing requests to crash the server when there is left padding and the prefix cache hit is less than one full chunk. In order to correctly write the test for this, I had to also fixup the llm caching to avoid `VLLM_DT_*` pollution between test runs.


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1929,7 +1929,7 @@ def _prepare_chunked_prefill(self, req_id: str):
 
         # last chunk
         blocks_to_recompute = 0
-        if request.total_hit_blocks > 0:
+        if request.usable_blocks > 0:
```
maxdebayser: Wait, I'm not sure if this is the correct solution. The "if" was testing for total_hit_blocks because even if the hit is shorter than a chunk, we want to deduplicate.

tests/llm_cache.py
```diff
@@ -284,6 +305,21 @@ def get_engine(
     def clear(self) -> None:
         self._cache.clear()
 
+    @staticmethod
+    def _wrap_model_forward(engine_core: EngineCore) -> None:
```
maxdebayser: Pretty clever! Maybe we don't need the model runner unit test PR after all.

yannicks1:
👀 at this now...

tests/llm_cache.py
joerunde: I did really like how the other PR tests the chunking plan separately though, so far we have very little test coverage of the things that the model runner uses

tests/spyre_util.py
```diff
@@ -508,10 +511,48 @@ def verify_block_tables(engine_core: EngineCore, step_ref: dict[str, Any], disab
...
+    # TODO: Need to map the keys (request ids) to correct order in tensor
+    # (But we only ever prefill one request right now)
+    for slot_mapping in prefill_slot_mappings.values():
+        slot_mapping_tensor = torch.arange(block_size, dtype=torch.int64).repeat(len(slot_mapping))
+        slot_mapping_tensor += (
+            torch.tensor(slot_mapping, dtype=torch.int64)
+            .repeat_interleave(block_size)
+            .mul_(block_size)
+        )
+        slot_mapping_tensor_list.append(slot_mapping_tensor)  # TODO: slice?
```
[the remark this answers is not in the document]
joerunde: Yes!

But also... I want this to pass the checks and merge instead of waiting another 2 hours in the GHA queue 😭

PR #617, opened 2026-01-05
📝  Update docs and scripts for new prek integration

joerunde:
# Description

This PR
- updates `format.sh` to run prek
- removes the now unused dependencies from the `lint` group
- updates the documentation wrt this change

The one quirk here is that `mypy` is still hanging out as the last thing left that isn't in our precommit config yet. I think we should quickly onboard to [ty](https://docs.astral.sh/ty/) as it's much faster and easier to set up than mypy, and get that into precommit so all of our linting and formatting is done in one place.

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -5,16 +5,16 @@
 # Cause the script to exit if a single command fails
 set -eo pipefail
 
-if [ -z "$@" ]; then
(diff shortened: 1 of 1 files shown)


maxdebayser (approved):
thanks!

PR #618, opened 2026-01-05
Fix compatibility issue when trust_remote_code==True

maxdebayser:
There is an vllm.utils import that only fails when `trust_remote_code=True` is set because the import is inside an `if`. The import still uses the pre utils refactoring path.


rafvasq (approved):
lgtm!

PR #619, opened 2026-01-06
Set default max_num_batched_tokens to 1k for CP

maxdebayser:
# Description

By setting a default in the flexible argument parser, the logic that sets it in vllm is overridden while still allowing the user to change the value on the command line.

### DIFF
--- a/tests/utils/test_cli_args.py
+++ b/tests/utils/test_cli_args.py
@@ -1,6 +1,5 @@
 import pytest
 from pydantic import ValidationError
-from huggingface_hub.errors import LocalEntryNotFoundError
 from vllm.entrypoints.openai.cli_args import make_arg_parser
 from vllm import EngineArgs
 
@@ -78,14 +77,11 @@ def test_generic_model_chunk_size_default(
         # Test that the upstream default is None but is changed to 2048 by
         # the VllmConfig initialization (when using CLI), but is then
         # overridden in our platform.py
-        try:
-            engine_args = _build_engine_args(
-                [
-                    *common_args,
-                ]
-            )
-        except LocalEntryNotFoundError:
-            pytest.skip("Skipping test of model not found in local cache")
+        engine_args = _build_engine_args(
+            [
+                *common_args,
+            ]
+        )
 
         assert engine_args.max_num_batched_tokens is None
         vllm_config = engine_args.create_engine_config()


vllm_spyre/platform.py
```diff
@@ -446,6 +449,8 @@ def _get_matching_warmup_shapes(
     def pre_register_and_update(cls, parser: FlexibleArgumentParser | None = None) -> None:
         if parser is not None:
             parser.set_defaults(enable_prefix_caching=False)
+            if envs_spyre.VLLM_SPYRE_USE_CHUNKED_PREFILL:
```
joerunde: Is it time to swap the line above as well and default prefix caching to true?

Or does that need to go with the 2.0 bump?

maxdebayser:
@joerunde , the tests are ready for review.

tests/utils/test_cli_args.py
```diff
@@ -0,0 +1,221 @@
...
+
+from vllm_spyre.platform import SpyrePlatform
+from spyre_util import environ_checkpoint, REFERENCE_MODELS
+
+try:
+    # old
+    from vllm.utils import FlexibleArgumentParser
+except ImportError:
+    # new
+    from vllm.utils.argparse_utils import FlexibleArgumentParser
```
tjohnson31415: We should add a little test to `test_upstream_compatibility.py` for this, as a reminder to remove it once we bump our lowest vllm version (i.e. like we have in [this change](https://github.com/vllm-project/vllm-spyre/pull/546/changes#diff-8bc36505735f045a39dd1e0b26c7d4ae8d9665b4df55a2f53713d342f7aff714R38-R46)).

If we are going to have a "new" vs "old" comment, we should include the version that changed it for reference.

tests/utils/test_cli_args.py
```diff
@@ -0,0 +1,221 @@
...
+        assert engine_args.max_num_batched_tokens == 64
+        vllm_config = engine_args.create_engine_config()
+        assert engine_args.max_num_batched_tokens == 64
+        if chunk_size == global_default:
+            assert vllm_config.scheduler_config.max_num_batched_tokens == 64
+        else:
+            assert vllm_config.scheduler_config.max_num_batched_tokens == chunk_size
+
+    with environ_checkpoint():
+        # Test that an invalid value will trigger an error
```
tjohnson31415:
```suggestion
        # Test that an invalid value that is not a multiple of 64 with CP will trigger an error
```

tests/utils/test_cli_args.py
```diff
@@ -0,0 +1,221 @@
...
+        # Test that the default is None but is changed to 2024 by
+        # the VllmConfig initialization
+        try:
+            engine_args = _build_engine_args(
+                [
+                    *common_args,
+                ]
+            )
+        except LocalEntryNotFoundError:
+            pytest.skip("Skipping test of model not found in local cache")
```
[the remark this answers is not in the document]
maxdebayser: Actually it looks like it can't be done, the failure wasn't in our code. In the construction of the EngineArgs and VllmConfig there are several places where the code tries to load the actual model configuration. Maybe it's enough if we can download just the `json` files from the repo.

tests/utils/test_cli_args.py
maxdebayser: Done

tests/utils/test_cli_args.py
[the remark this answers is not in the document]
tjohnson31415: Can we remove the `except LocalEntryNotFoundError` and `pytest.skip` now too?

PR #541, opened 2025-10-23
[chore] Update dev dependencies

ckadner:
# Description

- remove version constraints for upper bound on dev dependencies
- update `uv.lock` file

```
$ uv pip list --outdated

Using Python 3.12.11 environment at: ~/vllm-spyre/.venv

Package               Version   Latest     Type
--------------------- --------- ---------- -----
accelerate            1.9.0     1.11.0     wheel
aiohttp               3.12.15   3.13.1     wheel
aiu-fms-testing-utils 0.4.0     0.4.2      wheel
attrs                 25.3.0    25.4.0     wheel
blake3                1.0.6     1.0.8      wheel
cachetools            6.2.0     6.2.1      wheel
certifi               2025.8.3  2025.10.5  wheel
charset-normalizer    3.4.3     3.4.4      wheel
clang-format          18.1.5    21.1.2     wheel
codespell             2.3.0     2.4.1      wheel
compressed-tensors    0.11.0    0.12.2     wheel
datasets              4.0.0     4.2.0      wheel
depyf                 0.19.0    0.20.0     wheel
dill                  0.3.8     0.4.0      wheel
fastapi               0.117.1   0.119.1    wheel
fastapi-cli           0.0.13    0.0.14     wheel
fastapi-cloud-cli     0.2.1     0.3.1      wheel
filelock              3.19.1    3.20.0     wheel
frozenlist            1.7.0     1.8.0      wheel
fsspec                2025.3.0  2025.9.0   wheel
httptools             0.6.4     0.7.1      wheel
huggingface-hub       0.35.1    0.35.3     wheel
idna                  3.10      3.11       wheel
iniconfig             2.1.0     2.3.0      wheel
isort                 5.13.2    7.0.0      wheel
jiter                 0.11.0    0.11.1     wheel
lark                  1.2.2     1.3.0      wheel
llguidance            0.7.30    1.3.0      wheel
llmcompressor         0.7.1     0.8.1      wheel
llvmlite              0.44.0    0.45.1     wheel
markupsafe            3.0.2     3.0.3      wheel
mlx                   0.29.2    0.29.3     wheel
mlx-lm                0.28.0    0.28.3     wheel
mlx-metal             0.29.2    0.29.3     wheel
msgpack               1.1.1     1.1.2      wheel
multidict             6.6.4     6.7.0      wheel
multiprocess          0.70.16   0.70.18    wheel
numba                 0.61.2    0.62.1     wheel
numpy                 2.2.6     2.3.4      wheel
[...]

PR #623, opened 2026-01-08
🚀 Add vllm 0.13.0 support

joerunde:
# Description

Adds support for vllm 0.13.0, and upgrades all locked deps (we haven't done that in a while).

Dep upgrade done with `uv lock --upgrade`, see other chore PR here: #541


maxdebayser:
bot:test

maxdebayser (approved):

PR #627, opened 2026-01-14
Add support for mistral3 (no vision encoding)

alex-jw-brooks:
# Description
Supports running only the text part of multimodal mistral3 models through vLLM Spyre; I haven't been able to validate on AIU quite yet, but have at least ensured I can run it with the eager backend.

This PR is dependent on the following PR in FMS: https://github.com/foundation-model-stack/foundation-model-stack/pull/501

I will follow up in both FMS and here (after the granite vision PR is merged) to add support for pixtral in separate PRs & implement the multimodal utils.

### DIFF
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -350,6 +350,9 @@ class ContinuousBatchingFmsModel(FmsModelBase):
         elif self.config.model_type == "gpt_bigcode":
             self.kv_cache_specs["num_layers"] = self.config.n_layer
             self.kv_cache_specs["head_dim"] = self.config.n_embd // self.config.n_head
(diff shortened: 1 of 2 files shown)


vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -350,6 +350,9 @@ def __init__(
         elif self.config.model_type == "gpt_bigcode":
             self.kv_cache_specs["num_layers"] = self.config.n_layer
             self.kv_cache_specs["head_dim"] = self.config.n_embd // self.config.n_head
+        elif self.config.model_type == "pixtral":
```
gkumbhat: vllm resolves mistral3.1 and 3.2 as `pixtral`, so this if condition is accurate.

PR #628, opened 2026-01-15
Update logging around chunked prefill enablement

tjohnson31415:
Fixes and improvments around logging and configuration setting.

vllm-spyre currently configures chunked prefill based on the env var `VLLM_SPYRE_USE_CHUNKED_PREFILL`, but the `Initializing a V1 LLM engine` and `Chunked prefill is enabled` log messages were based on the value in `scheduler_config.chunked_prefill_enabled`.
- In vllm >= v0.11.1, setting the default value in the argparser is sufficient to fix the log messages
- In vllm < v0.11.1, the value of `scheduler_config.chunked_prefill_enabled` from the parser would be overwritten... so I can't easily fix the `Chunked prefill is enabled` message, but can inject the value from `VLLM_SPYRE_USE_CHUNKED_PREFILL` to fix the `Initializing a V1 LLM engine`  message.

Also updated the log of Spyre configurations in the "Overriding configurations based on warmup shapes" log. Removing the mention of "warmup shapes" and adding chunked prefill and prefix caching.

And added an config validation that raises an error if prefix caching is configured without chunked prefill to avoid confusion with that too.

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -150,7 +150,6 @@ class SpyrePlatform(Platform):
 
         # enable_prefix_caching will be defaulted to True when used with LLM();
         # only assert if our arg parser default was applied
-        print("Used with CLI", cls._used_with_cli)
         if cls._used_with_cli:
             assert (
                 cache_config.enable_prefix_caching and envs_spyre.VLLM_SPYRE_USE_CHUNKED_PREFILL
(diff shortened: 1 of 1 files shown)


vllm_spyre/platform.py
```diff
@@ -145,6 +148,16 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
         )
 
+        # enable_prefix_caching will be defaulted to True when used with LLM();
+        # only assert if our arg parser default was applied
+        print("Used with CLI", cls._used_with_cli)
+        if cls._used_with_cli:
+            assert (
+                cache_config.enable_prefix_caching and envs_spyre.VLLM_SPYRE_USE_CHUNKED_PREFILL
+            ) or not cache_config.enable_prefix_caching, (
+                "Cannot use prefix caching without chunked prefill."
```
joerunde: very 🌶️

joerunde (approved):
lpgtm, thanks @tjohnson31415!

PR #630, opened 2026-01-15
🐛 fix CP for FP8 model

prashantgupta24:
# Description

The one change needed for the `FP8` models is the `scale_indices` that we send back when we call `prepare_model_input`. The way it's set up in `chunked prefil`l is that we call
```
model_inputs = self._prepare_chunked_prefill(req_id)
self._maybe_prepare_last_prefill(req_id, scheduler_output)
```

But `_prepare_chunked_prefill` only has requests in the `prefill batch` and at the end of that function we try to send:

```
scale_indices=self.input_batch.request_indices
```
which results in the error we were seeing:
```
 assert len(attn_metadata.scale_indices) == 1
```
Ideally each request needs to have the same `scale_index` for `prefill` and `decode`. This PR moves the logic of calling `_maybe_prepare_last_prefill` into `_prepare_chunked_prefill` itself and changes `model_inputs` like we want it to.

Notes:

`Joe Runde`

- `set_past_key_value_states` is only used during model initialization / post-warmup to do initial kv cache - `setup_set_scale_for_fp8` is called before each model forward pass
- `_scaled` is the flag that tells fms-mo whether or not it should recompute the scales, where True means it should not recompute the scales because it's already scaled
- `_update_scale_for_fp8` is called post-forward pass to save the scale values that fms may have recomputed

### DIFF
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -551,6 +551,9 @@ class ContinuousBatchingFmsModel(FmsModelBase):
     def _set_scale_for_fp8(self, attn_metadata: SpyreAttentionMetadata):
         for layer_idx, (k, v) in enumerate(self.past_key_value_states):
             if attn_metadata.is_prefill:
+                # TODO: for chunked prefill, we don't need to be recreating the
+                # scale tensors every chunk. Possible future enhancement
+
                 # NOTE: Currently, prefill is only for a single prompt
                 # In prefill, we restore the scale (no scale) and
                 # reset to 1.
@@ -562,8 +565,8 @@ class ContinuousBatchingFmsModel(FmsModelBase):
                 v._scale = self.current_kv_scales[layer_idx][1][prefill_index] = torch.ones(
                     1, dtype=torch.float32
                 )
-                k._scaled = False
-                v._scaled = False
+                k._scaled = bool(envs_spyre.VLLM_SPYRE_USE_CHUNKED_PREFILL)
+                v._scaled = bool(envs_spyre.VLLM_SPYRE_USE_CHUNKED_PREFILL)
             elif len(attn_metadata.scale_indices) == 1:
                 # Decode
                 # Special case for decode of bs=1, pad the batch to be bs=2


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2061,6 +2063,34 @@ def _prepare_chunked_prefill(self, req_id: str) -> SamplingForwardInputs:
         self.model.n_pads_right = self.block_size - (((request_tkv - 1) % self.block_size) + 1)
         self.model.indices = torch.ones(1, dtype=torch.bool, device="cpu")
 
+        scale_indices = self.prefill_batch.request_indices
+        # Check if it is last prefill
```
prashantgupta24: moved the logic of `_maybe_prepare_last_prefill` here

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2061,6 +2063,34 @@ def _prepare_chunked_prefill(self, req_id: str) -> SamplingForwardInputs:
...
+
+            # Last prefill we need to setup the logitsprocessors to sampling
+            prefill_index = self.input_batch.add_request(request)
+            scale_indices = [prefill_index]
+            for logitsproc in self.input_batch.logitsprocs_wrappers:
+                logitsproc.set_prefill_index(prefill_index)
+
+            # Refresh sampling metadata after all request are added to the batch
+            self.input_batch.refresh_metadata()
+            self.prefill_batch.refresh_metadata()
```
prashantgupta24: earlier we had:
```
if num_computed_tokens + num_scheduled_tokens < prompt_len:
            return

```

tests/e2e/test_spyre_basic.py
```diff
@@ -66,6 +64,13 @@ def test_output(
         }
     )
 
+    if (
+        "micro-g3.3-8b-instruct-1b" in model.name
+        and model.is_quantized
+        and mode not in ["cb", "cp"]
+    ):
+        runtime_xfail(reason="SB sometimes causes failures with quantized model")
```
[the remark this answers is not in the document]
prashantgupta24: is there a `pc` mode?

vllm_spyre/v1/sample/golden_token_injector.py
```diff
@@ -159,23 +159,6 @@ def inject_token(
         expected_logprob = cast(torch.Tensor, expectation.logprobs)[current_token_idx]
         expected_prob = math.exp(expected_logprob)
 
-        # We'll inject only if the error is below the threshold
```
joerunde: Nice, this should help remove an interesting failure case for PGTI as well

maxdebayser: But why is this required for this PR?

vllm_spyre/v1/worker/spyre_model_runner.py
maxdebayser: Wouldn't it be cleaner to make `_maybe_prepare_last_prefill` return the indices and call it from `_prepare_chunked_prefill` instead of `prepare_model_input`? `_prepare_chunked_prefill` is getting pretty long.

prashantgupta24: I could call `_maybe_prepare_last_prefill` from within the `_prepare_chunked_prefill` function 🤔

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2061,6 +2063,34 @@ def _prepare_chunked_prefill(self, req_id: str) -> SamplingForwardInputs:
...
+            cur_n_blocks = math.ceil(self.tkv / self.block_size)
+            new_n_blocks = max(req_n_blocks, cur_n_blocks)
+            assert new_n_blocks > 0
+            base_n_tokens = (new_n_blocks - 1) * self.block_size
+            req_tkv_new_block = base_n_tokens + (prompt_len - 1) % self.block_size + 1
+            cur_tkv_new_block = base_n_tokens + (self.tkv - 1) % self.block_size + 1
+            self.tkv = max(req_tkv_new_block, cur_tkv_new_block)
+
+            # Last prefill we need to setup the logitsprocessors to sampling
+            prefill_index = self.input_batch.add_request(request)
```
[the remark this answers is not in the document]
joerunde: Yeah, in the CB model runner the `prefill_index` is saying "This is the index into my input batch of the request that is currently prefilling". But I'm not convinced that's correct to do here 🤔

vllm_spyre/v1/worker/spyre_model_runner.py
prashantgupta24: See now

vllm_spyre/v1/worker/spyre_model_runner.py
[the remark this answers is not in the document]
prashantgupta24: Added a TODO for the renaming of that variable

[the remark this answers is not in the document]
prashantgupta24: Hmm I think I see now what you mean. Trying to figure out how to test that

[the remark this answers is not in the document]
maxdebayser: ~Or instead of indexing the scales by the batch index, we could perhaps index them by request id.~ (in the model forward call we don't have the request ids) But keeping track of the scales could be moved to the model runner instead of the model implementation.

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -867,7 +867,7 @@ def __init__(
         self.prefill_batch = SamplingInputBatch(
             # TODO: review this, currently we only support prefill for
             # `batch_size=1`
-            max_num_reqs=1,
+            max_num_reqs=self.scheduler_config.max_num_seqs,
```
yannicks1: as far as I can tell increasing this dimension is only needed for chunked prefill, not for continuous batching. 
Would it make sense to only make this change in the derived `ChunkedPrefillModelRunner`? This would make it clearer to the reader and also not allocate bigger tensors than necessary for CB.

.github/workflows/test.yml
```diff
@@ -55,6 +55,11 @@ jobs:
             flags: "--timeout=600 -k 'basic and test_output' --durations=0"
             hf_model: "ibm-ai-platform/micro-g3.3-8b-instruct-1b-FP8"
             hf_model_rev: "0dff8bacb968836dbbc7c2895c6d9ead0a05dc9e"
+          - name: "fp8 scheduler tests with PC and CP"
+            markers: "cpu and quantized and prefix_caching"
+            flags: "--timeout=600 -k 'test_spyre_pc_scheduler_steps' --durations=0"
+            hf_model: "ibm-ai-platform/micro-g3.3-8b-instruct-1b-FP8"
+            hf_model_rev: "0dff8bacb968836dbbc7c2895c6d9ead0a05dc9e"
```
prashantgupta24: I have not changed the existing fp8 tests because those test the output and these check the scheduler. Worth leaving them separate?

vllm_spyre/v1/worker/spyre_input_batch.py
```diff
@@ -206,6 +206,7 @@ class ChunkedPrefillRequestState(SamplingRequestState):
     padding_blocks: int = 0
     usable_blocks: int = 0
     total_hit_blocks: int = 0
+    batch_index: int = -1  # Index in the batch, set when request is added
```
joerunde: I wouldn't mind `assert self.batch_index >= 0` in `__post_init__`

PR #635, opened 2026-01-16
:sparkles: support granite 4 dense

joerunde:
# Description

Adds support for configuring the unreleased granite 4 dense models the same way as the granite 3 models, so they work out of the box


tjohnson31415 (approved):
LGTM!

PR #637, opened 2026-01-20
📝 Update granite logs

joerunde:
# Description

This updates some log messages when granite 8b dense models are detected. This removes confusion when a user loads a granite 4 8b dense model, and would see log messages about `granite-3.3-8b-instruct`


tjohnson31415 (approved):
LGTM

PR #669, opened 2026-01-28
Overhaul configuration application and validation.

tjohnson31415:
## Description

Refactors the model configuration system from validation-only to a comprehensive registry-based approach:

- **Registry Architecture**: Centralized `ModelConfigRegistry` manages configurations from unified `model_configs.yaml`
- **Pattern Matching**: Model matching with complexity scoring to prioritize quantized over base models
- **Configuration Application**: `ModelConfigurator` applies environment variables and other model-device specific configurations
- **Structured Data**: Dataclasses replace ad-hoc dictionaries for type safety
- **Comprehensive Testing**: 110+ test cases (~1,500 lines) with unit, integration, and error handling coverage

## Key Changes

- Consolidates `known_model_configs.json` and `supported_configs.yaml` into single `model_configs.yaml` with YAML anchors for reuse
- Moves 150+ lines of hardcoded Granite configuration from `SpyrePlatform` to declarative YAML
- Adds `VLLM_SPYRE_REQUIRE_KNOWN_CONFIG` environment variable for strict validation
- Configuration summaries track default vs. applied values with override warnings

## Benefits

- **Extensibility**: Add models via YAML only (no code changes)
- **Maintainability**: Centralized configuration, structured data types
- **Testability**: Isolated component testing
- **Documentation**: Self-documenting YAML + comprehensive README

## Backward Compatibility

Maintains all existing model configurations and runtime behavior. New validation is opt-in via `VLLM_SPYRE_REQUIRE_KNOWN_CONFIG=1`.

Made with IBM Project Bob

### DIFF
--- a/vllm_spyre/config/README.md
+++ b/vllm_spyre/config/README.md
@@ -54,14 +54,17 @@ if configurator:
 
 ### Custom Configuration File
 
-By default, the registry loads from `vllm_spyre/config/model_configs.yaml`. You can override this in three ways (in priority order):
+By default, the registry loads from `vllm_spyre/config/model_configs.yaml`. You
+can override this in three ways (in priority order):
 
 1. **Explicit path**: Pass `config_path` to `initialize()`
+
    ```python
    registry.initialize(config_path=Path("/path/to/custom_config.yaml"))
    ```
 
 2. **Environment variable**: Set `VLLM_SPYRE_MODEL_CONFIG_FILE`
+
    ```bash
    export VLLM_SPYRE_MODEL_CONFIG_FILE=/path/to/custom_config.yaml
    ```


tjohnson31415:
bot:test
MARKERS="spyre and prefix_caching"

vllm_spyre/platform.py
```diff
@@ -209,11 +206,62 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                        "ibm-granite/granite-3.3-8b-instruct",
+                        "ibm-granite/granite-3.3-8b-instruct-FP8",
+                        "ibm-granite/granite-4-8b-dense",
+                    ]
+                ):
+                    logger.info(
+                        "Granite model detected. For backwards compatibility, "
+                        "defaulting --max-num-batched-tokens to 1024"
+                    )
+                    vllm_config.scheduler_config.max_num_batched_tokens = 1024
```
joerunde: does this disallow user overrides?

tests/config/test_integration.py
```diff
@@ -0,0 +1,373 @@
...
+        assert configurator.device_config is not None
+        assert configurator.device_config.env_vars is not None
+        assert configurator.device_config.num_gpu_blocks_override is not None
+
+
+class TestConfigurationApplication:
+    """Tests for applying configuration settings."""
+
+    @patch.dict(os.environ, {}, clear=True)
+    def test_apply_configuration_sets_env_vars(
```
joerunde: Should these test cases be parameterized to run on both the granite 3 and granite 4 configs? I think we'd want to ensure that granite 4 keeps configuring the same way too

vllm_spyre/config/model_registry.py
```diff
@@ -0,0 +1,384 @@
...
+    """Get the global model registry instance.
+
+    This is a convenience function that ensures the registry is initialized.
+
+    Returns:
+        The initialized ModelConfigRegistry instance
+    """
+    registry = ModelConfigRegistry.get_instance()
+    if not registry._initialized:
+        registry.initialize()
```
joerunde: Thoughts on adding an escape hatch here for users to author their own yaml and configure the path with an environment variable or something similar?

vllm_spyre/config/configurators/model_configurator.py
```diff
@@ -0,0 +1,275 @@
...
+class ConfigValue:
+    """Tracks a configuration value with override information.
+
+    Attributes:
+        expected: The expected/default value from model configuration
+        actual: The actual value that was applied (may differ if overridden)
+    """
+
+    expected: str | int | None
+    actual: str | int | None
```
joerunde: These names just make me think of testing 🤔 

Would `default / applied` better indicate the intent here?

vllm_spyre/platform.py
tjohnson31415: Yeah... the setting CLI arg defaulting was the last piece to be implemented, but should have been the solution instead of using VLLM_DT_CHUNK_LEN. Your comment does point out that technically I should check for TP=4 here and not just the model name.

> We've documented that this is overridable so I do think it might be worth changing to make it so here.

Worth changing the documentation or change the behavior (edge-case breaking change)?

I figured we'd just fix this in the next major release and have the CLI arg be the only way to configure the chunk size.

vllm_spyre/config/model_config.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Count non-None attributes
+        for attr_name, attr_value in self.attributes.items():
+            if attr_value is None:
+                continue
+
+            # Quantization config gets extra weight as it's a key differentiator
+            if attr_name == "quantization_config" and isinstance(attr_value, dict):
+                # Base score for having quantization_config
+                score += 10
```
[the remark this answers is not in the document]
tjohnson31415: Nope, it was just Bob's preference... I'm not a fan of the complexity score stuff anyhow, so I'll revisit this.

maxdebayser:
There seem to be some problems with github today and the approve button is returning an error. But I'm ok with the changes in this PR.

PR #684, opened 2026-02-04
test: fix test_spyre_seed to account for batch variance

tjohnson31415:
# Description

Test seeded random sampling behavior for both single and batched requests by comparing across requests instead of within a request.

`test_seed` test was made `runtime_xfail` in https://github.com/vllm-project/vllm-spyre/pull/661/changes/791b8b5e4d64ece45d9d5becb040399ca81a9243 because it was observed to fail with FP8 + Chunked Prefill. Due to numerical differences in batch processing, individual items within a batch may differ even when seeded; this is particularly true for FP8 and Chunked Prefill which increase the variance. However, entire batches should be reproducible across multiple requests when using the same seed.

### DIFF
--- a/tests/e2e/test_spyre_seed.py
+++ b/tests/e2e/test_spyre_seed.py
@@ -1,6 +1,9 @@
-"""Verification of seeded random sampling to be deterministic
+"""Verification of seeded random sampling behavior.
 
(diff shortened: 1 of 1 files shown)


joerunde:
bot:test
MARKERS="spyre and (chunked_prefill or prefix_caching) and quantized"

PR #696, opened 2026-02-09
🔥  Remove continuous-batching specific model code

joerunde:
# Description

This PR cleans up the model code, flattening all of our model classes because we now only have a single implementation. A couple changes leaked out into the model runner:
1. There was a nosy reference to self.model.model.model ❗  
2. `scale_indices` has been removed from the attention metadata as it's no longer used

~~This PR depends on #693 but github issues today prevent it from merging~~

Ready for review

### DIFF
--- a/vllm_spyre/utils.py
+++ b/vllm_spyre/utils.py
@@ -37,11 +37,6 @@ def stagger_region(limit: int, world_size: int, rank: int):
             if rank >= (_set + 1) * limit:
                 continue
             torch.distributed.barrier()
-
-        import traceback
-
-        traceback.print_stack()
-
         logger.info("Stagger Region: All Complete")
 
 


vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -461,53 +348,43 @@ def set_past_key_value_states(self, num_blocks) -> None:
...
+        masks: torch.Tensor,
         is_prompt: bool,
-        **extra_kwargs,
     ) -> torch.Tensor:
         forward_context = get_forward_context()
 
         attn_metadata = cast(SpyreAttentionMetadata, forward_context.attn_metadata)
         assert attn_metadata is not None
-        # import will be not be needed/ handled by FMS soon
+        # FMS does not eagerly register the paged attention algorithm, we must import it here
```
joerunde: "soon" here was too optimistic- this line still cannot be removed. Assuming it will never be fixed

maxdebayser (approved):

PR #698, opened 2026-02-10
feat: :arrow_up: support vllm v0.15.0

tjohnson31415:
# Description

Adds support for vLLM v0.15.0.

Also adds a pin for llvmlite which is actually not resolved correctly by uv. vLLM pins numba==0.61.2 which is supposed to keep `llvmlite < 0.45` [REF](https://github.com/numba/numba/blob/0.61.2/setup.py#L27-L28)


joerunde:
Thanks @tjohnson31415!

Do we wanna also do a full package refresh and update all dependencies in the lockfile?

PR #699, opened 2026-02-10
🔀 Merge changes from (#669) and (#684)

joerunde:
# Description

Merges in @tjohnson31415's changes from #669 and #684.

This also removes static batching support from the model configurators to bring them up to date with the 2.0 release prep branch

PR #700, opened 2026-02-10
:fire: remove continuous batching scheduler

joerunde:
# Description

This removes the CB scheduler class, and flattens the two model runner output classes that we had previously

Closes torch-spyre/spyre-inference#106

### DIFF
--- a/vllm_spyre/platform.py
+++ b/vllm_spyre/platform.py
@@ -159,14 +159,10 @@ class SpyrePlatform(Platform):
             )
 
         if is_decoder:
-            if envs_spyre.VLLM_SPYRE_USE_CHUNKED_PREFILL:
-                scheduler_config.scheduler_cls = (
-                    "vllm_spyre.v1.core.scheduler.ChunkedPrefillSpyreScheduler"
-                )
-            else:
-                scheduler_config.scheduler_cls = (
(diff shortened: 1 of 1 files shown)


maxdebayser:
One thing still missing is to remove the [reference](https://github.com/vllm-project/vllm-spyre/blob/3b401628dc9c58e9603670c0b9d3b3bae7cc532c/vllm_spyre/platform.py#L167) to the `ContinuousBatchingSpyreScheduler` class in `platform.py`

Also, shouldn't the PR title be "remove the continuous batching scheduler"?

joerunde:
> shouldn't the PR title be "remove the continuous batching scheduler"?

Ah, whoops 🤦 

> One thing still missing is to remove the [reference](https://github.com/vllm-project/vllm-spyre/blob/3b401628dc9c58e9603670c0b9d3b3bae7cc532c/vllm_spyre/platform.py#L167) to the ContinuousBatchingSpyreScheduler class in platform.py

ah yeah, I went to do that and then sidequested hard on getting all of Travis' changes into platform.py first: https://github.com/vllm-project/vllm-spyre/pull/699

but I'll just get that one reference out and then we can rip out all the CB stuff from platform.py after getting 699 in so we don't make way more churn

PR #714, opened 2026-02-12
🗑️ Update minimum vLLM and remove backwards compat cruft

tjohnson31415:
For 2.0, we can update vLLM backwards compatibility and remove a bunch of code meant to handle that.

Also re-linted some import lists after removing `if-then` import checks.


joerunde (approved):
LPGTM, @prashantgupta24 will be very happy

PR #715, opened 2026-02-12
Update examples

joerunde:
# Description

Updates example files to remove outdated configs.

I also had to set the multiproc method to spawn for mac os- these don't seem to run on my laptop without it.
The text examples also seemed to need an `if __name__ == "__main__"` otherwise the multiprocessing package barfs when trying to spin up the worker processes.

### DIFF
--- a/examples/online_inference/openai_spyre_text.py
+++ b/examples/online_inference/openai_spyre_text.py
@@ -9,9 +9,8 @@ First, start the server with the following command:
         --tensor-parallel-size=4 \
         --max-num-seqs=4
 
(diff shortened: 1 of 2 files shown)


examples/offline_inference/text_inference.py
```diff
@@ -0,0 +1,151 @@
+"""
+This example shows how to run offline inference using continuous batching.
```
tjohnson31415: NIT: Could remove the reference to continuous batching here.

Also, the heredoc for examples/offline_inference/vision_inference.py needs a similar update (and to remove `cb` from the referenced filename)

PR #728, opened 2026-02-17
:alebic: enable tests for 1.x support branch

joerunde:
# Description

Sets tests to run on PRs to the stable 1.x release branch


tjohnson31415:
I'm not sure it is necessary, but is it inentional that this PR is not adding `stable/1.x` to the on-push triggers?

PR #742, opened 2026-02-18
Add support for compile only backend

ktuvw:
# Description

Adds support for the "sendnn_compile_only" backend. Allows systems without Spyre cards to create compilation artifacts that are compatible with systems with Spyre cards.


joerunde:
Thanks @ktuvw!

This PR needs a docs update explaining this backend setting, additionally we have some test cases that test graph generation that could probably be edited to use this. Can you update `tests/aftu/test_compare_graphs.py` to use this new backend?

joerunde:
bot:test

joerunde:
There's one failing test but it's not related to the changes in this PR, so I think we can merge

PR #752, opened 2026-02-19
[WIP] Use vllm transformers backend for pooling model runner.

maxdebayser:
This PR explores using the vLLM model loader to load either the vLLM or the transformers modeling code in the pooling model runner. It also shows how to plug a custom attention class.

Why? Because 1) pooling models are simpler and don't require paged attention 2) We were already loading the transformers code for pooling, but in hacky way, without taking advantage of the vLLM pooler code.

This investigation showed the following:

1) In eager mode, both the vLLM and the tranformers backend work correctly
2) With torch 2.7, we can't compile the vLLM linear classes without graph breaks. Transformers backend tries to replace the liner classes in transformers code, but this can be circumvented.
3) The transformers backend only support the pooling models in transformers 5.0.0, but with a few hacks the 4.57 transformers can be supported
4) torch_sendnn can't transform the fx graph correctly for transformers 5.0.0

With a few hacks in this PR, both the 4.57 and the 5.0 transformers can run the pooling models correctly on cpu and inductor. But only 4.57 runs on the spyre device. With the 4.57 version, the custom attention class is not used. This means that this PR can be simplied a lot if we remove the code to support 5.0.

This PR builds on ideas from PR https://github.com/vllm-project/vllm-spyre/pull/217


maxdebayser:
Apparently the current stack works with torch 2.10. I'm going to test if the vllm modeling code compiles with this version.

PR #754, opened 2026-02-20
:sparkles: Add boh's custom RMS norm

joerunde:
# Description

Adds @bohnstingl's RMS norm implementation from [here](https://github.com/bohnstingl/vllm/commit/345b0f1eb954c80abad3129be13ef6b56077c5cf)

This sets up a `vllm_spyre_next.custom_ops` package starting with this new rms norm. The worker registers all custom ops at creation so that when it loads the model, any layers matching our custom ops will use our implementations instead.

I have this setup working (compilation passes with the custom RMS norm and the server boots) with a fresh install of the current main ranches of both vllm and torch-spyre, though this draft PR does not currently attempt to add the torch-spyre dependency and required build setup to compile everything correctly.

### DIFF
--- a/vllm_spyre_next/tests/test_vllm_spyre_next.py
+++ b/vllm_spyre_next/tests/test_vllm_spyre_next.py
@@ -1,11 +1,15 @@
-from vllm import LLM, RequestOutput, SamplingParams, sampling_params
+from vllm import LLM, RequestOutput, SamplingParams
+
+import pytest
+
 
 @pytest.mark.spyre
 def test_basic_model_load():
     model = LLM("ibm-ai-platform/micro-g3.3-8b-instruct-1b", max_model_len=128, max_num_seqs=2)
-    
+
     sampling_params = SamplingParams(max_tokens=5)
-    output: list[RequestOutput] = model.generate(prompts="Hello World", sampling_params=sampling_params)
+    output: list[RequestOutput] = model.generate(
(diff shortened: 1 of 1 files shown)


joerunde:
bot:test-next

joerunde:
bot:test-next

joerunde:
bot:next-test

joerunde:
bot:next-test

joerunde:
bot:next-test

joerunde:
bot:next-test

vllm_spyre_next/tests/test_vllm_spyre_next.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@pytest.mark.spyre
+def test_basic_model_load():
+    model = LLM("ibm-ai-platform/micro-g3.3-8b-instruct-1b", max_model_len=128, max_num_seqs=2)
+
+    sampling_params = SamplingParams(max_tokens=5)
+    output: list[RequestOutput] = model.generate(
+        prompts="Hello World", sampling_params=sampling_params
+    )
```
bohnstingl: Shall we compare the outputs here with the outputs running the vllm_spyre_next plugin in `cpu-mode`?
This raises a general question whether it would make sense to have a configuration possibility in this new plugin that certain layers (or the entire model) is excluded from the use of torch-spyre? It would enable detailed tests and also enable the user to disable spyre selectively if needed?

PR #774, opened 2026-02-27
[Spyre-Next] Integrated custom attention backend

bohnstingl:
## Description

This PR provides the sekeleton to add new attention backends to the vllm_spyre_next plugin.
It uses the first draft implementation of the pytorch-native paged attention as an example

## Related Issues

https://github.com/vllm-project/vllm-spyre/issues/648

## Checklist

- [X] I have read the [contributing guidelines](https://blog.vllm.ai/vllm-spyre/contributing/)
- [X] My code follows the project's code style (run `bash format.sh`)
- [X] My commits include a `Signed-off-by:` line (DCO compliance)

cc @tdoublep @dilipgb

### DIFF
--- a/vllm_spyre_next/tests/attention_test_utils.py
+++ b/vllm_spyre_next/tests/attention_test_utils.py
@@ -1,0 +1,361 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Utility functions for attention-related v1 tests."""
+
+from dataclasses import dataclass
+
+import pytest
+import torch
+
+from vllm.config import (
+    CacheConfig,
+    CompilationConfig,
+    DeviceConfig,
+    LoadConfig,
+    ModelConfig,
+    ParallelConfig,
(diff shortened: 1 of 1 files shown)


joerunde:
thanks @bohnstingl!

General questions:
1. Should we update the plugin to make this attention backend the default? This would make it easier to run upstream tests on it without modification, for example
2. Can we turn the example here into a small test case? That way we can be sure the attention backend continues to compile and run with each PR

joerunde:
> However, I have not yet tried this backend with torch-spyre and it will likely fail

Ah, interesting. I'm a little confused then about why we'd want to merge this in then, don't we want the code here to all work and run with torch-spyre?

> So I would rather want to run those upstream tests with the backends that we develop here integrated. WDYT?

Agreed, and we should probably start integrating upstream tests soon. Is your plan to put the pytorch native implementation here, or to put it upstream and consume it here?

If we do want to merge a working implementation here that runs with torch-spyre, then I think at least as a stop-gap we should have a small test here covering it while we work towards consuming upstream tests

PR #775, opened 2026-02-27
Remove last block deduplication

maxdebayser:
This PR further cleans up the prefix caching code by removing the de-duplication of the last block in case of a full match. This simplifies the model runner code and requires no patching of the vllm scheduler. There is no performance impact.


maxdebayser:
Closing because I forgot to push the merge from main in the other PR.

PR #778, opened 2026-02-27
chore(0.16.0-support): bump vllm to v0.16.0

prashantgupta24:
## Description

Bump to vLLM v0.16.0 without backwards compatibility (2.0 is still pre-release)


prashantgupta24:
bot:test

tjohnson31415:
I was able to test and confirm working the changes around using the profiler. Tested by passing `--profiler-config '{"profiler": "torch", "torch_profiler_dir": "./vllm_profile"}'` to `vllm serve`. Worked both with and without the AIU Kineto  tracing. In that testing, I saw some logging improvements to be made, so pushed those too (eg. adjusting the log order now that we use the wrapper).

joerunde (approved):
lpgtm

PR #784, opened 2026-03-02
[Spyre-Next][Spyre] ci: add workflow to run uv lock --check

tjohnson31415:
## Description

When doing dependency updates, it is easy to forget to also update uv.lock with the changes. This PR adds a workflow to run `uv lock --check` to remind contributors to do the update.

### DIFF
--- a/vllm_spyre_next/pyproject.toml
+++ b/vllm_spyre_next/pyproject.toml
@@ -47,15 +47,16 @@ root = ".."
 override-dependencies = [
     # No triton support yet, don't allow any dependencies to pull it in
     "triton; sys_platform == 'never'",
-    "intel-extension-for-pytorch; sys_platform == 'never'",
-    
-    # This torch version must match the one below in `build-constraint-dependencies`
-    # Skip packages on s390x and ppc64le, expected to be pre-installed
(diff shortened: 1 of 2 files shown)


prashantgupta24:
Hmm, the error didn't go away 🤔 
```
The lockfile at `uv.lock` needs to be updated, but `--check` was provided. To update the lockfile, run `uv lock`.

```

tjohnson31415:
Hmm, running some commands locally and it seems that `uv lock --check` always returns saying that the lockfile needs to be updated 🤔

Just repeating `uv lock; uv lock --check` always gets the same result. I see the same behavior on Mac and in a Linux container... So it seems that this idea will not work in CI because it will never pass.

prashantgupta24:
Huh, that worked?

tjohnson31415 (commented):
Nice find on the duplicate torch override!

I'm not 100% on all the uv.lock diff, but it seems fine 🤷 

I noticed `llvmlite` had a duplicate too, so I combined those. Also removed the `intel-extension-for-pytorch` override for "next" since v0.16.0 no longer has a reference to it in the requirements files.

prashantgupta24 (approved):
LGTM!

PR #815, opened 2026-03-09
[Spyre-Next] feat: add script to sync upstream test deps

rafvasq:
## Description

Adds `scripts/sync_upstream_tests.py` to help keep the `upstream-tests` dependency
group in sync with vLLM's test dependencies.

Prek integration can be a follow-up PR

## Related Issues

Closes torch-spyre/spyre-inference#54 

## Checklist

- [x] I have read the [contributing guidelines](https://blog.vllm.ai/vllm-spyre/contributing/)

### DIFF
--- a/vllm_spyre_next/scripts/sync_upstream_tests.py
+++ b/vllm_spyre_next/scripts/sync_upstream_tests.py
@@ -21,15 +21,15 @@ PROJECT_ROOT = Path(__file__).parent.parent
 def extract_vllm_commit(pyproject_path: Path) -> str:
     """
     Extract the vLLM git commit/tag from pyproject.toml.
-    
+
     Returns the commit hash or tag specified in [tool.uv.sources].
     """
     with open(pyproject_path, "rb") as f:
         data = tomllib.load(f)
-    
+
     try:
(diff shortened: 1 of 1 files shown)


prashantgupta24 (commented):
Do you want to run this by Bob for review once? Bob has some nice findings. I can post them here too if you want.

one of the high priority one being:
```
subprocess.run should capture stderr for better error reporting
```

joerunde:
> Do we really need all upstream test deps? When I'm testing stuff locally I usually only install pytest and tblib.

right, but this automation takes the guesswork out of figuring out which packages need to be installed as we integrate more upstream tests and update vllm. In practice I'm not worried about things like install overhead as we should be running on machines with uv caching enabled. The one potential problem I'm aware of is new upstream test dependencies causing install conflicts in our environments, but I'd still rather have this automation catch that when we update vllm.

> @joerunde do we still need to track dependencies in upstream-tests given the incoming testing framework or whether a minimal set like pytest and tblib is already sufficient?

I think so, yeah. We're still trying to contain all of the logic here in vllm-spyre related to running the upstream vllm tests, see the pytest plugin in https://github.com/vllm-project/vllm-spyre/pull/837

PR #853, opened 2026-03-20
[Spyre-Next] Pytorch Native Attention on Spyre

jvlunteren:
## Description

This PR builds on @bohnstingl’s branch (https://github.com/bohnstingl/vllm/tree/naive_attn_backend) and includes his commits. @bohnstingl is a co‑author of this work.

It introduces a vLLM v1 attention backend (`SpyreAttentionPagedBackend`) that runs on Spyre hardware using only PyTorch native operations (matmul, softmax). The backend implements the standard vLLM v1 triple (`AttentionBackend `/ `AttentionImpl `/ `AttentionMetadataBuilder`) and processes each forward pass in six steps:

1. Write new key/value tokens into the paged KV cache via vectorized scatter (`slot_mapping`)
2. Gather the relevant KV entries from scattered blocks into compact `[num_seqs, seq_len, num_kv_heads, head_size]` tensors
3. Reshape the flat query token batch into per-sequence padded tensors
4. Build a per-sequence boolean attention mask encoding padding validity and (for prefill) causality
5. Compute attention per sequence with query chunking
6. Extract only the real (non-padded) output tokens back to the flat batch layout

Due to current limitations in Spyre’s matmul support, attention is computed one sequence at a time rather than across the full batch. In addition, all attention heads are fused into a single 2D matmul by reshaping Q, K, and V from `[num_heads, seq_len, head_size]` into `[num_heads × seq_len, head_size]`. A block‑diagonal attention mask is applied to zero out spurious cross‑head interactions introduced by the concatenated layout. This makes the result equivalent to `num_heads `independent attention computations, at the cost of a single larger matrix multiply. To match the hardware’s expected memory layout, the kernel “stickifies” inputs to the matmul operation using a double `transpose `followed by `.contiguous()`.

Spyre requires static tensor shapes at compile time (`torch._dynamo.config.assume_static_by_default = True`). To avoid a full recompile on every decode step, KV tensors are padded to the next multiple of `KV_LENGTH_ALIGNMENT = 256`. This buckets sequence lengths into tiers (256, 512, 768, …) so that only the first request at each tier pays the compilation cost. Query tokens are chunked to a fixed `QUERY_CHUNK_SIZE = 32` for the same reason.

Grouped-query attention is handled by `repeat_interleave` on the KV heads before the kernel, expanding `num_kv_heads `to `num_heads `in one vectorized call.

A `use_sdpa `flag routes computation through `torch.nn.functional.scaled_dot_product_attention` instead of the Spyre kernel. This uses the semantic mask directly (no block-diagonal needed, since `SDPA `handles heads independently) and is primarily used for CPU testing and debugging.

## Related Issues

Relates to torch-spyre/spyre-inference#122 

## Test Plan

I created and executed a debug build that validates SDPA (CPU) outputs against the Spyre attention implementation for a range of models and prompts. The discrepancies were within acceptable bounds, accounting for Spyre’s float16 precision versus the precisions used by SDPA on the CPU. I also added a unit test (`test_spyre_paged_attn.py`) for this validation, which is included in this PR.

## Checklist

@tdoublep @bohnstingl


vllm_spyre_next/tests/test_spyre_paged_attn.py
```diff
@@ -0,0 +1,355 @@
...
+    if use_sdpa:
+        atol, rtol = 0.1, 0.1
+    elif max_query_len >= 32:
+        # Large prompts accumulate float16 rounding errors across many dot-product
+        # terms (query_len * kv_len * head_size multiplications per head). The
+        # transposed Spyre kernel sums all heads in one matmul, compounding the
+        # error further. The reference uses float32 softmax internally, widening
+        # the gap. rtol=5.0 is loose but expected; the results are numerically
+        # equivalent and the gap does not grow with model scale.
+        atol, rtol = 0.3, 5.0
```
bohnstingl: Are these tolerances okay? They seem a bit large to me. Torch-spyre has 0.1 absolute and relative tolerances for its attention tests, see https://github.com/torch-spyre/torch-spyre/blob/7b3b4dcfe83838c65e9ef1d3a6aedce2605ae7df/tests/inductor/test_building_blocks.py#L57

bohnstingl (changes requested):
I took a first look at the "infrastructure" surrounding the attention. I will next look at the attention implementation itself

vllm_spyre_next/tests/test_spyre_paged_attn.py
jvlunteren: If I recall correctly, I observed similar deviations when using SDPA during debugging. I initially attributed them to float16 precision, but I need to verify this.

vllm_spyre_next/vllm_spyre_next/compat_utils.py
```diff
@@ -0,0 +1,5 @@
+from dataclasses import fields
+
+
+def dataclass_fields(cls: type) -> list[str]:
+    return [f.name for f in fields(cls)]
```
bohnstingl: As mentioned in my other comment (https://github.com/vllm-project/vllm-spyre/pull/853#discussion_r2982965500), I think this can also be removed

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+    AttentionType,
+    CommonAttentionMetadata,
+    MultipleOf,
+)
+from vllm.v1.kv_cache_interface import AttentionSpec
+
+
+# Disable dynamic shapes to force static shape compilation for Spyre
+torch._dynamo.config.assume_static_by_default = True
+torch._dynamo.config.automatic_dynamic_shapes = False
```
bohnstingl: We probably shouldn't configure `torch._dynamo` globally here, but rather directly set `dynamic=False` when running `torch.compile(...)`, see my other comment below.

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+
+class SpyreAttentionPagedBackend(AttentionBackend):
+    """Pure PyTorch implementation of PagedAttention."""
+
+    accept_output_buffer: bool = True
+    supported_dtypes: ClassVar[list[torch.dtype]] = [
+        torch.float16,
+        torch.bfloat16,
+        torch.float32,
+    ]
```
bohnstingl: I think here we need to specify only `torch.float16`, as the other data types are not supported in torch-spyre, I believe.

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+    @staticmethod
+    def get_kv_cache_shape(
+        num_blocks: int,
+        block_size: int,
+        num_kv_heads: int,
+        head_size: int,
+        cache_dtype_str: str = "auto",
+    ) -> tuple[int, ...]:
+        """KV cache shape: [2, num_blocks, block_size, num_kv_heads, head_size]"""
+        return (2, num_blocks, block_size, num_kv_heads, head_size)
```
bohnstingl: @tdoublep was opting for an implementation following `(num_blocks, 2, ...)`, is this still the case, or do we want to use this format? See https://github.com/vllm-project/vllm-spyre/pull/774#discussion_r2864850033

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+        self.scale = float(scale)
+        self.num_kv_heads = num_kv_heads
+        self.num_queries_per_kv = num_heads // num_kv_heads
+        self.kv_cache_dtype = kv_cache_dtype
+        self.attn_type = attn_type
+
+        if working_precision is not None:
+            self.working_precision = working_precision
+        else:
+            self.working_precision = torch.bfloat16
```
bohnstingl: I think at the moment we have to enforce `torch.float16` because of spyre limitations?

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+        output: torch.Tensor | None = None,
+        output_scale: torch.Tensor | None = None,
+        output_block_scale: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        """Compute attention output using PyTorch native operations."""
+
+        assert output is not None, "Output tensor must be provided"
+
+        if attn_metadata is None:
+            return output.fill_(0)
```
bohnstingl: Is the explicit filling with `0` necessary? The cpu attention backend in upstream vLLM is not doing it, see https://github.com/vllm-project/vllm/blob/ffb5b32b5f2a37ea58261747aef9e5f3907b9941/vllm/v1/attention/backends/cpu_attn.py#L293-L294

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+        mask: torch.Tensor | None,  # [1, 1, max_query_len, max_seq_len]
+    ) -> torch.Tensor:
+        """Compute attention for a single sequence using Spyre.
+
+        Processes queries in fixed-size chunks of QUERY_CHUNK_SIZE tokens.
+        """
+        batch_size, query_len, num_heads, head_size = query.shape
+        _, kv_len, num_kv_heads, _ = key.shape
+
+        assert batch_size == 1, f"Expected batch_size=1, got {batch_size}"
```
bohnstingl: If this is a real limitation of the current implementation, can we bring this up already to the `__init__` or to some earlier point to error out already there?

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+                query_chunk_padded,
+                key_squeezed,
+                value_squeezed,
+                mask_chunk,
+                chunk_len,
+                num_heads,
+                head_size,
+                kv_len,
+                chunk_idx,
+            )
```
bohnstingl: ... and this should run on spyre, right? Can we add a note in the comments for that maybe?

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+                vt_stickified = vt_stickified.to(torch.float16)
+                sm_scale_1d = sm_scale_1d.to(torch.float16)
+                mask_values_stickified = mask_values_stickified.to(torch.float16)
+
+            # Transfer to Spyre device
+            qt_spyre = qt_stickified.to(spyre_device)
+            k_spyre = k_stickified.to(spyre_device)
+            vt_spyre = vt_stickified.to(spyre_device)
+            sm_scale_spyre = sm_scale_1d.to(spyre_device)
+            mask_values_spyre = mask_values_stickified.to(spyre_device)
```
bohnstingl: Could we do the movements to and from spyre using the helper functions in [custom_ops/utils.py](https://github.com/vllm-project/vllm-spyre/blob/8c8cd7a40a27a884cd5b7220449f8752db81c84d/vllm_spyre_next/vllm_spyre_next/custom_ops/utils.py#L50-L74)?

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_paged_attn.py
```diff
@@ -0,0 +1,814 @@
...
+
```
bohnstingl: Could we do the movements to and from spyre using the helper functions in [custom_ops/utils.py](https://github.com/vllm-project/vllm-spyre/blob/8c8cd7a40a27a884cd5b7220449f8752db81c84d/vllm_spyre_next/vllm_spyre_next/custom_ops/utils.py#L50-L74)? See [rms_norm](https://github.com/vllm-project/vllm-spyre/blob/8c8cd7a40a27a884cd5b7220449f8752db81c84d/vllm_spyre_next/vllm_spyre_next/custom_ops/rms_norm.py#L203-L218) for an example.

bohnstingl (changes requested):
I made a first pass through the attention mechanism. In general it looks good to me, but I think we could do some cleanup and consolidation. Please let me know if something is unclear.

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_attn.py
```diff
@@ -0,0 +1,814 @@
...
+        mask: torch.Tensor | None,  # [1, 1, max_query_len, max_seq_len]
+    ) -> torch.Tensor:
+        """Compute attention for a single sequence using Spyre.
+
+        Processes queries in fixed-size chunks of QUERY_CHUNK_SIZE tokens.
+        """
+        batch_size, query_len, num_heads, head_size = query.shape
+        _, kv_len, num_kv_heads, _ = key.shape
+
+        assert batch_size == 1, f"Expected batch_size=1, got {batch_size}"
```
jvlunteren: This limitation was introduced because Spyre’s matmul does not support four-dimensional input tensors. It will need to be modified anyway once the latter restriction is lifted.

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_attn.py
```diff
@@ -0,0 +1,800 @@
...
+
+    def build(
+        self,
+        common_prefix_len: int,
+        common_attn_metadata: CommonAttentionMetadata,
+        fast_build: bool = False,
+    ) -> SpyreAttentionMetadata:
+        """Build attention metadata from common metadata."""
+
+        # Extract information from common metadata
```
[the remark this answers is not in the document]
jvlunteren: I followed the approach used in our Triton attention backend, but I agree it could also be done the way you suggest.

bohnstingl (approved):
In principle the PR looks good to me.
There are some questions related to the test files that maybe @joerunde or @tjohnson31415 could comment on and apart from that, I think we should just get some more reviews in.

cc @tdoublep @yannicks1 @bringlein @maxdebayser

vllm_spyre_next/tests/test_attention_correctness_stripped.py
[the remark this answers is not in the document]
joerunde: Right, we should aim to reuse the upstream tests. Is the goal here to run the `test_causal_backend_correctness` test but with our `AttentionBackendEnum.CUSTOM` backend instead of with the full list of other supported backends?

We'll probably need to get the upstream test updated to correctly populate the supported attention backends based on the current platform, but in the meantime we could whip up something in our pytest plugin like this:
```yaml
    - rel_path: tests/v1/attention/test_attention_backends.py
      allow_list:
        - test: "test_causal_backend_correctness"
          mode: mandatory_pass
          tags: [attention]
          fixtures:
            - patch_backend_list
```

where we'd inject the provided fixture(s) into the test, which we could implement like:

```python
@pytest.fixture
def patch_backend_list(monkeypatch):
    # import test_attention_backends from the cached vllm upstream tests

    our_backend_list = [
        AttentionBackendEnum.CUSTOM,
    ]

    with monkeypatch.setattr(test_attention_backends, "BACKENDS_TO_TEST", our_backend_list)
        yield
```

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_attn.py
```diff
@@ -0,0 +1,779 @@
...
+    AttentionType,
+    CommonAttentionMetadata,
+    MultipleOf,
+)
+from vllm.v1.kv_cache_interface import AttentionSpec
+
+
+@dataclass
+class SpyreAttentionMetadata(AttentionMetadata):
+    """Metadata for PyTorch native attention computation."""
```
bringlein:
```suggestion
    """Metadata for PyTorch native attention computation on Spyre."""
```

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_attn.py
```diff
@@ -0,0 +1,779 @@
...
+
+    @classmethod
+    def supports_kv_cache_dtype(cls, kv_cache_dtype: CacheDType | None) -> bool:
+        if kv_cache_dtype is None:
+            return True
+        return kv_cache_dtype in cls.supported_kv_cache_dtypes
+
+
+class SpyreAttentionImpl(AttentionImpl[SpyreAttentionMetadata]):
+    """PyTorch native implementation of attention with paged KV cache."""
```
bringlein:
```suggestion
    """PyTorch native implementation of attention with paged KV cache on Spyre."""
```

bringlein (commented):
thanks for all your efforts @jvlunteren ! (and @bohnstingl ) 
Juts have a few comments

vllm_spyre_next/tests/attention_test_utils.py
```diff
@@ -0,0 +1,353 @@
+# SPDX-License-Identifier: Apache-2.0
```
yannicks1: this seems like mostly copied from [tests/v1/attention/utils.py](https://github.com/vllm-project/vllm/blob/4f6eed3bd4a92c6bd513460ee85b917d6df88a17/tests/v1/attention/utils.py), is there any fundamental changes (I did not do the diff) or could we just import the file from upstream vllm?

yannicks1 (commented):
disclaimer: only some initial comments. I am reviewing the actual backend in the afternoon. (just making sure no comments get lost)

vllm_spyre_next/tests/test_spyre_attn.py
```diff
@@ -0,0 +1,345 @@
...
+        sliding_window=sliding_window,
+        soft_cap=soft_cap,
+    )
+
+    if use_sdpa:
+        atol, rtol = 0.1, 0.1
+    elif max_query_len >= 32:
+        atol, rtol = 0.3, 5.0  # float16 accumulation errors for large prompts
+    else:
+        atol, rtol = 0.2, 0.2
```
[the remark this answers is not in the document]
bohnstingl: Basically, we hit every tolerance values in one way or another in the current tests.

vllm_spyre_next/vllm_spyre_next/platform.py
```diff
@@ -73,6 +81,13 @@ def log_server_boot(cls, vllm_config: VllmConfig) -> None:
...
 
+    @classmethod
+    def get_attn_backend_cls(cls, selected_backend, *args, **kwargs) -> str:
+        if selected_backend == AttentionBackendEnum.CUSTOM:
+            return AttentionBackendEnum.CUSTOM.get_path()
+        else:
+            return super().get_attn_backend_cls(selected_backend, *args, **kwargs)
+
     @classmethod
     def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
```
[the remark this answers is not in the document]
bohnstingl: I think this is actually not necessary anymore and I removed it again.

bohnstingl:
I made some new commits and changed:

1. The layout of the KV cache, it is now `(num_blocks, 2, ...)`
2. I revisited the logic about `max_num_seqs==1` and I think this was an old limitation that has meanwhile be resolved. The attention backend should support `max_num_seqs > 1`

@jvlunteren could you please confirm both points above?

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_attn.py
```diff
@@ -0,0 +1,814 @@
...
+    @staticmethod
+    def get_kv_cache_shape(
+        num_blocks: int,
+        block_size: int,
+        num_kv_heads: int,
+        head_size: int,
+        cache_dtype_str: str = "auto",
+    ) -> tuple[int, ...]:
+        """KV cache shape: [2, num_blocks, block_size, num_kv_heads, head_size]"""
+        return (2, num_blocks, block_size, num_kv_heads, head_size)
```
bohnstingl: I have changed the KV cache format to be (num_blocks, 2, ...). 
@jvlunteren please have a look at my changes and confirm.

bohnstingl: I have changed the KV cache format to be (num_blocks, 2, ...). 
@jvlunteren please have a look at my changes and confirm.

vllm_spyre_next/examples/torch_spyre_inference.py
```diff
@@ -16,6 +16,7 @@ def parse_args():
     parser.add_argument("--model", type=str, default="ibm-ai-platform/micro-g3.3-8b-instruct-1b")
     parser.add_argument("--max_model_len", "--max-model-len", type=int, default=2048)
     parser.add_argument("--max_num_seqs", "--max-num-seqs", type=int, default=2)
+    parser.add_argument("--max_num_batched_tokens", "--max-num-batched-tokens", type=int, default=2)
```
yannicks1, line 27: why would we want this to default to 2? this defines the number of tokens in the batch. 2 is super low, does it even work? we still use the base scheduler in `vllm_spyre_next` so the value is not overridden for our granite3.3-8b model as this was the case in `vllm_spyre`...

tdoublep: +1

This will create a lot of chunked prefills for the example below I guess? Is this something that even works right now?

vllm_spyre_next/vllm_spyre_next/v1/attention/backends/spyre_attn.py
```diff
@@ -0,0 +1,776 @@
...
+
+        # As fallback, use SDPA implementation
+        if self.use_sdpa:
+            return self._compute_attention_sdpa(query, key, value, mask)
+
+        # Allocate output tensor for all sequences
+        output_all_seqs = torch.zeros_like(query)
+
+        # Process each sequence separately
+        for seq_idx in range(num_seqs):
```
tdoublep, line 527: One of the next steps should be thinking about how we handle varlen batches in a proper way on Spyre

tdoublep (approved):
LGTM

tdoublep:
@jvlunteren @bohnstingl Can we resolve conflicts and merge?

PR #861, opened 2026-03-24
chore (1.x): support vllm 0.18.0 without backwards compat

tjohnson31415:
## Description

We want to make another release on the 1.x line and need to use the latest vLLM. We are allowing this change to drop support for older vLLM versions. vLLM supported range will now have 0.17.0 as a lower bound.

## Test Plan

Must pass all standard CI tests.

## Checklist


tjohnson31415:
Had a job fail with an client timeout / broken connection a couple times, but now it seems to have passed. I also reproduced it locally, but the was unable to reproduce it after a few more tries. So we may have a little bit of flakiness 🤔

PR #866, opened 2026-03-25
[Spyre-Next] Use vllm version from [tool.uv.sources]

sducouedic:
## Description

The current version of vllm dependency (v0.15.1) is wrong and conflicting with the version 0.17.1 specified in `[tool.uv.sources]`. That latter one should be used. We can remove the version specification entirely in `dependencies = [...` to have the version in a single place. 

## Test Plan

bot:next-test should still have two tests passing

## Checklist


tjohnson31415 (approved):
LGTM. The value here didn't have any effect in my testing, so this is a good cleanup

PR #870, opened 2026-03-26
Alternative fix for max_num_batched_tokens verification

maxdebayser:
Also change the defaults for max_num_tokens and max_num_seqs so that the pre-model loading verification in SchedulerConfig doesn't fail.


joerunde:
closing as stale and overcome

PR #884, opened 2026-04-01
[Spyre-Next] Custom fixtures for upstream tests

joerunde:
## Description

This PR adds the ability to specify custom fixtures to be injected for sets of upstream tests. This allows us to run custom setup and teardown where we may need to patch things like hardcoded inputs that assume cuda.

Eventually we will update the upstream tests to be more portable as part of a larger effort, but having custom fixtures allows us to run some upstream tests now so we don't have to wait until we've finished the work to fix the upstream tests. This is important for https://github.com/vllm-project/vllm-spyre/pull/853, which currently is vendoring upstream test files and editing them. This PR includes an example of a fixture to make those set of tests usable as-is.

This PR depends on #853, once that lands I will update the diff here to remove the duplicated test files.

## Related Issues

Removes duplicated test files that will land in #853 

## Test Plan

bot:next-test to ensure the upstream tests work 😄 

## Checklist

### DIFF
--- a/vllm_spyre_next/vllm_spyre_next/testing/pytest_plugin.py
+++ b/vllm_spyre_next/vllm_spyre_next/testing/pytest_plugin.py
@@ -34,7 +34,6 @@ XDG_CACHE_HOME          Base cache directory (default: ~/.cache)
 from __future__ import annotations
 
 import fnmatch
-import inspect
 import os
 import re
 import subprocess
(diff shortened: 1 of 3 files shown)


vllm_spyre_next/vllm_spyre_next/testing/pytest_plugin.py
```diff
@@ -325,6 +327,26 @@ def _prepare_upstream_tests_dir() -> Path:
     return tests_dir
 
 
+def _spicy_code_edits(upstream_tests_dir: Path):
```
tjohnson31415: NIT: "spicy" adds some good flavor, but maybe "_temp_upstream_workarounds" or "_fixme_code_edits" or something like that would be clearer.

bohnstingl (commented):
I like the idea of this PR a lot!
I haven't yet tested it myself, but @joerunde is this supposed to be working right now and can I try it, or are there limitations I should know about?

yannicks1:
was merged into new repo here: https://github.com/torch-spyre/spyre-inference/pull/141

PR #902, opened 2026-04-07
[bugfix] Fix token degradation issue

R3hankhan123:
## Description
Fix token degradation issues faced by power and Z when running benchmark script

## Test Plan
Run the benchmark script and test the token generated one

## Checklist

### DIFF
--- a/tests/e2e/test_chunked_prefill_tkv_steps.py
+++ b/tests/e2e/test_chunked_prefill_tkv_steps.py
@@ -390,3 +390,75 @@ def test_decode_padding_to_same_block(
     assert output.tkv == long_prompt_len + steps
     # One more real block was allocated for the short request
     assert runner.block_pool.get_num_free_blocks() == initial_free_blocks - 2
+
+
+@pytest.mark.cpu
+@pytest.mark.chunked_prefill
+@default_test_params
+def test_tkv_reset_on_new_batch(
+    model: ModelInfo,
+    max_model_len: int,
+    max_num_seqs: int,
+    max_num_batched_tokens: int,
+    monkeypatch: pytest.MonkeyPatch,
+):
+    """Test that TKV is properly reset when starting a new batch after all
+    requests are finished. This verifies that is_new_batch logic correctly
+    detects when _num_requests == 0 and resets TKV to the new prompt length."""
+    runner: SpyreModelRunner = get_cpu_model_runner(
+        model=model,
+        max_model_len=max_model_len,
+        max_num_seqs=max_num_seqs,
+        max_num_batched_tokens=max_num_batched_tokens,
+        monkeypatch=monkeypatch,
+    )
+
+    # First request with a specific prompt length
+    req_id_1 = "request_1"
+    prompt_len_1 = 50
+    new_req_data = make_new_request_data(req_id_1, prompt_len_1)
+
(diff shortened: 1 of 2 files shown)


vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -2382,7 +2382,7 @@ def add_new_request(self, request: NewRequestData):
         assert sampling_params is not None, "sampling_params are required for this model runner"
         assert prompt_token_ids is not None, "prompt token ids are required for this model runner"
 
-        is_new_batch = self.get_n_free_blocks() == self.n_blocks
+        is_new_batch = self.input_batch._num_requests == 0
```
tdoublep, line 2385: Shouldn't these two conditions be the same? If the batch is empty, shouldn't all blocks be free?

R3hankhan123: Ideally it should be the case but in practice when I added some debugging logs around these lines, i saw that when a new batch was coming, the TKV value was not resetting

yannicks1: this should be identical... Have you been able to verify that your solution solves the problem at hand?

tests/e2e/test_chunked_prefill_tkv_steps.py
```diff
@@ -390,3 +390,75 @@ def test_decode_padding_to_same_block(
     assert output.tkv == long_prompt_len + steps
     # One more real block was allocated for the short request
     assert runner.block_pool.get_num_free_blocks() == initial_free_blocks - 2
+
+
+@pytest.mark.cpu
+@pytest.mark.chunked_prefill
+@default_test_params
+def test_tkv_reset_on_new_batch(
```
yannicks1, line 398: I ran this test locally on main and it also passes. It does not seem to isolate the issue. [Was expecting a test that fails on main but passes with your proposed fix]

vllm_spyre/v1/worker/spyre_model_runner.py
R3hankhan123: Yes, san from power team and chandra from my team verified this PR it solved the problem reported

tests/e2e/test_chunked_prefill_tkv_steps.py
R3hankhan123: on the main branch this fix is already there

R3hankhan123: https://github.com/vllm-project/vllm-spyre/blob/fc6b54b9142eabd08eccb3879d32eaf43ef11b72/vllm_spyre/v1/worker/spyre_model_runner.py#L1251

yannicks1: sorry I meant the `stable/1.x` branch. Your test passes with and without your fix on CPU.

joerunde:
If we plan to fix anything on `stable/1.x` (though TBD if this is even a bug given the discussion on the internal issue) do any users exist to even pick it up? I thought that 1.10 was our final patch for our last user of the 1.x stream. We can discuss internally but let's not waste effort on this if all deadlines for 1.x dev have passed and no more patches will be picked up anyway

PR #903, opened 2026-04-07
feat: Add structured Output decoding support for vLLM-spyre

R3hankhan123:
## Description

Add structured decoding support for vllm spyre.

## Test Plan
Run the vllm server and set structured output backend to xgrammar and outlines

## Test output
1. With using xgrammar
 ```
[root@b314lp81 ~]# curl http://localhost:8000/v1/chat/completions \
  -X POST \
  -d '{
    "model": "ibm-granite/granite-3.3-8b-instruct",
    "messages": [
      {"role": "user", "content": "Classify sentiment: I love this product"}
    ],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "enum",
        "schema": {
          "type": "object",
          "properties": {
            "sentiment": {
              "type": "string",
              "enum": ["positive", "negative", "neutral"]
            }
          },
          "required": ["sentiment"]
        }
      }
    }
  }' | jq -r '.choices[0].message.content' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1161  100   623  100   538    629    543 --:--:-- --:--:-- --:--:--  1172
{
  "sentiment": "positive"
}
[root@b314lp81 ~]# curl http://localhost:8000/v1/chat/completions \
  -X POST \
  -d '{
    "model": "ibm-granite/granite-3.3-8b-instruct",
    "messages": [
      {"role": "user", "content": "Score this answer: The result is correct"}
    ],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "score",
        "schema": {
          "type": "object",
          "properties": {
            "score": { "type": "number" },
            "passed": { "type": "boolean" }
          },
          "required": ["score", "passed"]
        }
      }
    }
  }' | jq -r '.choices[0].message.content' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1129  100   627  100   502    430    345  0:00:01  0:00:01 --:--:--   775
{
  "score": 1,
  "passed": true
}
[root@b314lp81 ~]# curl http://localhost:8000/v1/chat/completions \
  -X POST \
  -d '{
    "model": "ibm-granite/granite-3.3-8b-instruct",
    "messages": [
      {"role": "user", "content": "Extract name and optional email: John (no email provided)"}
    ],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "optional",
        "schema": {
          "type": "object",
          "properties": {
            "name": { "type": "string" },
            "email": { "type": "string" }
          },
          "required": ["name"]
        }
      }
    }
  }' | jq -r '.choices[0].message.content' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1156  100   648  100   508    419    328  0:00:01  0:00:01 --:--:--   748
{
  "name": "John",
  "email": "not provided"
}
[root@b314lp81 ~]# curl http://localhost:8000/v1/chat/completions \
  -X POST \
  -d '{
    "model": "ibm-granite/granite-3.3-8b-instruct",
    "messages": [
      {"role": "user", "content": "Extract order: Order 123 has 2 apples and 3 bananas"}
    ],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "order",
        "schema": {
          "type": "object",
          "properties": {
            "order": {
              "type": "object",
              "properties": {
                "id": { "type": "string" },
                "items": {
                  "type": "array",
                  "items": {
                    "type": "object",
                    "properties": {
                      "product": { "type": "string" },
                      "quantity": { "type": "integer" }
                    },
                    "required": ["product", "quantity"]
                  }
                }
              },
              "required": ["id", "items"]
            }
          },
          "required": ["order"]
        }
  }' | jq -r '.choices[0].message.content' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1830  100   819  100  1011    179    222  0:00:04  0:00:04 --:--:--   244
{
  "order": {
    "id": "123",
    "items": [
      {
        "product": "apples",
        "quantity": 2
      },
      {
        "product": "bananas",
        "quantity": 3
      }
    ]
  }
}
[root@b314lp81 ~]# curl http://localhost:8000/v1/chat/completions \
  -X POST \
  -d '{
    "model": "ibm-granite/granite-3.3-8b-instruct",
    "messages": [
      {"role": "user", "content": "What is the capital of India?"}
    ],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "qa",
        "schema": {
          "type": "object",
          "properties": {
            "answer": { "type": "string" },
            "confidence": { "type": "number" }
          },
          "required": ["answer"]
        }
      }
    }
  }' | jq -r '.choices[0].message.content' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1124  100   641  100   483    411    309  0:00:01  0:00:01 --:--:--   720
{
  "answer": "New Delhi",
  "confidence": 1
}
```
With outlines
```
 [root@b314lp81 ~]# curl http://localhost:8000/v1/chat/completions \
  -X POST \
  -d '{
    "model": "ibm-granite/granite-3.3-8b-instruct",
    "messages": [
      {"role": "user", "content": "Extract name and age from: John is 28 and lives in Bangalore"}
    ],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "nested",
        "schema": {
          "type": "object",
          "properties": {
            "person": {
              "type": "object",
              "properties": {
                "name": { "type": "string" },
                "age": { "type": "integer" }
              },
              "required": ["name", "age"]
            },
            "city": { "type": "string" }
          },
          "required": ["person", "city"]
        }
      }
    }
  }' | jq -r '.choices[0].message.content' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2545  100  1818  100   727    224     89  0:00:08  0:00:08 --:--:--   464
{
  "person": {
    "name": "John",
    "age": 28
  },
  "city": "Bangalore"
}
[root@b314lp81 ~]# 
[root@b314lp81 ~]# curl http://localhost:8000/v1/chat/completions \
  -X POST \
  -d '{
    "model": "ibm-granite/granite-3.3-8b-instruct",
    "messages": [
      {"role": "user", "content": "Extract people: John is 28, Alice is 30"}
    ],
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "array",
        "schema": {
          "type": "object",
          "properties": {
            "people": {
              "type": "array",
              "items": {
                "type": "object",
                "properties": {
                  "name": { "type": "string" },
                  "age": { "type": "integer" }
                },
                "required": ["name", "age"]
              }
            }
          },
          "required": ["people"]
        }
      }
    }
  }' | jq -r '.choices[0].message.content' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1478  100   739  100   739    214    214  0:00:03  0:00:03 --:--:--   428
{
  "people": [
    {
      "name": "John",
      "age": 28
    },
    {
      "name": "Alice",
      "age": 30
    }
  ]
}
```

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,12 @@ override-dependencies = [
     # vllm 0.18.0 pins compressed-tensors==0.13.0; override to 0.14.0.1
     # TODO: remove once minimum vllm is bumped past 0.18.0
     "compressed-tensors==0.14.0.1",
+
+    # llguidance>=1.7.3 fixes s390x endianness issues.
+    # This conflicts with vLLM's version range (llguidance >= 1.3.0, < 1.4.0).
+    # TODO: Remove this override once vLLM's requirement range includes or moves past 1.7.3.
+    # See: https://github.com/vllm-project/vllm/blob/v0.19.1/requirements/common.txt#L22
+    "llguidance>=1.7.3"
 ]
 # This adds constraints to all dependent build environments, which will ensure everything is built
 # with the same version of torch. This CANNOT conflict with a package's existing build dependencies
--- a/sendnn_inference/v1/core/scheduler.py
+++ b/sendnn_inference/v1/core/scheduler.py
@@ -116,6 +116,7 @@ class PoolingSpyreScheduler(SpyreScheduler):
         while holdback_queue:
             self.waiting.append(holdback_queue.popleft())
 
+        outputs._spyre_grammar_output = self.get_grammar_bitmask(outputs)  # type: ignore[attr-defined]
         return outputs
 
     def _get_matching_warmup_shapes(
@@ -255,27 +256,17 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
         holdback_queue: deque[Request] = deque()
         while self.waiting:
             holdback_queue.append(self.waiting.popleft())
+        # Also drain skipped_waiting: structured-output requests whose
+        # grammar was not yet ready get placed here by the base scheduler.
+        # We must route them through holdback to enforce the
+        # one-prefill-at-a-time constraint.
+        while self.skipped_waiting:
+            holdback_queue.append(self.skipped_waiting.pop_request())
 
         # Check if new requests can be scheduled for prefill
         while holdback_queue:
             if self.can_schedule_prefill(holdback_queue[0]):
                 new_request = holdback_queue.popleft()
-                # Remove structured_output_request
-                # NB: SpyrePlatform.validate_request() removes structured_output
-                # before the request gets here in most cases
-                # TODO: We don't currently support structured output and it
-                # breaks some assumptions the code makes. The problems is that
-                # a structured output request will stay in waiting for multiple
-                # iterations with status WAITING_FOR_FSM. To handle this
-                # properly we need to exclude such requests from entering
-                # ongoing_prefills but still pass them in the waiting queue to
-                # the base scheduler to track the FSM initialization.
-                if new_request.structured_output_request is not None:
-                    logger.warning(
-                        "Removing structured output from request: %s", new_request.request_id
-                    )
-                    new_request.structured_output_request = None
-                    new_request.status = RequestStatus.WAITING
 
                 logger.debug(
                     "Scheduling a new request (%d prompt tokens), holding back %d requests",
@@ -300,6 +291,8 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
             "Ongoing prefill requests must be in the running queue."
         )
 
+        new_prefill_candidates: list[Request] = []
+
         # Check ongoing prefills
         if self.ongoing_prefills:
             # Some running requests are currently being prefilled. We need to
@@ -322,17 +315,45 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
 
         # Check new requests to prefill
         elif len(self.waiting) > 0:
-            self.ongoing_prefills.extend(self.waiting)
-            # Hide current decodes from the scheduler
-            running_holdback = self.running
-            self.running = []
-            self.previous_step_was_prefill = True
+            # Try to promote grammar-waiting requests whose FSM is now
+            # ready, so we correctly classify ready vs not-ready requests.
+            for r in list(self.waiting):
+                if r.status == RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR:
+                    so_req = r.structured_output_request
+                    if so_req and so_req.grammar:
+                        r.status = RequestStatus.WAITING
+
+            ready_to_prefill = [
+                r
+                for r in self.waiting
+                if r.status != RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR  # type: ignore[attr-defined]
+            ]
+            if ready_to_prefill:
+                new_prefill_candidates = list(self.waiting)
+                # Hide current decodes from the scheduler
+                running_holdback = self.running
+                self.running = []
+                self.previous_step_was_prefill = True
+            else:
+                # Grammar not yet initialized for any waiting request.
+                # Return them to holdback so the base scheduler doesn't
+                # try to promote and schedule them alongside decodes.
+                while self.waiting:
+                    holdback_queue.appendleft(self.waiting.pop())
+                running_holdback = []
+                self.previous_step_was_prefill = False
         else:
             self.previous_step_was_prefill = False
             running_holdback = []
 
         # delegate to super of SpyreScheduler: base V1 Scheduler
         outputs = super(SpyreScheduler, self).schedule()
+
+        # Track as ongoing prefills only the requests that were actually
+        # scheduled (i.e., moved from waiting to running by the base
+        # scheduler).
+        if new_prefill_candidates:
+            self.ongoing_prefills.extend(r for r in new_prefill_candidates if r in self.running)
 
         # restore holdbacks after running the base scheduler
         self.running = self.running + running_holdback
@@ -345,6 +366,15 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
             r.num_computed_tokens <= r.num_prompt_tokens + 1 for r in self.running
         ):
             logger.debug("Scheduled tokens in this step: %s", outputs.num_scheduled_tokens)
+
+        # Collect grammar bitmask synchronously for structured outputs.
+        # NOTE: This is done here because vllm-spyre currently combines token sampling
+        # in model_executor.execute_model() rather than implementing sample_tokens()
+        # in the model runner. This means we cannot collect the grammar bitmask
+        # asynchronously while the model is running (as done in vLLM core).
+        # TODO: Implement sample_tokens() in SpyreModelRunner to enable async grammar
+        # collection for better performance.
+        outputs._spyre_grammar_output = self.get_grammar_bitmask(outputs)  # type: ignore[attr-defined]
         return outputs
 
     def can_schedule_prefill(self, request: Request) -> bool:
--- a/tests/e2e/test_structured_outputs.py
+++ b/tests/e2e/test_structured_outputs.py
@@ -1,0 +1,279 @@
+"""End-to-end tests for structured output decoding.
+
+Tests structured output support across different backends (guidance, xgrammar, outlines)
+and ensures that prompts without structured output requests don't accidentally have them applied.
+"""
+
+import json
+import pytest
+import re
+from llm_cache import get_llm
+from spyre_util import ModelInfo
+from vllm import SamplingParams
+from vllm.sampling_params import StructuredOutputsParams
+from vllm.config import StructuredOutputsConfig
+
+pytestmark = [pytest.mark.chunked_prefill]
+
+
+# Parametrize all tests over the three structured output backends
+# Note: Backend support varies by feature:
+# - guidance: supports json_object, json (schema), regex, choice
+# - xgrammar: supports json_object, json (schema), regex, choice
+# - outlines: supports json (schema), regex, choice (NOT json_object)
+STRUCTURED_OUTPUT_BACKENDS = ["guidance", "xgrammar", "outlines"]
+
+# Backends that support json_object (free-form JSON without schema)
+JSON_OBJECT_BACKENDS = ["guidance", "xgrammar"]  # outlines requires schema
+
+
+@pytest.mark.parametrize("structured_output_backend", JSON_OBJECT_BACKENDS)
+def test_structured_output_json_object(
+    model: ModelInfo,
+    backend,
+    monkeypatch,
+    max_model_len,
+    max_num_seqs,
+    max_num_batched_tokens,
+    use_llm_cache,
+    structured_output_backend: str,
+):
+    """Test that structured output with json_object=True produces valid JSON."""
+    spyre_model = get_llm(
+        model=model,
+        max_model_len=max_model_len,
+        backend=backend,
+        monkeypatch=monkeypatch,
+        max_num_seqs=max_num_seqs,
+        max_num_batched_tokens=max_num_batched_tokens,
+        use_pc=True,
+        cached=True,
+        structured_outputs_config=StructuredOutputsConfig(backend=structured_output_backend),
+    )
+
+    prompt = "Generate a JSON object with name and age fields for a person."
+
+    params = SamplingParams(
+        temperature=0.0,
+        max_tokens=50,
+        structured_outputs=StructuredOutputsParams(json_object=True),
+    )
+
+    outputs = spyre_model.generate([prompt], [params])
+    output_text = outputs[0].outputs[0].text
+
+    # Verify output is valid JSON
+    try:
+        json_obj = json.loads(output_text)
+        assert isinstance(json_obj, dict), "Output should be a JSON object"
+    except json.JSONDecodeError as e:
+        pytest.fail(f"Output is not valid JSON: {output_text}\nError: {e}")
+
+
+@pytest.mark.parametrize("structured_output_backend", STRUCTURED_OUTPUT_BACKENDS)
+def test_structured_output_json_schema(
+    model: ModelInfo,
+    backend,
+    monkeypatch,
+    max_model_len,
+    max_num_seqs,
+    max_num_batched_tokens,
+    use_llm_cache,
+    structured_output_backend: str,
+):
+    """Test that structured output with a JSON schema validates correctly."""
+    spyre_model = get_llm(
+        model=model,
+        max_model_len=max_model_len,
+        backend=backend,
+        monkeypatch=monkeypatch,
+        max_num_seqs=max_num_seqs,
+        max_num_batched_tokens=max_num_batched_tokens,
+        use_pc=True,
+        cached=True,
+        structured_outputs_config=StructuredOutputsConfig(backend=structured_output_backend),
+    )
+
+    schema = {
+        "type": "object",
+        "properties": {
+            "name": {"type": "string"},
+            "age": {"type": "integer"},
+        },
+        "required": ["name", "age"],
+        "additionalProperties": False,
+    }
+
+    prompt = "Generate a person with name and age only."
+
+    params = SamplingParams(
+        temperature=0.0,
+        max_tokens=50,
+        structured_outputs=StructuredOutputsParams(json=schema),
+    )
+
+    outputs = spyre_model.generate([prompt], [params])
+    output_text = outputs[0].outputs[0].text
+
+    # Verify output is valid JSON matching the schema
+    try:
+        json_obj = json.loads(output_text)
+        assert isinstance(json_obj, dict), "Output should be a JSON object"
+        assert "name" in json_obj, "Output should have 'name' field"
+        assert "age" in json_obj, "Output should have 'age' field"
+        assert isinstance(json_obj["name"], str), "'name' should be a string"
+        assert isinstance(json_obj["age"], int), "'age' should be an integer"
+    except json.JSONDecodeError as e:
+        pytest.fail(f"Output is not valid JSON: {output_text}\nError: {e}")
+
+
+@pytest.mark.parametrize("structured_output_backend", STRUCTURED_OUTPUT_BACKENDS)
+def test_structured_output_regex(
+    model: ModelInfo,
+    backend,
+    monkeypatch,
+    max_model_len,
+    max_num_seqs,
+    max_num_batched_tokens,
+    use_llm_cache,
+    structured_output_backend: str,
+):
+    """Test that structured output with regex pattern is enforced."""
+    spyre_model = get_llm(
+        model=model,
+        max_model_len=max_model_len,
+        backend=backend,
+        monkeypatch=monkeypatch,
+        max_num_seqs=max_num_seqs,
+        max_num_batched_tokens=max_num_batched_tokens,
+        use_pc=True,
+        cached=True,
+        structured_outputs_config=StructuredOutputsConfig(backend=structured_output_backend),
+    )
+
+    # Regex for phone number format: XXX-XXX-XXXX
+    phone_regex = r"\d{3}-\d{3}-\d{4}"
+
+    prompt = "Generate a phone number in XXX-XXX-XXXX format."
+
+    params = SamplingParams(
+        temperature=0.0,
+        max_tokens=20,
+        structured_outputs=StructuredOutputsParams(regex=phone_regex),
+    )
+
+    outputs = spyre_model.generate([prompt], [params])
+    output_text = outputs[0].outputs[0].text.strip()
+
+    # Verify output matches the regex pattern
+    match = re.fullmatch(phone_regex, output_text)
+    assert match is not None, f"Output '{output_text}' does not match regex pattern '{phone_regex}'"
+
+
+@pytest.mark.parametrize("structured_output_backend", JSON_OBJECT_BACKENDS)
+def test_structured_output_mixed_batch(
+    model: ModelInfo,
+    backend,
+    monkeypatch,
+    max_model_len,
+    max_num_seqs,
+    max_num_batched_tokens,
+    use_llm_cache,
+    structured_output_backend: str,
+):
+    """Test that requests with and without structured outputs can coexist.
+
+    This is critical to ensure that prompts without structured output requests don't
+    accidentally have structured outputs applied. This test submits both structured
+    and non-structured requests in a single batch to verify they can run together.
+    """
+    spyre_model = get_llm(
+        model=model,
+        max_model_len=max_model_len,
+        backend=backend,
+        monkeypatch=monkeypatch,
+        max_num_seqs=max_num_seqs,
+        max_num_batched_tokens=max_num_batched_tokens,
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

File: vllm_spyre/v1/core/scheduler.py [hunk 1]
```diff
@@ -320,20 +305,43 @@ def schedule(self) -> "SchedulerOutput":
...
-            self.ongoing_prefills.extend(self.waiting)
-            # Hide current decodes from the scheduler
-            running_holdback = self.running
-            self.running = []
-            self.previous_step_was_prefill = True
+            # Separate requests that are ready to prefill from those waiting
+            # for grammar FSM initialization (WAITING_FOR_FSM). Only
+            # ready requests should hide the decode batch.
+            ready_to_prefill = [
+                r for r in self.waiting if r.status != RequestStatus.WAITING_FOR_FSM
```
rafvasq: `RequestStatus.WAITING_FOR_FSM` was renamed in v0.19.0: https://github.com/vllm-project/vllm/pull/38048

tests/v1/core/test_scheduler_structured_outputs.py
rafvasq, line 8: Seems like this docstring should be updated too

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -1469,6 +1415,42 @@ def is_cached_chunk(self, scheduler_output: SchedulerOutput):
...
+        output requests.
+        """
+        grammar_output = getattr(scheduler_output, "_spyre_grammar_output", None)
+        if grammar_output is None:
+            return
+
+        ## llguidance is not supported on s390x due to endianness issues.
+        if platform.machine() == "s390x":
+            backend = self.vllm_config.structured_outputs_config.backend
+            if backend == "guidance":
```
joerunde: Should this return an error to the user instead?

Before this PR, we always silently ignored structured output requests since we didn't support it at all and didn't want to break tool calling integrations that sometimes requested it. But now that we do support structured output, it might be better to return errors where it's misconfigured so that users know to go fix their deployments. Otherwise we'll be in a mixed state where sometimes structured output works, and sometimes it's not applied.

R3hankhan123: Now raising `RuntimeError` instead of warning

tests/v1/core/test_scheduler_structured_outputs.py
R3hankhan123: Updated the docstring

joerunde:
bot:test

vllm_spyre/v1/worker/spyre_model_runner.py
joerunde: Oh- @R3hankhan123 I'm not sure this can be done here in the model runner though. Any assertions that happen here generally crash the server. We would need to return an error for this request only, and the easiest way to do this is to validate the request before it even hits the engine, which we can do in `SpyrePlatform.validate_request`

joerunde (changes requested):
A few things we need to clean up before merging: 
 
- We should probably check for the existence of llguidance rather than doing a platform check on s390x, so that we can be forward-compatible if a version of guidance is released that works on Z
- We should validate the structured outputs config up-front when we validate the request in platform.py to avoid setting the guidance backend when llguidance is not installed
- We should validate that the model is not deployed with guidance as the default structured output backend if llguidance is not installed
- We should add tests for these validation cases, as well as one small test that calls a model with a structured output option so we can ensure it actually works

R3hankhan123:
@joerunde llgudiance version 1.7.3 contains the fix and i have added it in the overrides of pyproject, can you take a look thanks

vllm_spyre/v1/worker/spyre_model_runner.py
```diff
@@ -714,6 +717,7 @@ def __init__(
         self.chunk_blocks_count = self.chunk_size // self.block_size
 
         self.prefix_cache_stats = None
+        self._unsupported_backend_logged = False
```
joerunde: is this unused?

sendnn_inference/v1/core/scheduler.py
```diff
@@ -345,6 +353,8 @@ def schedule(self) -> "SchedulerOutput":
             r.num_computed_tokens <= r.num_prompt_tokens + 1 for r in self.running
         ):
             logger.debug("Scheduled tokens in this step: %s", outputs.num_scheduled_tokens)
+
+        outputs._spyre_grammar_output = self.get_grammar_bitmask(outputs)  # type: ignore[attr-defined]
```
joerunde, line 377: It looks to me like this is here because we combine the token sampling in `model_executor.execute_model`, and if we were to implement `sample_tokens` in the model runner then we wouldn't have to do this. 

The engine core runs this:
```
        scheduler_output = self.scheduler.schedule()
        future = self.model_executor.execute_model(scheduler_output, non_block=True)
        grammar_output = self.scheduler.get_grammar_bitmask(scheduler_output)

        model_output = future.result()
        if model_output is None:
            model_output = self.model_executor.sample_tokens(grammar_output)
```

The important thing that I see here is that this collects the grammar bitmask asynchronously while the model is running, which seems like something that we should be doing. 

Assuming the above is correct, then:
1. At a minimum we need to put some comments here explaining this is what we're doing
2. I have a heavy preference for actually implementing `sample_tokens()` so that we can take advantage of the performance benefit

Given the time constraints of releasing this feature for `sendnn_inference==2.0.0`, we could just add some comments here and open an issue to follow up

pyproject.toml
```diff
@@ -79,6 +79,9 @@ override-dependencies = [
     # vllm 0.18.0 pins compressed-tensors==0.13.0; override to 0.14.0.1
     # TODO: remove once minimum vllm is bumped past 0.18.0
     "compressed-tensors==0.14.0.1",
+
+    ## llguidance fix for s390x from version 1.7.3
```
joerunde: we should add a comment that this conflicts with vllm's version range, and that we can remove this override once vllm's requirement range includes or moves past 1.7.3: https://github.com/vllm-project/vllm/blob/v0.19.1/requirements/common.txt#L22

joerunde:
@R3hankhan123 let's update the PR description to remove the bit about llguidance not supported on s390x

Also before merging we still need to encode the examples from the description into unit tests.  We should probably flex all the different backends, and be sure that a prompt in the batch that doesn't request structured outputs does not accidentally have structured outputs applied. See `test_sampling_params.py` for examples of testing other sampling params, that should be a good fit for this test.

sendnn_inference/v1/core/scheduler.py
R3hankhan123: Added the comment and TODO

tests/e2e/test_structured_outputs.py
```diff
@@ -0,0 +1,355 @@
...
+
+@pytest.mark.parametrize("structured_output_backend", STRUCTURED_OUTPUT_BACKENDS)
+def test_structured_output_regex(
+    model: ModelInfo,
+    backend,
+    monkeypatch,
+    max_model_len,
+    max_num_seqs,
+    max_num_batched_tokens,
+    use_llm_cache,
```
joerunde, line 134: If we include this fixture then it assumes we're using a cached LLM so it won't clear the currently cached model. This will crash when we try to load a second one onto the spyre card(s), because we're creating our own model in this test.

joerunde: (this applies to all the tests in this file)

tests/e2e/test_structured_outputs.py
```diff
@@ -0,0 +1,355 @@
...
+    outputs = spyre_model.generate([prompt], [params])
+    output_text = outputs[0].outputs[0].text.strip().lower()
+
+    # Verify output is one of the allowed choices
+    assert output_text in choices, f"Output '{output_text}' not in allowed choices {choices}"
+
+
+@pytest.mark.parametrize("dynamo_backend", ["sendnn", "eager"])
+@pytest.mark.parametrize("structured_output_backend", JSON_OBJECT_BACKENDS)
+def test_structured_output_all_backends(
```
joerunde: Tests are already always parameterized with the backend as both `sendnn` and `eager`- we can delete this test

tests/e2e/test_structured_outputs.py
```diff
@@ -0,0 +1,355 @@
...
+        max_num_seqs,
+        max_num_batched_tokens,
+        structured_output_backend,
+    )
+
+    prompt = "Generate a JSON object with name and age fields for a person."
+
+    params = SamplingParams(
+        temperature=0.0,
+        max_tokens=200,  # Increased to ensure JSON completion
```
joerunde: 🤔 `max_tokens=200` could increase the test runtime quite a bit if this decides to generate a long json string

tests/e2e/test_structured_outputs.py
R3hankhan123: I have removed `use_llm_cache`

tests/e2e/test_structured_outputs.py
```diff
@@ -0,0 +1,355 @@
...
+        structured_outputs=StructuredOutputsParams(json_object=True),
+    )
+
+    outputs = spyre_model.generate([prompt], [params])
+    output_text = outputs[0].outputs[0].text
+
+    # Verify output is valid JSON
+    try:
+        json_obj = json.loads(output_text)
+        assert isinstance(json_obj, dict), "Output should be a JSON object"
```
joerunde, line 66: I'm skeptical that the micro models we use for testing will actually always generate an object- but if this passes on spyre for both fp16 and fp8 then it's probably okay

tests/e2e/test_structured_outputs.py
R3hankhan123: Removed the test

tests/e2e/test_structured_outputs.py
```diff
@@ -0,0 +1,355 @@
...
+
+    # Request without structured output (free-form text)
+    prompt_freeform = "Write a short story about a cat."
+    params_freeform = SamplingParams(
+        temperature=0.0,
+        max_tokens=50,
+    )
+
+    # Generate with structured output
+    output_structured = spyre_model.generate([prompt_structured], [params_structured])[0]
```
joerunde: these need to be a single request, like

```
output = spyre_model.generate([prompt_structured, prompt_freeform], [params_structured, params_freeform])
```

the point is to ensure that requests that are running within the same batch can have different structured output setttings

tests/e2e/test_structured_outputs.py
R3hankhan123: Reduced the number of tokens to 50

tests/e2e/test_structured_outputs.py
```diff
@@ -0,0 +1,355 @@
...
+    output_text = outputs[0].outputs[0].text
+
+    # Verify output is valid JSON matching the schema
+    try:
+        json_obj = json.loads(output_text)
+        assert isinstance(json_obj, dict), "Output should be a JSON object"
+        assert "name" in json_obj, "Output should have 'name' field"
+        assert "age" in json_obj, "Output should have 'age' field"
+        assert isinstance(json_obj["name"], str), "'name' should be a string"
+        assert isinstance(json_obj["age"], int), "'age' should be an integer"
```
joerunde, line 121: very nice 🌶️

tests/e2e/test_structured_outputs.py
```diff
@@ -0,0 +1,355 @@
+"""End-to-end tests for structured output decoding.
```
joerunde, line 1: The tests here in general are pretty great!

There's one big problem here though- we're not using cached models which will really slow down the testing. On spyre if we don't use the model cache then it takes ~30s to reload a new model (for g3.3-micro that we test with). The choices here are either:

1. Consolidate all the cases into a single test so we load the model once with each backend and send a few batches of prompts to it
2. Add the `structured_output_backend` as a model cache key and use the cached llm with that backend

I'd prefer (2), it should be pretty quick to get that hooked up in the model cache. In our CI runs we also have the benefit of running all the tests with a cached `LLM` in a single process instead of forking the tests which is much faster as well

tests/e2e/test_structured_outputs.py
R3hankhan123: Done

tests/e2e/test_structured_outputs.py
R3hankhan123: @joerunde i have added code for option 2, can you review it and let me know if its the correct approach, thanks

tests/e2e/test_structured_outputs.py
R3hankhan123: @joerunde using this e2e tests is failing with
```
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]            ^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/usr/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     raise self._exception
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/home/runner/work/sendnn-inference/sendnn-inference/.venv/lib/python3.12/site-packages/vllm/v1/executor/uniproc_executor.py", line 84, in collective_rpc
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     result = run_method(self.driver_worker, method, args, kwargs)
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/home/runner/work/sendnn-inference/sendnn-inference/.venv/lib/python3.12/site-packages/vllm/v1/serial_utils.py", line 510, in run_method
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     return func(*args, **kwargs)
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]            ^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/home/runner/work/sendnn-inference/sendnn-inference/.venv/lib/python3.12/site-packages/vllm/v1/worker/worker_base.py", line 332, in execute_model
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     return self.worker.execute_model(scheduler_output)
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/home/runner/work/sendnn-inference/sendnn-inference/.venv/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     return func(*args, **kwargs)
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]            ^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/home/runner/work/sendnn-inference/sendnn-inference/.venv/lib/python3.12/site-packages/vllm_spyre/v1/worker/spyre_worker.py", line 770, in execute_model
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     output = self.model_runner.execute_model(scheduler_output)
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/home/runner/work/sendnn-inference/sendnn-inference/.venv/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     return func(*args, **kwargs)
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]            ^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/home/runner/work/sendnn-inference/sendnn-inference/.venv/lib/python3.12/site-packages/vllm_spyre/v1/worker/spyre_model_runner.py", line 1541, in execute_model
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     self.maybe_setup_new_prefill(scheduler_output)
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]   File "/home/runner/work/sendnn-inference/sendnn-inference/.venv/lib/python3.12/site-packages/vllm_spyre/v1/worker/spyre_model_runner.py", line 1478, in maybe_setup_new_prefill
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]     assert len(scheduler_output.scheduled_new_reqs) == 1, (
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=3614) ERROR 04-22 17:44:37 [core.py:1110] AssertionError: Can only schedule one chunked prefill at a time
```

sendnn_inference/v1/core/scheduler.py
joerunde: followup issue opened here: https://github.com/torch-spyre/sendnn-inference/issues/947

tests/e2e/test_structured_outputs.py
joerunde: @R3hankhan123 there's a sort key class that probably needs to be updated as well, it needs to sort these test cases so that cases with identical LLMs are grouped together

joerunde:
It also looks like there's a test failure with the mixed batch:
```
AssertionError: Can only schedule one chunked prefill at a time
```

so something is wrong and we've violated the constraint that we can only prefill a single request at a time

R3hankhan123:
> It also looks like there's a test failure with the mixed batch:
> 
> ```
> AssertionError: Can only schedule one chunked prefill at a time
> ```
> 
> so something is wrong and we've violated the constraint that we can only prefill a single request at a time

@joerunde this change has caused the test to fail so i have reverted it for now
https://github.com/torch-spyre/sendnn-inference/pull/903#discussion_r3125671829

joerunde:
> this change has caused the test to fail so i have reverted it for now

@R3hankhan123 don't be like Bob and delete the tests lol

This cannot merge without that working, we obviously _must_ support mixed batches where some prompts requests structured output and others don't. Otherwise all users would always need to deploy with `--max-num-seqs 1` and run everything serially to avoid that bug.

Can you prioritize fixing this bug? Otherwise structured output support won't make the 2.0 cut this week

R3hankhan123:
> > this change has caused the test to fail so i have reverted it for now
> 
> @R3hankhan123 don't be like Bob and delete the tests lol
> 
> This cannot merge without that working, we obviously _must_ support mixed batches where some prompts requests structured output and others don't. Otherwise all users would always need to deploy with `--max-num-seqs 1` and run everything serially to avoid that bug.
> 
> Can you prioritize fixing this bug? Otherwise structured output support won't make the 2.0 cut this week

I will dig around the code and see how to fix it

R3hankhan123:
@joerunde i have resolved the bug, so the scheduler's holdback mechanism only controlled `self.waiting`, but grammar waiting requests leaked into `self.skipped_waiting` where they bypassed the one prefill at a time constraint entirely

vllm_spyre/v1/core/scheduler.py
```diff
@@ -0,0 +1,584 @@
+# SPDX-License-Identifier: Apache-2.0
+
```
joerunde: this file should not exist

R3hankhan123: looks like one file didnt get deleted during the rebase, now have removed it

joerunde:
@R3hankhan123 I've added you (and @GOavi101) to the allow list to trigger the spyre-ci jobs from your PRs. See the docs for that here: https://docs.vllm.ai/projects/spyre/en/latest/contributing/maintaining.html#testing-changes-on-spyre-hardware

You can trigger a full run with just `bot:test`, though I usually split out the markers at least with `spyre and not quantized` and `spyre and quantized` separately so that the runs don't take quite as long. You can start with just your test file too with `TEST_FILE=tests/e2e/test_structured_outputs.py`.

We'll just need to see the passing spyre-ci tests before we merge this PR

R3hankhan123:
bot:test
 TEST_FILE=tests/e2e/test_structured_outputs.py

R3hankhan123:
> bot:test TEST_FILE=tests/e2e/test_structured_outputs.py

@joerunde did the CI start or is something wrong with the bot command i have written?

joerunde:
hmmm, you should be on the list 🤔 

let me take a look

joerunde:
@R3hankhan123 it looks like there was a leading space in your comment:

```
    "body": " bot:test\r\n TEST_FILE=tests/e2e/test_structured_outputs.py",
```

looks like our trigger was `^bot:test.*`, but I can change it to `.*bot:test.*`. Mind trying again without the leading space though?

R3hankhan123:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

R3hankhan123:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

R3hankhan123:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

R3hankhan123:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

joerunde:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

joerunde:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

joerunde:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

joerunde:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

joerunde:
bot:test
TEST_FILE=tests/e2e/test_structured_outputs.py

joerunde:
bot:test
MARKERS="spyre and not quantized and not multi and chunked_prefill"

joerunde:
There was a small issue with the test ordering when running all the tests together but otherwise everything is passing

joerunde (approved):
good luck us!

PR #905, opened 2026-04-08
:zap: don't set KV cache on model

joerunde:
## Description

Based on recent profiling that saw ~1ms being spent on a `toch.nn.Module.setattr` call directly after the forward pass that we tracked down to setting `self.past_key_value_states`, this PR updates the model wrapper to ignore the kv cache "outputs" from the model.

These tensors don't contain the actual kv cache data, so we shouldn't waste time assigning them back to the model

## Test Plan

perf benchmarking to validate the speedup.

## Checklist

### DIFF
--- a/vllm_spyre/model_executor/model_loader/spyre.py
+++ b/vllm_spyre/model_executor/model_loader/spyre.py
@@ -62,9 +62,8 @@ class SpyreCausalLM(nn.Module):
         # number of right pads
         self.n_pads_right = 0
 
-        self._mask_dtype = (
-            torch.float16 if SpyrePlatform.is_backend_sendnn_enabled() else torch.float32
-        )
+        self.on_spyre = SpyrePlatform.is_backend_sendnn_enabled()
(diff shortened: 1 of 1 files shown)


joerunde:
bot:test
MARKERS="spyre and basic"

joerunde:
bot:test
MARKERS="spyre and prefix_caching"

vllm_spyre/model_executor/model_loader/spyre.py
```diff
@@ -141,6 +141,8 @@ def __init__(
             tuple[torch.Tensor | ScaledTensor, torch.Tensor | ScaledTensor]
         ] = []
 
+        self.on_spyre = SpyrePlatform.is_backend_sendnn_enabled()
```
yannicks1: nit: if you set this before self.load_weights() call, you can reuse it in line 97 :)

yannicks1 (approved):
LGTM, if all the tests pass -> ship it!

joerunde:
bot:test

PR #906, opened 2026-04-09
[Spyre-Next] Simplify SiluAndMul by removing custom op barrier

GOavi101:
## Description

This PR simplifies the `SpyreSiluAndMul` implementation by removing the custom op barrier pattern.

### Changes
- Remove opaque custom-op boundary in SpyreSiluAndMul
- Simplify forward_oot() to compute directly without layer registry
- Add .contiguous() calls to ensure zero storage offset for Spyre

### Technical Details

**Background:**
The split cannot happen on Spyre inside a compiled graph — Spyre's Flex backend rejects non-zero storage offsets (aten.slice.Tensor unsupported), confirmed with a standalone repro script (torch-spyre/torch-spyre#1333)

**Implementation:**
- Split input tensor with `.contiguous()` to ensure zero storage offset
- Compute `silu(x1) * x2` on Spyre device
- Convert result back to original device/dtype

## Related Issues

Related to https://github.com/vllm-project/vllm-spyre/issues/887

## Test Plan

Tested on `ibm-granite/granite-3.3-8b-instruct`:
- Warmup completes successfully
- Inference produces correct output

## Checklist

### DIFF
--- a/vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
+++ b/vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
@@ -8,19 +8,17 @@ vllm/model_executor/layers/activation.py when instantiated.
 
 Architecture:
     - OOT Registration: @SiluAndMul.register_oot() replaces upstream at instantiation
-    - forward_oot(): Entry point for OOT dispatch, calls custom op for
-      torch.compile opacity
-    - Custom Op Boundary: torch.ops.vllm.spyre_siluandmul is opaque to torch.compile,
-      so _forward_spyre_impl runs eagerly outside the compiled graph
-    - Separate Compilation: forward_spyre is compiled independently via maybe_compile
+    - forward_oot(): Entry point for OOT dispatch, fully transparent to the outer
+      torch.compile graph (no opaque custom-op boundary).
+    - Halves use .contiguous() to ensure zero storage offset before device transfer.
+    - convert() utility handles device/dtype transfers efficiently.
 
 Spyre Device Constraints:
-    - Computations performed in torch.float16:
-      Input (dtype defined by model / user) converted to torch.float16 for
-      operations on spyre and then converted back to original dtype for cpu.
+    - Splitting (aten.slice.Tensor) inside compiled Spyre graphs is unsupported —
+      non-zero storage offsets are rejected by the Flex backend.
+    - The .contiguous() call ensures zero storage offset before transfer to Spyre.
 
 Output Shape Note:
-    Unlike RMSNorm (same input/output shape), SiluAndMul halves the last dimension:
     input shape: [..., 2*d] -> output shape: [..., d]
 
 References:
@@ -31,11 +29,9 @@ import torch
 import torch.nn.functional as F
 
 from vllm.logger import init_logger
-from vllm.utils.torch_utils import direct_register_custom_op
 from vllm.model_executor.layers.activation import SiluAndMul
-from functools import lru_cache
 
-from .utils import convert, register_layer, get_layer, _fake_impl
+from .utils import convert
 
 logger = init_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
```diff
@@ -44,134 +41,60 @@
...
+        requiring layer registry or custom op registration.
         """
         super().__init__(*args, **kwargs)
-
         logger.debug("Building custom SiluAndMul")
-
         self._target_device = torch.device("spyre")
         self._target_dtype = torch.float16
-        self.maybe_compiled_forward_spyre = self.maybe_compile(self.forward_spyre)
-
```
bohnstingl, line 67: So we don't need this anymore and this will be triggered by the upstream implementation through `enforce_eager`, right? Can you check the difference when using `enforce_eager=True` and without?

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
```diff
@@ -44,134 +41,60 @@
...
-        self._layer_name = register_layer(self, "spyre_siluandmul")
-
-        logger.debug_once(
-            "SpyreSiluAndMul: Dispatch: enabled=%s, Forward method=%s, Compiled=%s",
-            self.enabled(),
-            self._forward_method.__name__,
-            self.maybe_compiled_forward_spyre is not self.forward_spyre,
-        )
```
bohnstingl, line 75: Would a logging like this not make sense anymore? Like, the CustomOp can still be `enabled` or not, right?

bohnstingl (changes requested):
Looks good to me in principle.

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
GOavi101: Good point! The old logging referenced attributes that no longer exist (_forward_method,maybe_compiled_forward_spyre ). In the simplified implementation, we could add simpler logging like:
```
logger.debug_once(
    "SpyreSiluAndMul: enabled=%s",
    self.enabled(),
)
```
something like this? It won't show compilation status since that's now handled by the outer graph, but it would still indicate if the CustomOp is active.

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
GOavi101: sure I will run a test comparing the two modes

GOavi101: ### I tested both modes to verify the behavior:

### With enforce_eager=True (eager execution):
- Compilation mode: NONE
- Custom ops: 'all'
- Direct eager execution through **forward_oot()**

### With enforce_eager=False (compilation enabled):
- Compilation mode: DYNAMO_TRACE_ONCE
- Custom ops: 'none'
- torch.compile wraps the entire model including SpyreSiluAndMul
- Longer warmup (130s vs 68s) due to compilation overhead
-  ~2x faster inference (92s vs 180s)

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
```diff
@@ -44,134 +41,64 @@
...
+        zero storage offset (Spyre's Flex backend rejects non-zero offsets).
 
-        Computes silu(x1) * x2 on the Spyre device, relying on torch-spyre's
-        registered aten::silu.out kernel. The two halves are passed in as
-        separate tensors because the Spyre device does not yet support tensor
-        slicing (strided views); the split is therefore performed on CPU before
-        this method is called (see _forward_spyre_impl).
+        The convert() utility handles device/dtype transfers efficiently,
+        skipping data movement when the tensor is already on the correct
+        device/dtype (vllm-spyre#863 optimization).
```
bringlein: so we have a dependency on https://github.com/vllm-project/vllm-spyre/pull/863? Or can we merge it independently?

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
```diff
@@ -44,134 +41,64 @@
...
-            convert(x1, self._target_device, self._target_dtype),
-            convert(x2, self._target_device, self._target_dtype),
-        )
-
-        # Transfer back to original device and restore original dtype
-        return convert(out, x_device, x_dtype)
+        # Call .contiguous() to ensure zero storage offset (Spyre's Flex backend
+        # rejects non-zero offsets). convert() then transfers to Spyre device/dtype,
+        # skipping transfer if already correct (vllm-spyre#863 optimization).
+        x1 = convert(x[..., :d].contiguous(), self._target_device, self._target_dtype)
```
bringlein, line 90: this looks like an elegant solution. But out of curiosity, do we know how these slicing is handled by torch-spyre?

bringlein (commented):
Thanks @GOavi101 , looks good. 

I have two questions. But I think if there is no dependency, we should be good to merge.

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
GOavi101: The convert()utility function already exists and works correctly. PR #863 added an optimization to convert() that skips redundant data movement when a tensor is already on the correct device/dtype, but it's not a hard dependency.

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
GOavi101: The slicing happens on the CPU before transfer to Spyre.

The split cannot happen on Spyre inside a compiled graph — Spyre's Flex backend rejects non-zero storage offsets (aten.slice.Tensor unsupported)(https://github.com/torch-spyre/torch-spyre/issues/1333)

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
bringlein: ok, make sense. Can we then just remove the comment that links it to PR 863?

vllm_spyre_next/vllm_spyre_next/custom_ops/silu_and_mul.py
bringlein: ok, but if the preceding op is also executed on spyre, how should this work? (but maybe out-of-scope for this PR)

GOavi101: Spyre tensor → CPU (convert detects device mismatch) → slice on CPU → contiguous → back to Spyre

it isn't optimal, but it's the current workaround for Spyre's **aten.slice.Tensor** limitation.

bringlein (approved):
LGTM

bohnstingl:
I think we can close this PR in favor of https://github.com/torch-spyre/spyre-inference/pull/148?

PR #913, opened 2026-04-13
Exceed TKV Limit Unit Test and Bugfix

Daniel-Schenker:
## Description

This PR adds a unit test case to reproduce [redacted]
A potential fix for the bug is also included in this PR.

## Related Issues

Relates to [redacted]

## Test Plan

Utilize existing and newly added vllm-spyre unit tests

## Checklist

### DIFF
--- a/tests/v1/worker/test_scheduler_tkv_limits.py
+++ b/tests/v1/worker/test_scheduler_tkv_limits.py
@@ -1,0 +1,96 @@
+import pytest
+from scheduling_utils import create_request_for_scheduler_test, random_prompt
+from v1.worker.mock_model import InstrumentedModelRunner
+from spyre_util import REFERENCE_MODELS
+from vllm_spyre.platform import SpyrePlatform
+
+@pytest.mark.cpu
+@pytest.mark.chunked_prefill
+def test_batch_tkv_padding_alignment_bug(
+    monkeypatch: pytest.MonkeyPatch,
+):
+    monkeypatch.setenv("VLLM_DT_MAX_BATCH_TKV_LIMIT", "131072")
+
+    model = REFERENCE_MODELS[InstrumentedModelRunner.DEFAULT_TEST_MODEL]
+
+    model_runner = InstrumentedModelRunner.build(
+        monkeypatch=monkeypatch,
+        max_num_batched_tokens=512,
+        max_num_seqs=32,
+        max_model_len=32768,
+        available_blocks=32768,
+    )
+    
+    req0_prompt = random_prompt(model=model, seed=0, length=940)
+    req1_prompt = random_prompt(model=model, seed=1, length=412)
+    req2_prompt = random_prompt(model=model, seed=2, length=969)
+    req3_prompt = random_prompt(model=model, seed=2, length=949)
+    req4_prompt = random_prompt(model=model, seed=2, length=11946)
+    max_tokens = 16384
+    req0 = create_request_for_scheduler_test(
+        model=model,
+        request_id=0,
+        add_step=0,
+        max_tokens=max_tokens,
+        prompt=req0_prompt,
+        use_golden_token_injection=False,
+        generate_hf_results=False,
+    ).request
+
+    req1 = create_request_for_scheduler_test(
+        model=model,
+        request_id=1,
+        add_step=0,
+        max_tokens=max_tokens,
+        prompt=req1_prompt,
+        use_golden_token_injection=False,
+        generate_hf_results=False,
+    ).request
+
+    req2 = create_request_for_scheduler_test(
+        model=model,
+        request_id=2,
+        add_step=0,
+        max_tokens=max_tokens,
+        prompt=req2_prompt,
+        use_golden_token_injection=False,
+        generate_hf_results=False,
+    ).request
+
+    req3 = create_request_for_scheduler_test(
+        model=model,
+        request_id=3,
+        add_step=0,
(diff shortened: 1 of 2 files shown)


joerunde:
bot:test

vllm_spyre/v1/core/scheduler.py
```diff
@@ -505,13 +505,13 @@ def check_batch_tkv_limit_cp(
...
+        dec_req_tkv = max(self.tkv, request.num_prompt_tokens)
         for req in running:
-            # current tkv of the (left aligned) decode sequence
-            dec_req_tkv = (
-                n_blocks + 1
-            ) * self.block_size + req.num_computed_tokens % self.block_size
             n_generated_output_tokens = req.num_computed_tokens - req.num_prompt_tokens
             dec_req_max_tkv = dec_req_tkv + (req.max_tokens - n_generated_output_tokens) - 1
+            dec_req_max_tkv += self.block_size # Account for potential padding block
+            # Do we need to round to next block size here?
```
joerunde: @yannicks1 we could use your input here!

This change is a little more conservative than we were doing before- it removes the rounding up to the next block that was being done before with `n_blocks + 1`, instead adding the worst-case full-block padding after calculating the max tkv in tokens. (This removes a division in the loop so it should also be a touch faster ⚡)

However, what we don't know is if we need to round up to the next full block yet again here. The assertion in the model runner only checks `tkv * batch_size < max_batch_tkv_limit`, but if the real limit on the max batch tkv area needs to be calculated based on the width of the batch in blocks, then we would need to do that extra rounding both in our check in the model runner and in the scheduler here.

I forget what the actual limitation is, or if the compiler actually compiles programs with enough buffer around the tkv limit for this to even matter.

tests/v1/worker/test_scheduler_tkv_limits.py
```diff
@@ -0,0 +1,95 @@
...
+    monkeypatch.setenv("VLLM_DT_MAX_BATCH_TKV_LIMIT", "131072")
+
+    model = REFERENCE_MODELS[InstrumentedModelRunner.DEFAULT_TEST_MODEL]
+
+    model_runner = InstrumentedModelRunner.build(
+        monkeypatch=monkeypatch,
+        max_num_batched_tokens=512,
+        max_num_seqs=32,
+        max_model_len=32768,
+        available_blocks=32768,
```
yannicks1, line 44: I don't see this having an impact here, but maybe set this to 8192 as this is the max number of blocks we currently support? 32K blocks (of 64 tokens) seems like a looot.

tests/v1/worker/test_scheduler_tkv_limits.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    
+    req0_prompt = random_prompt(model=model, seed=0, length=940)
```
yannicks1: lines 24 - 78 can be neatly put into a loop for better readability.

tests/v1/worker/test_scheduler_tkv_limits.py
Daniel-Schenker: So we did actually have `available_blocks=8192` at first... But there was an error stating that 16385 blocks were needed. Maybe @joerunde can comment further on this.

prashantgupta24:
@Daniel-Schenker it looks like we have internal GH links in the PR description. We were warned not to do that

joerunde:
^^ good catch @prashantgupta24!

tests/v1/worker/test_scheduler_tkv_limits.py
joerunde: +1, though we ended up with some startup errors because the max batch tkv limit wasn't being respected, which ended up crashing because we calculated a required 16k blocks 😢 

We should probably understand why the env patch wasn't getting picked up, but that also doesn't need t block getting this in

Daniel-Schenker:
@joerunde @yannicks1 I have cleaned up the test case a bit and added some (hopefully) helpful comments as this is pretty niche test. Let me know if it looks good.

Daniel-Schenker:
I confirmed that the test was still failing without scheduler changes and passing after changes.

tests/v1/worker/test_scheduler_tkv_limits.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Test that the scheduler correctly enforces the TKV limit constraint.
+
+    The test creates 5 requests with varying prompt lengths and a large
+    max_tokens value, then runs the scheduler to completion. With the bug
+    present, the scheduler incorrectly accepts a batch configuration that
+    violates the TKV limit of 131072.
+
+    Expected behavior (when bug is fixed):
+    - Scheduler should reject requests that would cause TKV limit violations
```
joerunde: nit: I'd phrase this more as "Scheduler should hold back the next request in queue until it is guaranteed not to violate TKV limits later". Saying "reject" makes it sound like we reject the request entirely but we're just holding it back a few more iterations

vllm_spyre/v1/core/scheduler.py
joerunde: @Daniel-Schenker let's either remove this comment or update it to mention that the assertion in the model runner does not currently round up to the nearest block, so neither does this

Daniel-Schenker:
Comments have been addressed.

joerunde (approved):
get it!

PR #915, opened 2026-04-14
fix: add profile_prefix param to SpyreWorker.profile()

rafvasq:
## Description

This PR adds the missing `profile_prefix` parameter to `SpyreWorker.profile()` and introduces unit tests to prevent future regressions.

## Test Plan

Added new unit tests in `tests/v1/worker/test_spyre_worker_profile.py` covering starting, stopping, unconfigured, and comparing signatures parameter changes. 

## Checklist

### DIFF
--- a/tests/v1/worker/test_spyre_worker_profile.py
+++ b/tests/v1/worker/test_spyre_worker_profile.py
@@ -1,0 +1,105 @@
+import inspect
+
+import pytest
+from unittest.mock import Mock
+from vllm.config import (
+    VllmConfig,
+    ModelConfig,
+    CacheConfig,
+    ParallelConfig,
+    SchedulerConfig,
+    ProfilerConfig,
+)
+from vllm.v1.worker.gpu_worker import Worker as UpstreamWorker
+from vllm_spyre.v1.worker.spyre_worker import SpyreWorker
+
+
+@pytest.fixture
+def mock_vllm_config():
+    """Create a mock VllmConfig with profiler enabled."""
+    model_config = Mock(spec=ModelConfig)
+    model_config.model = "test-model"
+    model_config.runner_type = "generate"
+    model_config.seed = 0
+    model_config.max_model_len = 2048
+
+    cache_config = Mock(spec=CacheConfig)
+
+    parallel_config = Mock(spec=ParallelConfig)
+    parallel_config.world_size = 1
+    parallel_config.tensor_parallel_size = 1
+    parallel_config.pipeline_parallel_size = 1
+
+    scheduler_config = Mock(spec=SchedulerConfig)
+    scheduler_config.max_num_seqs = 256
+
+    vllm_config = Mock(spec=VllmConfig)
+    vllm_config.model_config = model_config
+    vllm_config.cache_config = cache_config
+    vllm_config.parallel_config = parallel_config
+    vllm_config.scheduler_config = scheduler_config
+    vllm_config.profiler_config = ProfilerConfig(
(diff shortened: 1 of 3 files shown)


vllm_spyre/v1/worker/spyre_worker.py
```diff
@@ -730,7 +730,7 @@ def _warmup_model_forward_pass(
         }
         self.execute_model(scheduler_output)  # Prefill
 
-    def profile(self, is_start: bool = True):
+    def profile(self, is_start: bool = True, profile_prefix: str | None = None):
```
joerunde, line 733: do we need to actually do anything with profile_prefix though?

I didn't dig enough to find out what this is used for upstream and if it's important

tests/v1/worker/test_spyre_worker_profile.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+@pytest.mark.cpu
+@pytest.mark.worker
+def test_profile_start(monkeypatch, mock_vllm_config):
+    """Test that profile() starts the profiler."""
+    worker = _make_worker(monkeypatch, mock_vllm_config)
+    worker.profiler = Mock()
+
+    worker.profile(is_start=True)
```
joerunde, line 68: The root cause of the crash I got was that the engine calls `worker.profile` with its own set of arguments, so I expected us to test this at least one level higher where we let the upstream code call our worker instead of calling it directly.

Coupled with `test_profile_signature_matches_upstream` this is _probably_ fine as is, but I wonder if there's an easy way to build an AsyncLLMEngine with the model mocked out like we do here with `_make_worker` so that we can actually test the integration.

Or... maybe we want to just spin up a server and send requests to start and stop profiling so we can ensure that we actually get profile files saved to disk?  Were you thinking that'd be too heavyweight to do in the unit test suite?

vllm_spyre/v1/worker/spyre_worker.py
rafvasq: It's for custom trace naming, something like `<prefix>_dp<X>_pp<Y>_tp<Z>`, [default naming used otherwise](https://docs.vllm.ai/en/latest/api/vllm/entrypoints/llm/#vllm.entrypoints.llm.LLM.start_profile)

tests/v1/worker/test_spyre_worker_profile.py
rafvasq: I was just keeping it simple to catch changes in the parameters upstream, but I can add an E2E approach and make sure the files get saved.

vllm_spyre/v1/worker/spyre_worker.py
joerunde: Right, and it looks like the `LLM` class accepts that as an arg to `start_profile`, so users of the python api might expect to be able to set it. 

Maybe we should at least log a warning if a prefix is passed and we're not gonna implement it? But any idea if we can just slap the prefix on with the current profiler implementation?

joerunde:
Nice, looks like the e2e test runs as part of the chunked_prefill suite:
```
tests/e2e/test_profiler.py::test_profiler[ibm-ai-platform/micro-g3.3-8b-instruct-1b-max_model_len(512)-max_num_seqs(4)-max_num_batched_tokens(128)] PASSED [  2%]
```

LPGTM- do we want to actually implement the prefix or just merge it?

rafvasq:
@joerunde I think we can for the narrow scope of fixing what's broken. I can follow up with the implementation.

joerunde (approved):

PR #927, opened 2026-04-16
Ministral offline fix

joerunde:
## Description

This PR adds a simple class that sets default values for vllm's CLI args based on other cli args provided by the user.

This fixes a bug where with `HF_HUB_OFFLINE=1`, vllm is unable to determine whether a model is a mistral model or not. This requires users to explicitly pass `--config-format mistral` and `--tokenizer-mode mistral`, which is _not_ required when hf hub is in online mode.

This instead sets the default values for `--config-format` and `--tokenizer-mode` to "mistral" if a `params.json` file is detected. This works if `--model` either:
- References an hf hub model name that currently has a `params.json` cached, like`--model mistralai/Mistral-7B-Instruct-v0.1`
- Points to a directory with `params.json`, like `--model /path/to/my/mounted/model`

Because these are only overriding the default values for the CLI args, any user-specified values are still always respected.

## Checklist

### DIFF
--- a/tests/utils/test_platform_validation.py
+++ b/tests/utils/test_platform_validation.py
@@ -272,6 +272,12 @@ class TestPreRegisterAndUpdate:
         yield
         ConditionalDefaultManager.clear()
 
+        # Re-import huggingface_hub constants to reset any patched env vars
+        import importlib
+        import huggingface_hub.constants
+
+        importlib.reload(huggingface_hub.constants)
+
     @pytest.fixture
     def arg_parsers(self):
         """Create main parser and serve subparser like vLLM's CLI structure.
--- a/tests/utils/test_platform_validation.py
+++ b/tests/utils/test_platform_validation.py
@@ -258,3 +258,73 @@ class TestSendnnConfigurationValidation:
 
         # Verify FLEX_DEVICE was set to COMPILE
         assert os.environ.get("FLEX_DEVICE") == "COMPILE"
+
+
+class TestPreRegisterAndUpdate:
+    """Test SpyrePlatform.pre_register_and_update conditional defaults."""
+
+    @pytest.fixture(autouse=True)
+    def clear_conditional_defaults(self):
+        """Clear conditional defaults before each test to ensure isolation."""
+        from vllm_spyre.argparse_utils import ConditionalDefaultManager
+
+        ConditionalDefaultManager.clear()
+        yield
+        ConditionalDefaultManager.clear()
+
+    def create_test_parser(self):
+        """Create a parser with the arguments used by pre_register_and_update."""
+        from vllm.utils.argparse_utils import FlexibleArgumentParser
+
+        parser = FlexibleArgumentParser()
+        parser.add_argument("--config-format", dest="config_format", default="auto")
+        parser.add_argument("--tokenizer-mode", dest="tokenizer_mode", default="auto")
+        parser.add_argument("model_tag", nargs="?", default=None)
+        return parser
+
+    def test_config_format_defaults_to_mistral_when_model_tag_contains_mistral(self):
+        """Test that config_format defaults to 'mistral' when model_tag contains 'mistral'."""
+        # Parse with a mistral model
+        parser = self.create_test_parser()
+        SpyrePlatform.pre_register_and_update(parser)
+        args = parser.parse_args(["mistralai/Mistral-7B-Instruct-v0.1"])
+
+        assert args.config_format == "mistral"
+        assert args.tokenizer_mode == "mistral"
+
+    def test_config_format_defaults_to_mistral_when_params_json_exists(self, tmp_path):
+        """Test that config_format defaults to 'mistral' when model_tag points to dir with
+        params.json."""
+        # Create a temporary directory with params.json (NB: path does not have mistral in the name)
+        model_dir = tmp_path / "some_model"
+        model_dir.mkdir()
+        params_file = model_dir / "params.json"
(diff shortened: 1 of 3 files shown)


vllm_spyre/platform.py
```diff
@@ -490,12 +495,47 @@ def _get_matching_warmup_shapes(
...
+            not the case as we load mistral models from the chunked safetensors files.
+            """
+            # Check both 'model' and 'model_tag' since vLLM uses different
+            # attribute names in different contexts
+            model = getattr(namespace, "model_tag", None) or getattr(namespace, "model", "") or ""
+            maybe_local_path = Path(model)
+            local_mistral_configs = (
+                maybe_local_path.exists() and (maybe_local_path / "params.json").is_file()
+            )
+            return "mistral" if "mistral" in model.lower() or local_mistral_configs else "auto"
```
tjohnson31415: Instead of including the string match in the check (which can be brittle), is there a function from vLLM we can use to resolve the model name to a path to then check for the file?
Maybe could reuse the `any_pattern_in_repo_files` that is used in `is_mistral_model_repo` [REF](https://github.com/vllm-project/vllm/blob/978a4462bbc529ff204647543526e4caa08ed974/vllm/transformers_utils/repo_utils.py#L157). It takes a `model_name_or_path` so could also remove the file check code from here.

vllm_spyre/argparse_utils.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        Apply the conditional default logic to the parser.
+
+        This method:
+        1. Replaces the action for each managed argument with ConditionalDefaultAction
+        2. Patches the parser's parse_args method to apply conditional defaults
+        """
+        logger.debug(
+            "Enabling conditional defaults with %d config(s)",
+            len(cls._all_conditional_defaults),
```
tjohnson31415: NIT: for debug logs, can just dump the dict? Or at least list all the `dest` values instead of just a count.

vllm_spyre/argparse_utils.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+
+        # Step 1: Replace actions for managed arguments
+        seen_dests: set[str] = set()
+        for config in cls._all_conditional_defaults:
+            dest = config["dest"]
+            if dest in seen_dests:
+                continue
```
tjohnson31415: This looks like it skips duplicates and lets the first one "win", but I think it would be better to throw an error if registering multiple actions for the same dest.

vllm_spyre/argparse_utils.py
```diff
@@ -0,0 +1,208 @@
...
+                            value,
+                        )
+                        setattr(result, dest, value)
+                        setattr(result, applied_attr, True)
+                except Exception as e:
+                    logger.debug(
+                        "Failed to compute conditional default for '%s': %s",
+                        dest,
+                        e,
+                    )
```
tjohnson31415, line 205: Seems like this should at least be a warning instead of debug (or an error)

tests/utils/test_platform_validation.py
```diff
@@ -258,3 +258,73 @@ def test_flex_device_set_for_sendnn_compile_only(self, monkeypatch):
...
+    @pytest.fixture(autouse=True)
+    def clear_conditional_defaults(self):
+        """Clear conditional defaults before each test to ensure isolation."""
+        from vllm_spyre.argparse_utils import ConditionalDefaultManager
+
+        ConditionalDefaultManager.clear()
+        yield
+        ConditionalDefaultManager.clear()
+
+    def create_test_parser(self):
```
tjohnson31415: NIT: Can this just be a fixture?

tjohnson31415 (commented):
Nice find on this mechanism!

vllm_spyre/platform.py
joerunde: Sure, were you thinking of only doing the `params.json` check then?

tjohnson31415: Right, but also resolving the model name to the directory to check the files for `params.json` instead of assuming that `mistral` in the model name means that `params.json` exists.

vllm_spyre/argparse_utils.py
joerunde: This is actually required as used, since `pre_register_and_update` is called with every subparser.

But, I think this is silly to save as a list so I'll prod the AI overlords to make this a dict instead with proper duplicate checking up-front instead of at apply-time

joerunde: done!

vllm_spyre/argparse_utils.py
joerunde: Yeah this wasn't that helpful. I removed it and just made sure that in debug mode there's a log for every logic branch during the actual override

joerunde:
@tjohnson31415 I think we should be good to go if you wanna take a quick look again

tjohnson31415:
Could your "AI overlords" also whip up tests for the model name resolution behavior?

Also, please update the PR description w.r.t. the 'Contains the string "mistral",' check

joerunde:
Yeah I'll see what the AI overlords say

joerunde:
@tjohnson31415 what say you now?

tjohnson31415 (approved):
lpgtm.
The `importlib.reload` invocations are less than ideal... I think you could monkeypatch instead (and use a mock for `get_model_path`), but will leave that up to you.

joerunde:
yeah, the reload is heavy handed but I didn't want to have to try to monkeypatch the places that the constants were used. 🤷 

We can always make it better later!

PR #933, opened 2026-04-17
:memo: Initiate MM vision encoder dedup ADR

gkumbhat:
## Checklist

### DIFF
--- a/docs/ADR/001-mm-vision-encoder-dedup.md
+++ b/docs/ADR/001-mm-vision-encoder-dedup.md
@@ -1,0 +1,92 @@
+# ADR-001: De-duplicate Vision Encoder Computation Across TP Workers
+
+## Context
+
+When running multimodal requests with tensor parallel (TP), each TP worker independently calls `get_maybe_mm_embeddings()` in `mistral3.py` during chunked prefill. This function runs on CPU and performs vision encoding followed by language model embedding projection.
+
+For TP=4, this means:
+- Vision encoding runs **4×** for every multimodal request (Duplication)
+- All workers **block** synchronously while encoding runs  - the causes ITL spike even though MM doesn't get processed during decode.
+- There is **no batching** of concurrent multimodal requests through the vision encoder
+
+### Key constraint
+
+`get_maybe_mm_embeddings()` lives in the model code, which is loaded inside worker processes. The engine process and scheduler have no access to the vision encoder. **Encoding must happen inside a worker process.**
+
+
+## Decision
+
+Encoding is assigned exclusively to **rank-0's worker process**, running in a **background thread pool**. Encoding is triggered as early as possible — the first time rank-0 sees `mm_features` for a request in `execute_model` — so that encoding overlaps with prefill chunk processing rather than blocking it.
+
+The result is written to a `SharedMemory` block. Each rank has a **dedicated result queue** fed by rank-0's background thread. Each worker's **polling thread** watches its queue, reads from shared memory, and sets a `threading.Event` to unblock the main thread when the embedding is ready.
+
+### Why rank-0, not a racing claim among workers
+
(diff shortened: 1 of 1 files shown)


docs/ADR/001-mm-vision-encoder-dedup.md
```diff
@@ -0,0 +1,92 @@
...
+#### Achieved
+- Vision encoding runs **once per request**, regardless of TP degree
+- Encoding runs in a **background thread** — rank-0's main thread is not blocked
+- Encoding runs off the main thread, so the main thread is not blocked. True overlap with the forward pass requires the scheduler to defer dispatch until the embedding is ready (future work).
+
+#### Pending
+
+- Still no explicit batching currently, which could improve performance further
+- Only partially helps with ITL Spike
+    - Since this solution reduces duplicate encoding, it will improve encode time, thus help with reducing prefill overbload and blocking
```
prashantgupta24, line 38:
```suggestion
    - Since this solution reduces duplicate encoding, it will improve encode time, thus help with reducing prefill overload and blocking
```

docs/ADR/001-mm-vision-encoder-dedup.md
```diff
@@ -0,0 +1,92 @@
...
+│   │
+│   ├── first sight of mm_features?
+│   │       YES → submit to thread pool
+│   │       NO  → already submitted
+│   │
+│   ├── do other chunk work...
+│   │
+│   └── before forward pass:
+│           future.result()
+│           (instant if done, waits if not)
```
prashantgupta24, line 58: is the `do other chunk work` significant enough to warrant a separate thread?

joerunde:
@gkumbhat:

1. do you have any initial perf numbers from a POC to show what kind of savings this could have?
2. Does vllm handle image encoding in a similar way today, or would this be blazing our own path for broadcasting embeddings across workers?

gkumbhat:
@joerunde 

> do you have any initial perf numbers from a POC to show what kind of savings this could have?

am working on that

> Does vllm handle image encoding in a similar way today, or would this be blazing our own path for broadcasting embeddings across workers?

vllm upstream doesn't do image encoding quite this way.. They have multiple paths:
1. Handling this via token budgeting with same TP for decoder model - in this case, vision model also gets sharded along with decoder across the TP workers
2. Ability to do DP for MM and TP for decoder models
3. Disaggregated encoder

This particular approach is trying to optimize current limitation of running vision encoder on CPU, but I think would be beneficial for new stack as well.. (Pending validations).. This approach draws some inference from disaggregated encoder approach, but not going quite as far as running different server, for model load, compile, and data sharing simplicity

joerunde:
> but I think would be beneficial for new stack as well..

But if we have full torch inductor integration, why wouldn't we want to use vllm's path (1) of sharding the vision model along with the language model? Intuitively it seems to me like adding an extra step for the rank 0 worker to run the vision model and then broadcast results to other workers would be slower- you'd tie up the device on rank 0 which would necessarily pause the other ranks from doing any useful work too. And for any reasonably sized visual encoders, running them on the cpus would be too slow, right?

gkumbhat:
> But if we have full torch inductor integration, why wouldn't we want to use vllm's path (1) of sharding the vision model along with the language model? Intuitively it seems to me like adding an extra step for the rank 0 worker to run the vision model and then broadcast results to other workers would be slower- you'd tie up the device on rank 0 which would necessarily pause the other ranks from doing any useful work too. And for any reasonably sized visual encoders, running them on the cpus would be too slow, right?

Definitely. On new stack we would want to run vision encoder on the device itself (instead of CPU). Running it on accelerator would give lot of performance benefit itself. What I meant was, running it on device + some iteration of this approach could be useful.. Not running it on rank-0, but some type of shared queue based mechanism could be a middle-ground option until we do (3) completely.

gkumbhat:
Closing as we have a different and simpler implementation available in #992

PR #934, opened 2026-04-17
fix: Improve debuggability and fix ptr address replacement for test_compare_graphs.py

tjohnson31415:
## Description

I noticed some internal testing failing the graph comparison test for the sendnn stack when running on the side-by-side dev image. This PR enhances the graph comparison test and fixes the failure:
 - Fixed pointer normalization regex to handle variable-length hex addresses
 - Added save_normalized_graphs() to dump normalized graphs for easier debugging
 - Store graph results using pytest's `tmp_path_factory` to persist them for debugging (default is to preserve tmp directories for 3 pytest runs)
 - Enhanced failure diagnostics with graph counts, missing keys, and diff commands

## Test Plan

Tested in my dev pod with
```
pytest -svx tests/aftu/test_compare_graphs.py 
```
Will also run CI tests.

## Checklist


tests/aftu/test_compare_graphs.py
```diff
@@ -93,25 +106,58 @@ def test_compare_graphs_chunked_prefill(
...
     finally:
         # Restore in case of exception
         os.chdir(original_cwd)
 
-    assert compare_graphs(vllm_graphs, aftu_graphs)
+    graphs_match = compare_graphs(vllm_graphs, aftu_graphs)
+
+    if not graphs_match and keep_graphs:
+        print("\n[DEBUG] Graph comparison failed. Directories preserved:")
+        print(f"[DEBUG]   AFTU: {aftu_tmpdir}")
```
joerunde: the `[DEBUG]` bits read like real debug logs- should we use `logger.debug` instead? Or `logger.error` because these are test error logs that we should probably always print?

tests/aftu/graph_compare_utils.py
```diff
@@ -58,6 +58,27 @@ def collect_graph_files(input_dir: str) -> dict[str, tuple[str, str]]:
...
+    Args:
+        graph_map: Dictionary mapping graph keys to (filepath, normalized_content) tuples
+        output_dir: Directory to save normalized graphs
+    """
+    normalized_dir = path.join(output_dir, "normalized_graphs")
+    os.makedirs(normalized_dir, exist_ok=True)
+
+    for key, (original_path, normalized_content) in graph_map.items():
+        # Create a clean filename from the key
+        filename = f"graph{key}.ops"
```
joerunde, line 93: do we need to worry about `key` containing path separators?

joerunde (approved):
lpgtm!

tests/aftu/graph_compare_utils.py
tjohnson31415: Right now, it is just a number like 0000, 0001 parsed from the source files. So I don't think we need to worry about that.

tests/aftu/test_compare_graphs.py
tjohnson31415: Yeah, good point. This mostly was "real debug logs" that I just threw into a commit. I've pushed a refactor for a better integration: debug info and graph references dumped on test failure to the pytest fail message.

tjohnson31415:
bot:test
TEST_FILE=tests/aftu

tests/aftu/test_compare_graphs.py
```diff
@@ -93,25 +113,58 @@ def test_compare_graphs_chunked_prefill(
...
+        if missing_in_aftu:
+            error_msg.append(f"Missing in AFTU: {sorted(missing_in_aftu)}")
+
+        if differences:
+            error_msg.append("\nDifferences:")
+            error_msg.extend(differences)
+
+        error_msg.append(f"\nCompare with: diff -ur {aftu_normalized_dir} {vllm_normalized_dir}")
+
+        pytest.fail("\n".join(error_msg))
```
joerunde, line 170: yooo I never knew this was a thing. Slick!

joerunde (approved):

PR #941, opened 2026-04-21
Introduce async scheduler implementation with mixin pattern

GOavi101:
## Async Scheduling for Spyre Generative Models

Wires Spyre’s chunked-prefill scheduler to vLLM’s upstream AsyncScheduler so async scheduling (run-ahead / batch queue as implemented by the pinned vLLM v1 engine) can be used for generative models. Spyre-specific reconcile and worker changes keep scheduling correct under optimistic num_computed_tokens and related state.

---

### Background

Upstream async scheduling lets the engine overlap scheduling with execution: the scheduler can advance before the previous step’s outputs are fully committed (run-ahead). In v1 this is tied to the engine’s async / batch-queue execution path (see comments in scheduler.py referencing step_with_batch_queue — exact wiring lives in vLLM, not reimplemented here).

Spyre’s ChunkedPrefillSpyreScheduler keeps extra mutable state (ongoing_prefills, TKV / volumetric admission, previous_step_was_prefill, …) that vanilla Scheduler does not own. Base _update_after_schedule optimistically bumps num_computed_tokens for scheduled prefill chunks; the Spyre runner then commits actual progress (left-padding, prefix-cache hits can differ). Without reconciliation, a speculative schedule(N+1) could read wrong committed vs in-flight positions and make bad admission decisions.

---

### What changed

**`sendnn_inference/v1/core/async_scheduler.py`** _(new file)_

`AsyncChunkedPrefillSpyreScheduler` is a thin subclass combining `ChunkedPrefillSpyreScheduler` and the upstream `AsyncScheduler` via multiple inheritance. Python's MRO ensures `super().schedule()` hits `AsyncScheduler` before the base `Scheduler`. No pooling async variant is created — async scheduling has no benefit for pooling models.

**`sendnn_inference/v1/core/scheduler.py`**

- `_inflight_prefill_tokens: dict[str, int]` — tracks the optimistic `num_computed_tokens` delta per request between `_update_after_schedule` and the runner's actual commit. In sync mode this is a no-op (added and cleared in the same step). Under run-ahead it lets `update_from_output` reconcile to the runner's real report, and lets the speculative `schedule()` call distinguish committed from in-flight tokens when evaluating `ongoing_prefills` and TKV admission.
- `update_from_output` — reconciles `num_computed_tokens` using the inflight delta.
- `finish_requests` — pops specific request IDs (or clears on finish-all) so aborted or mid-prefill-finished requests don't leave stale entries.
- Defensive `assert len(_inflight_prefill_tokens) <= max_num_seqs`.

**`sendnn_inference/v1/worker/spyre_worker.py`**

The async engine calls `execute_model` and `sample_tokens` as separate steps. `execute_model` defers its result to `_pending_sample`; `sample_tokens` drains it. Warmup and cleanup paths that must not leave `_pending_sample` populated use `_execute_and_sample` instead of calling `execute_model` directly. Grammar masking falls back gracefully when `_spyre_grammar_output` is absent.

**`sendnn_inference/platform.py`**

`get_spyre_scheduler_cls` centralises class selection, validates non-string `scheduler_cls` values as `Scheduler` subclasses, and logs the selected class name at startup so pooling cases cannot misleadingly appear as async-enabled.

**`tests/v1/core/test_scheduler.py`**

New and extended tests: sync/async schedule-update cycles, `_inflight_prefill_tokens` correctness under normal flow and mid-prefill abort, MRO ordering (`mro.index(AsyncScheduler) < mro.index(Scheduler)`), `finish_requests` cleanup.

  

## Checklist


vllm_spyre/v1/core/scheduler.py
```diff
@@ -1,24 +1,11 @@
...
-from vllm_spyre.platform import SpyrePlatform
-from vllm_spyre.v1.worker.spyre_model_runner import SpyreModelRunnerOutput
 
-if TYPE_CHECKING:
-    from vllm.v1.core.sched.output import SchedulerOutput
-else:
-    SchedulerOutput = None
-
-logger = init_logger(__name__)
+from vllm_spyre.v1.core.scheduler_impl import (
```
joerunde: @GOavi101 it looks like most of this file has been deleted and moved to scheduler_impl. Can you put the implementation back in this file so that reviewers can see what's changed?

GOavi101: sure joe

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -756,9 +761,17 @@ def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
...
+        #   2. sample_tokens() - retrieves the result
+        # Spyre samples inside execute_model(), so we cache the output and
+        # return it here. EMPTY_MODEL_RUNNER_OUTPUT is returned if execute_model
+        # produced no output (e.g. a prefill-only step).
         from vllm.v1.outputs import EMPTY_MODEL_RUNNER_OUTPUT
 
-        return EMPTY_MODEL_RUNNER_OUTPUT
+        cached = self._last_execute_model_output
+        self._last_execute_model_output = None
+        return cached if cached is not None else EMPTY_MODEL_RUNNER_OUTPUT
```
joerunde: Ideally we would actually run the sampling here - see related comment on the structured output PR: https://github.com/torch-spyre/sendnn-inference/pull/903#discussion_r3120759211

I'm fine with leaving this as-is and then fixing it to work with both async scheduling and structured outputs in a followup. Issue opened here: https://github.com/torch-spyre/sendnn-inference/issues/947

tests/v1/core/test_async_scheduler.py
```diff
@@ -0,0 +1,477 @@
+"""Unit tests for async scheduler mixin classes.
+
+Tests the PoolingSpyreMixin and ChunkedPrefillSpyreMixin behaviour when used
+with AsyncScheduler (async mode) vs Scheduler (sync mode).
+
+Key behaviours under test:
+  - _is_async_scheduler() correctly identifies async vs sync instances
+  - PoolingSpyreMixin.schedule() applies warmup-shape constraints in both modes
+  - ChunkedPrefillSpyreMixin.schedule() bypasses Spyre constraints in async mode
```
joerunde: This statement seems incorrect- we definitely can't just bypass spyre constraints because there are hard limits to what we can run on the cards. What's really going on?

sendnn_inference/platform.py
```diff
@@ -249,7 +319,19 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
             vllm_config.scheduler_config.enable_chunked_prefill = False
 
-            scheduler_config.scheduler_cls = "vllm_spyre.v1.core.scheduler.PoolingSpyreScheduler"
+            # Select scheduler using get_spyre_scheduler_cls(), following upstream's pattern
+            scheduler_cls = cls.get_spyre_scheduler_cls(
+                scheduler_config=scheduler_config,
+                is_pooling=True,
+            )
+            # Set as string path for vLLM's resolution (matches upstream behavior)
+            # Only convert to string if it's not already a string
```
joerunde: a class should be fine to pass here though, what goes wrong?

sendnn_inference/platform.py
```diff
@@ -219,9 +271,27 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
 
         if is_decoder:
-            scheduler_config.scheduler_cls = (
-                "vllm_spyre.v1.core.scheduler.ChunkedPrefillSpyreScheduler"
+            # With MultiprocExecutor (TP>1), worker futures are truly async (ZMQ),
+            # so step_with_batch_queue runs schedule() twice before update_from_output().
+            # The mixin's pre-filter pattern is not safe under that run-ahead scenario.
+            # For TP=1 (UniProcExecutor), futures are immediately done so it's safe.
+            if parallel_config.world_size > 1:
+                scheduler_config.async_scheduling = False
```
joerunde: Interesting- if we wanted to support this feature then it would likely need to work with TP=4 which is how we run most models. I thought this was only incompatible with pipeline parallel upstream - does it also not work with tensor parallel?

vllm_spyre/v1/core/scheduler.py
joerunde: Thanks, I've looked through the tests but I'll wait to review the code changes until after this diff is in nicer shape- I don't really want to try to recreate the diff myself 😉

joerunde:
Thanks @GOavi101!

A few notes:

1. If this can't be done with tensor parallel, then maybe it's not worth pursuing. Is that a hard blocker?
2. We need to have an end-to-end test that shows this working, ie using an `LLM` with async scheduling enabled. It would also be good to include an illustrative test at the engine level (see https://github.com/torch-spyre/sendnn-inference/blob/main/tests/e2e/test_spyre_pc_scheduler_steps.py) that shows the effects of async scheduling. From my quick skim it sounds like the engine is speculatively scheduling batches one step ahead, so we should see a "dead token" in some cases where the engine schedules a decode past the end of a sequence.
3. It would be really great to see a profile of this in action, or at least some minimal `vllm bench` results showing what kind of performance improvement we can expect.

sendnn_inference/platform.py
GOavi101: @joerunde 

The fix is SpyreMultiprocExecutor — a thin MultiprocExecutor subclass that overrides max_concurrent_batches to return 1 instead of 2. This forces the engine to use the simpler step() path (strictly schedule → execute → update) rather than step_with_batch_queue, which was the only thing that broke TP>1.
 Spyre's  forward pass is synchronous, so there's no compute/schedule overlap to lose. The AsyncScheduler base class and its _update_after_schedule TTFT benefit are still fully active — we just removed the run-ahead that its state tracking couldn't handle.
So TP=1, TP=2, and TP=4 should all work with async scheduling now. Not a blocker.

 what do you think?

sendnn_inference/v1/worker/spyre_worker.py
GOavi101: Agreed, thanks for opening the issue. I've added a TODO(#947) comment pointing to it so it's tracked directly in the code.

tests/v1/core/test_async_scheduler.py
GOavi101: Correct, nothing is bypassed — that docstring was wrong. same constraints apply in both modes. only async-specific code is a stale ongoing_prefills cleanup needed because _update_after_schedule speculatively advances num_computed_tokens before update_from_output() confirms it. fixed the docstring.

sendnn_inference/platform.py
GOavi101: Nothing goes wrong — you're right. SchedulerConfig.scheduler_cls is typed str | type | None and get_scheduler_cls() handles a class directly. string conversion was unnecessary. removed it in the latest push.

vllm_spyre/v1/core/scheduler.py
GOavi101: Done!!!

GOavi101:
After vLLM 0.14, the async scheduler is enabled by default. All the tests below are running using the async scheduler. To run with the synchronous scheduler:
we have to add the `--no-async-scheduling` flag.

sendnn_inference/platform.py
joerunde: That doesn't quite line up with my understanding- IIUC the `step_with_batch_queue` method is what works with the speculative scheduling: The engine runs the scheduler again while the model is running, assuming that the requests in the batch will continue.

> Spyre's forward pass is synchronous, so there's no compute/schedule overlap to lose

I don't quite understand this either- the multiproc executor is definitely async, it broadcasts an RPC to the workers to run the model and the engine gets back a future that it waits on. `step_with_batch_queue` queues up that future so that it can speculatively schedule the next pass.

This TP=1 profile shows the scheduler running in between the model forward passes, the goal with async scheduling is to get the scheduler running for the next step during the model forward pass instead:

image

> The AsyncScheduler base class and its _update_after_schedule TTFT benefit are still fully active — we just removed the run-ahead that its state tracking couldn't handle.
So TP=1, TP=2, and TP=4 should all work with async scheduling now. Not a blocker.

Based on the above, my understanding is that the run-ahead state is the whole point and we won't gain any performance benefit from this unless we support it, so this is a blocker. Is there something else I'm missing?

GOavi101: You're right, thanks for the correction. I'll fix this  — snapshot the mixin's mutable state (ongoing_prefills, tkv, previous_step_was_prefill) before delegating to super().schedule() so the run-ahead second schedule() call sees consistent state, and remove SpyreMultiprocExecutor. That way TP≥2 gets the full async scheduling benefit.

maxdebayser: Yeah, the `step_with_batch_queue` is actually what makes async scheduling work at all. In [the PR](https://github.com/vllm-project/vllm/pull/19970) where Woosuk added this, he reused the existing batch queue that was originally intended for Pipeline Parallel (PP). With PP you have to wait N steps for each of the N PP stages. But with async scheduling the intention is for the scheduler to be ahead by 1 step, so the `num_output_placeholders` were introduced to prevent the scheduler from waiting for the current step.

sendnn_inference/platform.py
```diff
@@ -95,13 +95,70 @@ def import_kernels(cls) -> None:
...
+        # Import from appropriate module based on async_scheduling config
+        # These modules have classes created at module level, so they're importable
+        if scheduler_config.async_scheduling:
+            # Use async scheduler variants
+            if is_pooling:
+                from sendnn_inference.v1.core.async_scheduler import (
+                    AsyncPoolingSpyreScheduler,
+                )
+
+                return AsyncPoolingSpyreScheduler
```
maxdebayser: I would prefer removing the AsyncPoolingSpyreScheduler for now. It's really an optimization for generative use cases and it's disabled upstream because it negatively affects the pooling performance: https://github.com/vllm-project/vllm/blob/13338644082bdab10c4ff017de1a97c03ba62524/vllm/config/vllm.py#L813

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -760,9 +763,21 @@ def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
...
+        #
+        # TODO(#947): Ideally sampling should happen here to support structured
+        # outputs with async scheduling. Tracked in:
+        # https://github.com/torch-spyre/sendnn-inference/issues/947
         from vllm.v1.outputs import EMPTY_MODEL_RUNNER_OUTPUT
 
-        return EMPTY_MODEL_RUNNER_OUTPUT
+        cached = self._last_execute_model_output
+        self._last_execute_model_output = None
+        return cached if cached is not None else EMPTY_MODEL_RUNNER_OUTPUT
```
maxdebayser: Isn't it better to just always return None in `SpyreModelRunner.execute_model()`? The engine core knows that it has to call `sample_tokens` in that case. In this case, you just need to split out the sampling code in `execute_model` into a `sample_tokens` method.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1398,6 +1397,23 @@ def prepare_model_input(self, scheduler_output: SchedulerOutput) -> SamplingForw
 
             return model_inputs
         else:
+            # Reconcile input_batch with the current decode set.
+            # With async run-ahead scheduling a request can appear in both
+            # finished_req_ids and scheduled_cached_reqs simultaneously (the
```
maxdebayser: I don't think this should happen. The scheduler is the component that decides when requests are done so it should never schedule a finished request. With async scheduling, a request can be technically finished, but the scheduler only discovers that in the next step, so it's never "knowingly" scheduling a finished request.

maxdebayser (changes requested):
Thanks for taking on this issue, @GOavi101. I think perhaps it would better to take a step back to re-think this PR a bit.

There are a few unecessary changes in this PR. I would suggest the following:
1) Remove the `AsyncPoolingSpyreScheduler` class and the code that selects it in `platform.py` as async scheduling is not advantageous for pooling
2) Undo the mixin class structure. Currently the only purpose of this is for `_is_async_scheduler()` to run `isinstance(self, AsyncScheduler)` but the same can be achieved by returning `vllm_config.scheduler_config.async_scheduling`

In [PR 19970](https://github.com/vllm-project/vllm/pull/19970) Woosuk added async scheduling based in this idea from the NanoFlow paper:

```
Asynchronous scheduling: Batch formation, including estimating memory 
usage, scheduling new requests, retiring finished requests, and adjusting the 
page table for PagedAttention [17], consumes a non-negligible amount of time 
on the CPU side [42]. In most serving frameworks [17,58], only after the GPU 
executes one iteration, the scheduler on the CPU is able to detect EOS tokens, 
remove the finished request from the batch, and refill the batch with new 
requests. However, GPU is under-utilized during this time. To avoid this waste, 
NanoFlow asynchronously schedules batch formation in parallel to the GPU 
executions. At any iteration i, NanoFlow forms the batch for the next iteration 
before the end of the current iteration. This means that NanoFlow cannot 
detect the EOS tokens generated at iteration i. After launching iteration i+1, 
NanoFlow forms the batch for cycle i+2, detects the EOS token from iteration i, 
and removes finished requests.  Fortunately, since the average decode length 
surpasses 100 for typical workloads (See Table 4), the overhead of one extra 
decode token is negligible (< 1%), given the benefit of hiding the batch 
formation overhead.
```

In this first PR, no model runner changes were needed. Later additions were made to overlap the model runner execution with other steps such as sampling and the `WorkerBase` `sample_tokens` method was added to separate sampling from the execution of the model. 

I think that the main problem to solve is that in Spyre we can't run prefills in the same batch as the decode requests. If it weren't for this fact, probably the only required code change would be implementing the `sample_tokens()` method in the SpyreWorker. 
Since we can currently interleave prefill chunk batches and decode batches, in principle there is no problem in doing so asynchronously.
But once prefill is done, we either must wait and add the request in the current decode batch or in the one after that.

sendnn_inference/v1/core/scheduler.py
```diff
@@ -141,58 +119,66 @@ def _get_matching_warmup_shapes(
...
+    cleared within a single step (a no-op). ``update_from_output()`` also
+    filters scheduler outputs to only include requests that were actually
+    executed (relevant under run-ahead).
+
+    Note: we deliberately do NOT mirror the prefill advance into upstream's
+    ``Request.num_output_placeholders``; that field feeds upstream's
+    ``num_new_tokens = num_tokens_with_spec + num_output_placeholders -
+    num_computed_tokens`` formula and is reserved for in-flight *decode*
+    tokens. Bumping it for prefill chunks would cause the formula to
+    re-schedule the full prompt length on every run-ahead step.
```
maxdebayser: I'm not sure I understand this. In the upstream scheduler prefill chunks are skipped, so the num_output_placeholders is not bumped: https://github.com/vllm-project/vllm/blob/0c2e9d48927f8122cba25f9a649a72f131faffb3/vllm/v1/core/sched/async_scheduler.py#L23

sendnn_inference/v1/core/scheduler.py
```diff
@@ -141,58 +119,66 @@ def _get_matching_warmup_shapes(
...
         # Theoretically, only one request can be prefilled at a time, but we
         # keep a list to be able to batch prefills in the future.
         self.ongoing_prefills: list[Request] = []
 
+        # Per-request counter of prefill-chunk tokens whose optimistic
+        # advance has been applied to ``Request.num_computed_tokens`` but
+        # not yet committed by ``update_from_output``. The committed prefill
+        # position for ``req`` is
+        # ``req.num_computed_tokens - self._inflight_prefill_tokens.get(rid, 0)``.
+        self._inflight_prefill_tokens: dict[str, int] = {}
```
maxdebayser, line 210: Why is this necessary? Upstream the scheduler skips the chunks:
```
            if request.is_prefill_chunk:
                continue
```

sendnn_inference/v1/core/scheduler.py
```diff
@@ -207,35 +193,159 @@ def __init__(self, *args, **kwargs) -> None:
...
+        # weren't executed yet.  Filter to only pass executed requests to the
+        # base scheduler.
+        filtered_num_scheduled_tokens = {
+            req_id: num_tokens
+            for req_id, num_tokens in scheduler_output.num_scheduled_tokens.items()
+            if req_id in executed_req_ids
+        }
+        if len(filtered_num_scheduled_tokens) != len(scheduler_output.num_scheduled_tokens):
+            import dataclasses
+
```
maxdebayser: I haven't found examples of the `scheduler_output` being modified in the upstream code. I would be careful with this type of thing because the scheduler is a complex piece of code with many hidden assumptions.

sendnn_inference/v1/core/scheduler.py
GOavi101: the original comment was misleading. Upstream AsyncScheduler._update_after_schedule already guards with if request.is_prefill_chunk: continue, so num_output_placeholders is never bumped for prefill chunks to begin with. The note was trying to explain why we don't need to undo anything in update_from_output for decode placeholders, but it read as if we were doing something special to prevent bumping. 

I've reworded the docstring to say: "Upstream already skips prefill chunks via the is_prefill_chunk guard, so num_output_placeholders is only ever used for in-flight decode tokens. _inflight_prefill_tokens is a separate, Spyre-specific mechanism to track the optimistic num_computed_tokens advance so update_from_output can reconcile against the runner's actual committed tokens."

sendnn_inference/v1/core/scheduler.py
GOavi101: The is_prefill_chunk skip in upstream's _update_after_schedule only avoids bumping num_output_placeholders (which is decode-only). The base Scheduler._update_after_schedule still optimistically advances request.num_computed_tokens by num_scheduled_tokens for prefill chunks — that's what enables schedule(N+1) to pick the next chunk while execute(N) is in flight.
We need _inflight_prefill_tokens because on Spyre the committed prefill position after the runner returns is not necessarily equal to the optimistically-scheduled amount: the model runner adjusts for left-padding and prefix-cache hits and reports back what was actually consumed. Without tracking the per-chunk optimistic delta, update_from_output cannot reconcile num_computed_tokens to the runner's actual report, and (under run-ahead) schedule() cannot tell which portion of num_computed_tokens is committed vs. in-flight when deciding whether a request is still in ongoing_prefills.
In sync mode this is a no-op (added and cleared in the same step). It only matters when the async run-ahead inserts a schedule() call between the optimistic advance and the runner's commit.

PR #951, opened 2026-04-24
:bug: Fix mistral tokenizer loads

joerunde:
## Description

This fixes a bug where mistral models fail to load a tokenizer, even if `--tokenizer-mode mistral` is supplied. This bug was introduced in vllm 0.19.1, where now a mistral tokenizer is loaded iff a `consolidated.safetensors` file exists in the model directory, regardless of CLI inputs.

We need to follow up to see if this is fixed in vllm main yet, and fix it upstream if it isn't.

## Test Plan

ministral should load correctly again without requiring a cached consolidated.safetensors file or HF hub access.

## Checklist

### DIFF
--- a/tests/utils/test_upstream_compatibility.py
+++ b/tests/utils/test_upstream_compatibility.py
@@ -17,7 +17,7 @@ VLLM_VERSION = os.getenv("TEST_VLLM_VERSION", "default")
 def test_tokenizer_registry_get_config_patch():
     """
     When this test starts failing because get_config exists in the lowest supported vllm version,
-    the compatibility patch in platform.py can be removed.
+    the check to conditionally _not_ apply _patch_tokenizer_registry_get_config can be removed.
     """
     import vllm.tokenizers.registry as tokenizer_registry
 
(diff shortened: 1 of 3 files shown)


tests/utils/test_upstream_compatibility.py
```diff
@@ -13,5 +13,17 @@
 
 VLLM_VERSION = os.getenv("TEST_VLLM_VERSION", "default")
 
-# We reset the minimum version, so no tests currently
-# Check git history of this file for examples of the tests
+
+def test_tokenizer_registry_get_config_patch():
+    """
+    When this test starts failing because get_config exists in the lowest supported vllm version,
+    the compatibility patch in platform.py can be removed.
```
tjohnson31415: NIT: This comment is incorrect, we'd still need the patch when the lowest version has get_config.

The patch isn't really a backwards compat patch, but more of a fowards compat hack... 😅

joerunde: ah, I had meant the bit that _doesn't_ apply the hack lol

(edited)

tjohnson31415 (approved):
🙈 
LGTM

PR #952, opened 2026-04-28
chore: add vllm v0.20.0 support and model smoke tests

rafvasq:
## Description

- Bumps to vLLM v0.20.0 (minimum), remove v0.19.0-related compat guard in `_patch_tokenizer_registry_get_config` and related test 
- Bumps to torch `2.11.0`, and removes the build constraint around it (v0.19.1 uses 2.10.0, v0.20.0 uses 2.11.0)
- Adds `manual_seed_all` stub (https://github.com/vllm-project/vllm/pull/38468) 
- `compile_or_warm_up_model` returns `CompilationTimes` (https://github.com/vllm-project/vllm/pull/39240/), plus backwards compat shim
- Adds new test workflow to download and boot a subset of supported models on relevant PRs to catch model-loading regressions (along with a marker for these tests (`@pytest.mark.smoke`)

## Test Plan

- Get GHA and spyre-ci tests passing

## Checklist


rafvasq:
bot:test
MARKERS="spyre and prefix_caching and not quantized"

.github/workflows/model_smoke.yml
```diff
@@ -0,0 +1,135 @@
...
+
+      - name: "Install vLLM with Spyre plugin"
+        run: |
+          uv venv .venv --system-site-packages --clear
+          source .venv/bin/activate
+
+          uv sync --frozen
+          uv pip install -v .
+
+      - name: "Standardize HF model names for caching"
```
joerunde: 🤔  we're at capacity in our GHA cache currently with the tiny models that we use for our existing unit tests. If we want to smoke test larger models, we won't be able to cache them

.github/workflows/model_smoke.yml
```diff
@@ -0,0 +1,135 @@
...
+            smoke_test: "test_embedding_model_load_smoke"
+            hf_model: "sentence-transformers/all-roberta-large-v1"
+            hf_model_rev: "cf74d8acd4f198de950bf004b262e6accfed5d2c"
+          - name: "scoring"
+            smoke_test: "test_scoring_model_load_smoke"
+            hf_model: "cross-encoder/stsb-roberta-large"
+            hf_model_rev: "2b12c2c0088918e76151fd5937b7bba986ef1f98"
+          - name: "mistral"
+            smoke_test: "test_decoder_model_load_smoke"
+            hf_model: "mistralai/Mistral-Small-3.2-24B-Instruct-2506"
```
joerunde: We could try `mistralai/Ministral-3-3B-Instruct-2512-BF16` for a much smaller model to smoke test

joerunde:
bot:test

pyproject.toml
```diff
@@ -53,7 +53,7 @@ git_describe_command = "git describe --dirty --tags --long --match 'v*'"
 # So we set to 'never' to prevent installing it or eventually any other
 # by accident
 override-dependencies = [
-    "torch==2.10.0",
+    "torch==2.11.0",
```
joerunde, line 56: cc @gkumbhat 

We'll need to be extra careful about the bump to 2.11 here. Let's start with the spyre unit tests then let's get an image built on the 1.2.1 runtime stack with these changes and run through the full set of integration tests internally before merging

maxdebayser: BTW, why do we need to set a specific torch version? Can't we set a range?

maxdebayser (approved):

pyproject.toml
joerunde: We're so open that we don't even have a range in our own dependencies, we just specify `torch`!

For vllm 0.20.0 torch is pinned to 2.11 so any builds with 0.20 will only resolve torch to 2.11. This override here is a uv-only dependency-resolution-time override to ensure we're building and testing with 2.11, which product teams from Red Hat would be using in their builds too.

pyproject.toml
```diff
@@ -88,12 +88,12 @@ override-dependencies = [
 ]
 # This adds constraints to all dependent build environments, which will ensure everything is built
 # with the same version of torch. This CANNOT conflict with a package's existing build dependencies
-build-constraint-dependencies = ["torch==2.10.0"]
+build-constraint-dependencies = []
```
joerunde: Were there build-time conflicts when updating this?

This override is to ensure that every dependency is built with the same version of torch, to avoid situations where e.g. vllm is compiled with torch 2.10 at build time but then we install torch 2.11 at runtime. 

This probably isn't that big of a concern for this stack since we don't expect to be using any precompiled artifacts from anything except our own runtime stack, which is always pre-installed from rpms, but if there were conflicts here it would be good to know which builds complain about the upgrade in case there is something that we need to go fix

rafvasq: Yeah, left at torch 2.10 it caused a conflict with vllm 0.20.0. It sounds like I should update it to torch 2.11 then, I wasn't sure how it affected consumers/v0.19.1 support so I left it empty.

(edit) Ah, this also causes the build failure for v0.19.1, so I think it has to stay empty.

joerunde: ohhh, because we build for older versions in our test matrix :(((((


well :rip: but that's not the end of the world

PR #954, opened 2026-04-29
Tkv limit bug

Daniel-Schenker:
## Description

This PR addresses a bug in the scheduler code where the scheduler would allow requests that would eventually exceed the max batch tkv limit during an ongoing batch. The PR includes a new test case to reproduce the error and the fix for said error.

## Test Plan

Run the test file with: `pytest tests/v1/worker/test_scheduler_tkv_limits.py -sv`
Second test case will pass with changes in `sendnn_inference/v1/core/scheduler.py` included in this PR, and will fail without.

## Checklist

### DIFF
--- a/sendnn_inference/v1/core/scheduler.py
+++ b/sendnn_inference/v1/core/scheduler.py
@@ -20,12 +20,15 @@ else:
 
 logger = init_logger(__name__)
 
+# Ensure that block_size is 64
+# This ensures the rounding function is correct
+assert SpyrePlatform.get_block_size() == 64
+
 
 def round_up_to_block_size(n: int) -> int:
     # Helper function to round up to the nearest block size
     # Uses bitwise alignment for better performance
-    block_size = SpyrePlatform.get_block_size()
-    return (n + block_size - 1) & ~(block_size - 1)
+    return (n + 63) & ~63
 
 
 class SpyreScheduler(Scheduler):
@@ -538,7 +541,7 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
         """
 
         # Compute the effective token length of the new request
-        # Rounded up to the nearest block size
+        # Rounded up to the nearest block size to account for potential padding
         new_req_max_tkv = round_up_to_block_size(new_req_tkv + request.max_tokens - 1)
 
         # Compute token lengths for all running requests (decode batch)
@@ -547,6 +550,7 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
         dec_req_tkv = max(self.tkv, request.num_prompt_tokens)
         for req in running:
             n_generated_output_tokens = req.num_computed_tokens - req.num_prompt_tokens
+            # Rounded up to the nearest block size to account for potential padding
             dec_req_max_tkv = round_up_to_block_size(
                 dec_req_tkv + (req.max_tokens - n_generated_output_tokens) - 1
             )
--- a/sendnn_inference/v1/core/scheduler.py
+++ b/sendnn_inference/v1/core/scheduler.py
@@ -515,7 +515,7 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
         return num_prefills < max_concurrent_prefills
 
     def check_batch_tkv_limit_cp(
-        self, request: Request, new_req_tkv: int, n_blocks: int, running
+        self, request: Request, new_req_tkv: int, running
     ) -> bool:
         """
         Check whether adding a new sequence to the decode batch would violate
--- a/sendnn_inference/v1/core/scheduler.py
+++ b/sendnn_inference/v1/core/scheduler.py
@@ -20,6 +20,9 @@ else:
 
 logger = init_logger(__name__)
 
+def round_up_to_block_size(n: int) -> int:
+    block_size = SpyrePlatform.get_block_size()
+    return (n + block_size - 1) & ~(block_size - 1)
 
 class SpyreScheduler(Scheduler):
     """Base class inheriting from the V1 scheduler to support static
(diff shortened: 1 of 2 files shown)


Daniel-Schenker:
Initial proposed changes are here. More than happy to address any comments and add/change code and comments if things aren't clear.

sendnn_inference/v1/core/scheduler.py
```diff
@@ -21,6 +21,13 @@
 logger = init_logger(__name__)
 
 
+def round_up_to_block_size(n: int) -> int:
+    # Helper function to round up to the nearest block size
+    # Uses bitwise alignment for better performance
+    block_size = SpyrePlatform.get_block_size()
+    return (n + block_size - 1) & ~(block_size - 1)
```
joerunde: I'm not sure we should make this generic, because the bitwise ops only work if the block size is a power of 2. What about hardcoding 64 here, and sticking an assert in the calling code that block_size == 64?

I have a reasonably low care amount because we never expect the block size to change, only a slight preference to not leave a foot-gun lying around.

tests/v1/worker/test_scheduler_tkv_limits.py
```diff
@@ -80,3 +80,135 @@ def test_scheduler_tkv_limits(monkeypatch: pytest.MonkeyPatch):
...
+    )
+
+    # Configure the TKV limit
+    scheduler = model_runner.scheduler
+    scheduler.max_batch_tkv_limit = 131072
+    SpyrePlatform._max_batch_tkv_limit = 131072
+    monkeypatch.setenv("VLLM_DT_MAX_BATCH_TKV_LIMIT", "131072")
+
+    # Define prompt lengths for first set of requests
+    prompt_lengths_1 = [
```
joerunde: [1024] * 16?

or [1018] + [1024] * 15 if the first 1018 is relevant

sendnn_inference/v1/core/scheduler.py
```diff
@@ -531,17 +538,18 @@ def check_batch_tkv_limit_cp(
         """
 
         # Compute the effective token length of the new request
-        new_req_max_tkv = new_req_tkv + request.max_tokens - 1
+        # Rounded up to the nearest block size
```
joerunde: Should we mention here (and below) that we're rounding up to the end of the block to account for padding?

tests/v1/worker/test_scheduler_tkv_limits.py
```diff
@@ -80,3 +80,135 @@ def test_scheduler_tkv_limits(monkeypatch: pytest.MonkeyPatch):
         scheduler.update_from_output(sched_output, output)
         if len(scheduler.running) == 0:
             break
+
+
+@pytest.mark.cpu
+@pytest.mark.chunked_prefill
+def test_scheduler_tkv_limits_ongoing_batch(monkeypatch: pytest.MonkeyPatch):
```
joerunde, line 87: A high-level overview of what's going on here would be great.

IIUC this situation is:
- We first schedule a 16x8k batch that would fully fill the 128k TKV limit
- We then inject a batch of smaller requests partway through processing, which should be able to schedule only because they are guaranteed to finish processing just before the TKV is long enough to overrun the limit with the larger batch size
- This flexes the logic for injecting shorter requests into a running batch, which is not tested by the other test in this file

sendnn_inference/v1/core/scheduler.py
Daniel-Schenker: Addressed

tests/v1/worker/test_scheduler_tkv_limits.py
Daniel-Schenker: Addressed, also realized that the second set of prompt lengths was identical to the first. Only difference between the 2 sets of requests is the max tokens so consolidated that as well.

sendnn_inference/v1/core/scheduler.py
Daniel-Schenker: Addressed

tests/v1/worker/test_scheduler_tkv_limits.py
Daniel-Schenker: Added

Daniel-Schenker:
Initial comments have been addressed. I have re-confirmed that the new test case fails without scheduler.py changes and passes with changes.

sendnn_inference/v1/core/scheduler.py
```diff
@@ -20,6 +20,16 @@
...
 
+# Ensure that block_size is 64
+# This ensures the rounding function is correct
+assert SpyrePlatform.get_block_size() == 64
+
+
+def round_up_to_block_size(n: int) -> int:
+    # Helper function to round up to the nearest block size
+    # Uses bitwise alignment for better performance
+    return (n + 63) & ~63
```
yannicks1, line 31: spicy stuff 🌶️

sendnn_inference/v1/core/scheduler.py
yannicks1: `n_blocks` is unused, please remove

yannicks1 (commented):
To me it looks like this bug was (at least partially) introduced in https://github.com/torch-spyre/sendnn-inference/pull/913 (please correct me if I am wrong). Prior to 913 we passed `n_blocks` into the function which had information about the new request and did calculation based on this. While we fixed an edge case in 913, the new approach did not use `n_blocks`, hence the new request was not considered. 

With this PR we consider the new request again while still fixing the edge case bug.

yannicks1:
> Initial comments have been addressed. I have re-confirmed that the new test case fails without scheduler.py changes and passes with changes.

could you quickly run your newly added test (that captured the new failure) with a pre 913 commit? This would help me understand better...

sendnn_inference/v1/core/scheduler.py
Daniel-Schenker: Removed

yannicks1: this needs to be gone too ..

Daniel-Schenker: Removed

joerunde (approved):
ship it!

PR #946, opened 2026-04-21
[Migration] Sendnn inference rename

joerunde:
## Description

This PR updates the `vllm_spyre` package to `sendnn_inference`.

Docs and publication workflows have been updated in preparation to publish this as `sendnn-inference==2.0.0`

❗❗❗

Breaking configuration changes!
- The plugin name for `VLLM_PLUGINS` is now `sendnn_inference`
- All config options are now `SENDNN_INFERENCE_*`
- The precompiled model parser now expects `sendnn_inference_version`

## Related Issues

Closes torch-spyre/spyre-inference#8 

## Test Plan

- CPU unit tests
- Spyre unit tests
- Spot checks with full model integration tests
- Ensure Test PyPI deploy works

## Checklist

- [ ] I have read the [contributing guidelines](https://docs.vllm.ai/projects/spyre/en/latest/contributing)
- [ ] My code follows the project's code style (run `bash format.sh`)
- [ ] I have added tests for my changes (if applicable)
- [ ] I have updated the documentation (if applicable)
- [ ] My commits include a `Signed-off-by:` line (DCO compliance)

PR #956, opened 2026-04-30
Add documentation for the scheduler, padding logic and benchmarking interpretation

sducouedic:
## Description

This PR will help a new contributer understand the expected behavior of the scheduling logic and padding. There are two new pages in the docs:
- https://vllm--956.org.readthedocs.build/projects/spyre/en/956/contributing/scheduler.html describes the padding and scheduling logic (dummy blocks padding, padding to next block boundaries, scheduling prefill, ...)
- https://vllm--956.org.readthedocs.build/projects/spyre/en/956/user_guide/performance.html describes how to run benchmarks with `vllm bench serve` and interpret the results in relation to what is described in the scheduling section. The methodology for benchmarking multimodal model is very limited for the moment

### Details

Both pages contain visualizations coming from plotly. This is why there are many `.json` and `generate_***.py`. 

- The `.json` files were generated using `docs/mkdocs/hooks/generate_examples.py` locally. Meaning: I ran `generate_example.py` manually with some requests in mind, and then committed the output `.json` files, I also committed the script `generate_examples.py` for record). 

- The other `generate_*.py` scripts in `docs/mkdocs/hooks` generate **dynamically** (as hook) the `.html` files used in `scheduler.md` and `performance.md`

### Note for reviewers

I would encourage the reviewers to ignore the logic or correctness of the `.json` files and `generate_*.py` files, as they are all Bob and Claude generated, and just check for harmful/private content. We only need the **final dynamic plots to be correct**: please help verify that at every step the tkv, padding, and acceptance logic is correct (visually).

## Related Issues

Closes to #946 and #959

## Test Plan

None

## Checklist


joerunde:
Extremely 🌶️ !

I ran through the visualizations, they're really good. 10/10, no notes.

Though if I have to find a note, there seem to be `Play` buttons on the last two visualizations with no animations?

image

If it was intended to animate through the decode and prefill there then it seems broken, but the static view is fine as is too

joerunde (approved):
very very nice 😎

sducouedic:
thanks Joe, the plots that have only a play/stop button (no slider) are because there is only "one step". Let me try to fix it with Claude

docs/.nav.yml
```diff
@@ -34,7 +37,12 @@ nav:
...
     - Supported Models: user_guide/supported_models.md
+    - Performance Tuning: user_guide/performance.md
   - Developer Guide:
     - Contributing: contributing/README.md
     - Maintaining: contributing/maintaining.md
-    - Architecture: contributing/architecture.md
+    - Design Documents:
+      - Architecture: contributing/architecture.md
+      - Scheduling & Padding: contributing/scheduler.md
+    - MultiModal:
```
rafvasq: I think MultiModal was supposed to fall under "Design Documents" as well?

rafvasq (commented):
Wow, nice! LGTM aside from where MultiModal should go in `.nav`

rafvasq (approved):

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Max batch-tkv limit**: A hardware ceiling on the total KV-cache volume. The product `batch_size × max-tkv` must never exceed this limit at any point during execution.
+
+---
+
+## Padding
+
+The padding strategy is designed to meet two constraints specific to Spyre:
+
+- **Decode alignment:** All sequences in the decode batch must occupy the same number of KV-cache blocks at every step. Shorter sequences are padded with full dummy blocks on the left.
```
maxdebayser: Suggestion: "... on the left. Block 0 is reused for padding, so the memory used for padding blocks is a constant total of 64 bytes."

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+The scheduler uses continuous batching: new requests are prefilled in chunks while previously admitted requests continue decoding. Because Spyre imposes specific constraints on the KV-cache, the scheduler applies additional admission rules on top of vLLM's defaults, and pads requests to meet alignment requirements.
+
+---
+
+## Key concepts
+
+**Block size**: The number of tokens per KV-cache block.
+
+**Chunk size**: The number of tokens processed in a single prefill step. Always a multiple of the block size. A prompt longer than the chunk size is split across multiple successive steps.
```
maxdebayser: Suggestion: "the number of prompt tokens" for more clarity.

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+This page explains how sendnn-inference overrides the vLLM V1 scheduler for decoder (generation) models to respect Spyre hardware constraints, and how requests are padded during prefill and decode execution.
+
+!!! warning
+
+    All values and parameters (such as the chunk size) used in the figures are for illustration only; for the real configuration values refer to [configuration](../user_guide/configuration.md)
+
+## Overview
+
+The scheduler uses continuous batching: new requests are prefilled in chunks while previously admitted requests continue decoding. Because Spyre imposes specific constraints on the KV-cache, the scheduler applies additional admission rules on top of vLLM's defaults, and pads requests to meet alignment requirements.
```
maxdebayser: Suggestion: perhaps separate the concepts of continuous batching and chunked prefill.

"The scheduler uses continuous batching: new requests are prefilled while previously admitted requests continue decoding. To minimize the decoding delays, long prompts are split into chunks and the prefill is interleaved with decoding steps."

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+**Single chunk prefill (prompt len = 15):**
+
+<iframe src="../assets/plots/prefill_single_chunks.html" width="100%" height="450px" frameborder="0"></iframe>
+
+**Three chunks prefill (prompt len = 302):**
+<iframe src="../assets/plots/prefill_three_chunks.html" width="100%" height="450px" frameborder="0"></iframe>
+
+### Decode padding
+
+During decode, every request generates exactly one new token per step.
```
maxdebayser: Perhaps we could add here that no speculative decoding or jump-forward decoding is supported.

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+In addition to full-block left-padding, we also pad the rightmost block of each request (the block containing the tkv) up to its right boundary. These padded tokens are also masked out by the attention mechanism and do not affect outputs.
+
+#### Per-request tkv
+
+Each request's tkv is its left-padding offset plus the number of tokens computed so far, plus one for the token being generated in the current step. The max-tkv for a step is therefore determined by the request with the largest left-padding offset relative to its decode progress.
+
+#### tkv jumping
+
+When a request generates enough tokens to require an additional KV-cache block, it would need one more block than the rest of the batch. To keep all block tables the same width, one fewer dummy block is prepended instead. This causes that request's tkv to jump forward by one block in a single step, while all other requests' tkv values increase by one as usual.
```
maxdebayser: Is "one fewer dummy block" right or did you mean "one new dummy block"?

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+When the KV cache already contains blocks matching the beginning of a prompt, those blocks can be reused without recomputation.
+
+### Skipping full chunks
+
+Whole chunks whose blocks are entirely cached can be skipped. However, the last chunk of a prompt is always recomputed even if all its blocks are cached, because the model must produce the first generated token at the end of the last prefill step.
+
+### Boundary chunk
+
+The chunk that straddles the cache boundary — where some of its blocks are cached and some are not — is always fully recomputed. The KV writes for those blocks are redirected to a dummy block, leaving the cached values untouched while attention still reads from the real cached blocks.
```
maxdebayser: Maybe add at the end: "This prevents the early divergence of the block list of prompts with shared prefixes. Only the last block is duplicated. This replicates the vLLM KV-cache behavior on GPU."

maxdebayser (approved):
The animations are awesome! I've left a few ideas for the text here and there but I'm fine with merging as is. Thanks!

docs/.nav.yml
sducouedic: yes probably it fits there better

docs/contributing/scheduler.md
sducouedic: right, that's a pretty important detail

docs/contributing/scheduler.md
sducouedic: oh good catch, that sentence is super confusing actually. I confused myself there. When the request has left padding, we need to remove one block, if it has no left padding, all the other requests need to be added one block.

sducouedic:
thanks for the review all! @yannicks1 will also give a look before merging

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Block size**: The number of tokens per KV-cache block.
+
+**Chunk size**: The number of prompt tokens processed in a single prefill step. Always a multiple of the block size. A prompt longer than the chunk size is split across multiple successive steps.
+
+**Tkv**: The tkv stands for Token-Key-Value. For a given request, it is the padded token position up to which the KV cache is populated for that request.
+
+**Max-tkv**: The maximum tkv value across all requests currently in the batch.
+
+**Max batch-tkv limit**: A hardware ceiling on the total KV-cache volume. The product `batch_size × max-tkv` must never exceed this limit at any point during execution.
```
yannicks1, line 25: do we want to say here what the limit is in practice as it is hard coded and will never change (for performance reasons) ?

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+---
+
+## Padding
+
+The padding strategy is designed to meet two constraints specific to Spyre:
+
+- **Decode alignment:** All sequences in the decode batch must occupy the same number of KV-cache blocks at every step. Shorter sequences are padded with full dummy blocks on the left. Block 0 is always used for padding, so the memory used for padding is a constant total of 64 bytes.
+- **Prefix caching:** Variable-length left-padding inside a block would corrupt hash-based block comparisons. By aligning prompts to block boundaries (using dummy block ids) we avoid left-padding inside any block, keeping block contents identical for equal token sequences.
```
yannicks1: Is this spyre specific? I though also upstream prompts are left aligned to the block boundary. @joerunde

yannicks1: > By aligning prompts to block boundaries (using dummy block ids) we avoid left-padding inside any block

I do not understand the _(using dummy block ids)_ . the are not really used to have the first token of the prompt in a first slot of a block (left aligned). the right padding does ensure this... 

block level left padding with dummy blocks is to ensure chunk alignment (right aligned, such that we can read out of the last block of the last chunk).

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Max-tkv**: The maximum tkv value across all requests currently in the batch.
+
+**Max batch-tkv limit**: A hardware ceiling on the total KV-cache volume. The product `batch_size × max-tkv` must never exceed this limit at any point during execution.
+
+---
+
+## Padding
+
+The padding strategy is designed to meet two constraints specific to Spyre:
```
yannicks1, line 34: in my understanding the two "spyre" specific constraints that justfiy our padding strategy are: 
- rectangular decode block aka homogeneous tkv (which is your first point) -> requires left padding
- prefills (batch size 1) have to be chunk aligned and we can only read out the last block of the last chunk. -> requires left and right padding.

docs/contributing/scheduler.md
```diff
@@ -0,0 +1,181 @@
...
+
+All requests must share the same number of KV-cache blocks so that the block table is rectangular. The request with the most blocks sets the common width; shorter requests are left-padded with dummy blocks (i.e., any block from the block table). Dummy blocks are masked out by the attention mechanism and do not affect outputs. We always keep the number of left-padding blocks minimal, so when a long request that was forcing other requests to be left-padded finishes, we remove that left-padding for those requests.
+
+#### Right-padding until next block boundary
+
+In addition to full-block left-padding, we also pad the rightmost block of each request (the block containing the tkv) up to its right boundary. These padded tokens are also masked out by the attention mechanism and do not affect outputs.
+
+#### Per-request tkv
+
+Each request's tkv is its left-padding offset plus the number of tokens computed so far, plus one for the token being generated in the current step. The max-tkv for a step is therefore determined by the request with the largest left-padding offset relative to its decode progress.
```
yannicks1: > The max-tkv for a step is therefore determined by the request with the largest left-padding offset relative to its decode progress.

not really sure how to read this. can we phrase that simpler: The max-tkv for a step is therefore determined by the request with the largest offset in the last block?

yannicks1 (approved):
finished the padding section. couple of nits, but AWESOME work @sducouedic . 

I havent read the scheduler part yet, but feel free to merge already.

docs/contributing/scheduler.md
sducouedic: can't it increase in the future? but yes let me add it

docs/contributing/scheduler.md
sducouedic: ah what I wanted to say it is that on the left we use full blocks padding only

sducouedic: like there is left-padding smaller than a full-block, but I changed that

sducouedic: removed the parenthesis entirely

docs/contributing/scheduler.md
sducouedic: ah with your other comment I understand what you mean. So we have two Spyre-specific constraints:
- Decode alignment (-> let me rename to "Rectangular Decode") 
- Prefill alignment: like you say

Additional constraint (not Spyre specific) for prefix caching: left-alignment

docs/contributing/scheduler.md
sducouedic: see my other comment, changed to

- two spyre-specific constraints: rectangular decode and chunk-alignment prefill
- one prefix-caching constraint: left alignment to block boundary

docs/contributing/scheduler.md
sducouedic: I blame Claude for this nonsense

sducouedic:
Merging now as I will be out for a few days, we can address further comments in a follow-up PR

PR #961, opened 2026-05-08
[WIP] fix check_batch_tkv_limit_cp conservatively

yannicks1:
added an extra block of slack (left-padding can push a sequence's runtime tkv up to one block past the scheduler's estimate when the batch re-aligns on admission) for both new and decode requests.

### DIFF
--- a/sendnn_inference/v1/core/scheduler.py
+++ b/sendnn_inference/v1/core/scheduler.py
@@ -540,6 +540,9 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
         # Compute the effective token length of the new request
         # Rounded up to the nearest block size to account for potential padding
         new_req_max_tkv = round_up_to_block_size(new_req_tkv + request.max_tokens - 1)
+        # Extra block of slack: left-padding can push a sequence's runtime tkv up to
+        # one block past the scheduler's estimate when the batch re-aligns on admission.
+        new_req_max_tkv += self.block_size
 
         # Compute token lengths for all running requests (decode batch)
         decode_req_max_tkvs = []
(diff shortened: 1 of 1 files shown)


yannicks1:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=4

yannicks1:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=8

yannicks1:
bot:test

yannicks1:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=8

sendnn_inference/v1/core/scheduler.py
```diff
@@ -540,6 +540,9 @@ def check_batch_tkv_limit_cp(self, request: Request, new_req_tkv: int, running)
         # Compute the effective token length of the new request
         # Rounded up to the nearest block size to account for potential padding
         new_req_max_tkv = round_up_to_block_size(new_req_tkv + request.max_tokens - 1)
+        # Extra block of slack: left-padding can push a sequence's runtime tkv up to
+        # one block past the scheduler's estimate when the batch re-aligns on admission.
+        new_req_max_tkv += self.block_size
```
joerunde, line 545: dang, yeah my reasoning on only needing to round up was that the padding only happens when a request actually gets to the next block, so if our math is correct here then you'd only get to the next block if some request was going to end inside it, in which case that request would round up and cover all the cases of other requests getting padded to somewhere within that block.

But clearly that reasoning isn't sound 🙃

joerunde (approved):
ship it

sducouedic (approved):

sendnn_inference/v1/core/scheduler.py
yannicks1: I also thought that was okay, but somehow there were still certain workloads that would cause requests passing that violate the constraint. This is a dumb fix, just creating a safety margin, but a this stage I am a bit uncertain what else to do.

PR #962, opened 2026-05-12
chore: bumps vllm to v0.20.2

rafvasq:
## Description

Adds support for https://github.com/vllm-project/vllm/releases/tag/v0.20.2

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ dependencies = [
     "ibm-fms>=1.9.0,<2",
     # NB: use strict < with the next patch version to not exclude versions with
     # build metadata suffixes
-    "vllm>=0.19.1,<0.20.2",
+    "vllm>=0.19.1,<0.20.3",
     
     # Specific torch version overrides handled by uv
     "torch",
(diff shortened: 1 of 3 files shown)


rafvasq:
bot:test

pyproject.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
     "ibm-fms>=1.9.0,<2",
     # NB: use strict < with the next patch version to not exclude versions with
     # build metadata suffixes
-    "vllm>=0.19.1,<0.20.2",
+    "vllm>=0.19.1,<0.20.3",
```
joerunde, line 18: Do we need to update the compatibility test matrix to have tests for 0.20 and 0.20.1?

rafvasq: Hm, shouldn't it be okay not to since they were already tested in their respective support PRs? Is the idea that code changes introduced could potentially break v0.20.0 support and not v0.20.2?

joerunde: Right, we'd want to make sure that these or any future code changes that we make don't break support for any intermediate versions. We put the test matrix in place because that happened once before, where the lowest supported version and the highest supported version both worked, but the intermediate versions were busted.

joerunde (approved):
nice!

PR #963, opened 2026-05-13
Granite vision 3.3 2B enablement

San-Nguyen:
## Description

Add the Granite-Vision-3.3-2B model configurations in the yaml file for one card usage. The max_model_lens and max_num_seqs placed is what I found as the highest value to run on a single card

## Test Plan

1)Create a container
``podman run -d --entrypoint '["sleep","infinity"]' --device=/dev/vfio --privileged=true --pids-limit 0 --userns=keep-id --group-add=keep-groups --memory 500G --shm-size 64G -v ${HOST_MODELS_DIR}:/models -e AIU_PCIE_IDS="${AIU_IDS}" -e SENDNN_INFERENCE_REQUIRE_PRECOMPILED_DECODERS=0 -p 127.0.0.1::8000 <IMAGE_ID>``
2)Added the model configuration to model_configs.yaml in the container
3) Ran Power's start up script for vllm. 
`container-scripts/vllm_start_ppc64le.sh -tp 1 --model /models/granite-vision-3.3-2b --max-model-len 8192 --max-num-seqs 16`
4) When the vllm server is ready, run a curl command to see if the inference/vision worked. (download a small jpg file)
``` 
IMAGE_BASE64=$(base64 -i /home/snguy/scripts/post.jpg | tr -d '\n')

curl -X POST http://127.0.0.1:37287/v1/chat/completions \
  -d '{
    "model": "/models/granite-vision-3.3-2b/",
    "messages": [
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "What is in this image?"
          },
          {
            "type": "image_url",
            "image_url": {
              "url": "data:image/jpeg;base64,'"$IMAGE_BASE64"'"
            }
          }
        ]
      }
    ],
    "max_tokens": 512
  }' 
  ```

## Checklist

### CODE REVIEW

File: sendnn_inference/config/model_configs.yaml [hunk 1]
```diff
@@ -222,6 +236,14 @@ models:
         max_num_seqs: 32
         device_config: *granite_8b_tp4_device_config
 
+  # Granite Vision 3.3 2B
+  ibm-granite/granite-vision-3.3-2b:
+    architecture: *granite_vision_33_2b_architecture
+    continuous_batching_configs:
+      - tp_size: 1
+        max_model_len: 8192
+        max_num_seqs: 16
```
joerunde, line 256: This is exactly 128k so it doesn't need the TKV override, but would we want to apply a value for `num_gpu_blocks_override`?

This will default to 128k/64 = 2k blocks, though if there is still space on the card for more then up to 8k should be supportable.

(Not sure if that's really relevant here though since honestly I don't even know if we support prefix caching with multimodal models)

joerunde:
Should we toss in a 2 or 4 card config as well while we're at it? Or are y'all only interested in single-card deployments for this?

San-Nguyen:
> Should we toss in a 2 or 4 card config as well while we're at it? Or are y'all only interested in single-card deployments for this?

I believe initially we're mostly looking at one card currently, but I'm trying to collect some performance data for 1 card vs 2 cards to see the benefits, but these decision will be made later when after analyzing the perf. data between 1/2/4-cards.

sendnn_inference/config/model_configs.yaml
San-Nguyen: If those specific lines are not needed, then I can remove it. Trying to increase the `num_gpu_blocks_override` to 8k causes an error: `DtException: Need to find a valid memory space for KV cache`

joerunde (approved):
lpgtm if we want to get the single-card config merged and documented

joerunde (changes requested):
ah- actually the failing tests show the change required here:
```
ValueError: Model 'ibm-granite/granite-vision-3.3-2b' must have at least one runtime configuration (either static_batching_configs or continuous_batching_configs)
```

@San-Nguyen as we talked in standup, the `continuous_batching_configs` are required. 

Specifically for the 16x8k case we do not need the tkv override because it's <= 128k, but for any larger 2 or 4 card case we would need to set both the tkv and kv cache overrides

San-Nguyen:
> ah- actually the failing tests show the change required here:
> 
> ```
> ValueError: Model 'ibm-granite/granite-vision-3.3-2b' must have at least one runtime configuration (either static_batching_configs or continuous_batching_configs)
> ```
> 
> @San-Nguyen as we talked in standup, the `continuous_batching_configs` are required.
> 
> Specifically for the 16x8k case we do not need the tkv override because it's <= 128k, but for any larger 2 or 4 card case we would need to set both the tkv and kv cache overrides

Re-added the continuous batching config back in, we can just merge the 1 card for now and can add the detail for 2 cards usage later if we (Power) do see benefits with running it.

San-Nguyen:
Added two card config since I/we saw some general improvements when comparing to 1 card

sendnn_inference/config/model_configs.yaml
```diff
@@ -222,6 +236,17 @@ models:
...
+      - tp_size: 2
+        max_model_len: 16382
+        max_num_seqs: 16
```
joerunde, line 259: I think we'll probably want to have:
```
device_config:
  env_vars:
      VLLM_DT_MAX_BATCH_TKV_LIMIT: 131072  # 128k
```

here, because 16x16k > 128k which is the largest batch size that the compiler team claims to support.

That said, if this works empirically then maybe it's not 100% necessary, but any programs larger than 128k for sure haven't been tested on lower-level correctness tests like DPP

joerunde (approved):
we should probably add in the tkv limit for the 2 card case, but otherwise lgtm

San-Nguyen:
@joerunde > we should probably add in the tkv limit for the 2 card case, but otherwise lgtm
Added the tkv limit for 2 cards and 4 cards as well. Tested both and they worked on Power

sendnn_inference/config/model_configs.yaml
```diff
@@ -36,6 +36,31 @@ _templates:
...
+     hidden_size: 2048
+     vocab_size: 49156
+     num_key_value_heads: 8
+     num_attention_heads: 32
+   vision_config:
+     num_hidden_layers: 27
+     hidden_size: 1152
+     num_attention_heads: 16
+
+  ggranite_vision_2b_tp2_device_config: &granite_vision_2b_tp2_device_config
```
joerunde: `ggranite` looks like a typo here?

probably doesn't matter, but would be nice to fix up

San-Nguyen: Went and fixed the typo

joerunde (approved):
ship it

PR #969, opened 2026-05-20
Upgrade fms-mo and aftu to remove llmcompressor dependecy

maxdebayser:
llmcompressor is a transitive dependency and it's not required for infrence. It is however causing problems because it has a dependency to an older compressed-tensors library than vllm, causing dependency conflicts. It also is not compatible with transformers 5 yet.


gkumbhat (approved):
Looks good to me!

Daniel-Schenker (approved):
lgtm

PR #973, opened 2026-05-21
Implement request pausing

maxdebayser:
## Description

vLLM supports preemption of requests when it runs out of KV cache blocks but the preempted requests must be re-computed from scratch. In `sendnn-inference` there are other constraints that are tighter than the number of blocks so we want the  "pause" or request from the currently executing batch without relinquishing the blocks that it is holding. In this way we can schedule requests optimistically.


maxdebayser:
Maybe `can_schedule_prefill` could check if there are paused requests so that we don't add new requests while the decode batch is over commited.

sendnn_inference/v1/worker/spyre_input_batch.py
```diff
@@ -532,6 +539,146 @@ def remove_request(self, req_id: str):
 
         self.bad_words_token_ids.pop(req_index, None)
 
+    def pause_request(self, req_id: str) -> None:
```
maxdebayser, line 464: Once all the tests are passing and we're satisfied with the design and implementation, I think we can factor out the common code between remove_request/pause_request and add_request/resume_request

sendnn_inference/v1/core/scheduler.py
```diff
@@ -513,84 +502,171 @@ def _has_scheduling_priority(self, request):
...
+        if not running_requests:
+            return 0
+
+        # Step 1: Find the maximum number of blocks across all requests
+        # Account for requests that will need a new block after the next token
+        max_n_blocks = 0
+        for request in running_requests:
+            block_ids_per_kv_cache_group = self.kv_cache_manager.get_block_ids(request.request_id)
+            assert len(block_ids_per_kv_cache_group) == 1
+            num_blocks = len(block_ids_per_kv_cache_group[0])
```
maxdebayser: Since here we don't need to distinguish between cached and non-cached blocks, I think `num_blocks = math.ceil(request.num_computed_tokens, self.block_size)` is also fine and potentially cheaper to compute.

sendnn_inference/v1/core/scheduler.py
```diff
@@ -513,84 +502,171 @@ def _has_scheduling_priority(self, request):
...
+    def predict_next_decode_tkv(self, running_requests: list[Request]) -> int:
+        """
+        Predicts the TKV after the next decode step for a given batch of running
+        requests.
+
+        This method replicates the TKV calculation logic from the model runner's
+        _prepare_decode method, accounting for:
+        - Block alignment (left-padding to make batch rectangular)
+        - The next token that will be generated (+1)
+        - Maximum TKV across all requests in the batch
```
sducouedic, line 773: This function is obviously wrong. When doing a test run of 500 requests it happened once that during inference we overflow the max_tkv_batch_limit:
```
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949] WorkerProc hit an exception.
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949] Traceback (most recent call last):
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]   File "/opt/vllm/lib64/python3.12/site-packages/vllm/v1/executor/multiproc_executor.py", line 944, in worker_busy_loop
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]     output = func(*args, **kwargs)
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]              ^^^^^^^^^^^^^^^^^^^^^
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]   File "/opt/vllm/lib64/python3.12/site-packages/vllm/v1/worker/worker_base.py", line 332, in execute_model
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]     return self.worker.execute_model(scheduler_output)
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]   File "/opt/vllm/lib64/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]     return func(*args, **kwargs)
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]            ^^^^^^^^^^^^^^^^^^^^^
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]   File "/home/senuser/sendnn-inference/sendnn_inference/v1/worker/spyre_worker.py", line 785, in execute_model
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]     output = self.model_runner.execute_model(scheduler_output)
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]   File "/opt/vllm/lib64/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]     return func(*args, **kwargs)
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]            ^^^^^^^^^^^^^^^^^^^^^
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]   File "/home/senuser/sendnn-inference/sendnn_inference/v1/worker/spyre_model_runner.py", line 1565, in execute_model
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949]     self.tkv * len(scheduler_output.num_scheduled_tokens)
(Worker_TP3 pid=72446) ERROR 06-01 13:55:47 [multiproc_executor.py:949] AssertionError: Exceeded max batch tkv limit 131072! tkv: 26235, batch_size: 5
```

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1447,6 +1437,38 @@ def _update_batch(self, scheduler_output: SchedulerOutput):
...
+                self.input_batch.refresh_metadata()
+                need_metadata_refresh = False
+
+        # Find requests that are in scheduler output but not in input_batch
+        # (restore from pausing)
+        restored_req_ids = scheduled_req_ids - current_batch_req_ids
+        for req_id in restored_req_ids:
+            # Only restore requests that were previously paused
+            if req_id in self.paused_req_ids and req_id in self.requests:
+                logger.info("Restoring paused request %s to input_batch", req_id)
```
sducouedic: happens too often to be logged as info
```suggestion
                logger.debug("Restoring paused request %s to input_batch", req_id)
```

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1447,6 +1437,38 @@ def _update_batch(self, scheduler_output: SchedulerOutput):
...
+        scheduled_req_ids = set(req_data.req_ids)
+        current_batch_req_ids = set(self.input_batch.req_id_to_index.keys())
+        need_metadata_refresh = True
+
+        # Find requests that are in input_batch but not in scheduler output (paused)
+        paused_req_ids = current_batch_req_ids - scheduled_req_ids
+        for req_id in paused_req_ids:
+            # Only pause if it's not a finished request (finished requests are handled separately)
+            if req_id not in (scheduler_output.finished_req_ids or []):
+                logger.info("Pausing request %s from input_batch", req_id)
```
sducouedic: use logger.debug

tests/e2e/test_spyre_pc_scheduler_steps.py
```diff
@@ -397,7 +397,7 @@ def test_prefix_hit_decoded_block_within_batch(
             "step": 68,
             # seq 1 tkv (193) is in 4th block. Need to pad seq 0 tkv to 4th
             # block as well: 192 + 64 = 256
-            "tkv": 256,
+            "tkv": 193,
```
sducouedic: Need to adapt the comment for this one

sendnn_inference/v1/sample/spyre_logits_processor.py
```diff
@@ -91,25 +155,39 @@ def update_state(self, batch_update: BatchUpdate | None):
                     BatchUpdate(batch_size=1, removed=[0], moved=[], added=[])
                 )
 
+            for index, req_id in getattr(batch_update, "resumed", []):
+                assert req_id in self._saved
+                self.logitprocs[index] = self._saved.pop(req_id)
```
sducouedic, line 160: [optimization]
When we pop, we basically overwrite whatever was at this index (`self.logitsproces[index]`), and this will trigger the gb if there was a logitsprocess at this index. 
The code as it is now introduces many creation / deletion of logitprocessors. Maybe we should find a way to create a minimal number of new logitsprocessor (call `self._factory()` only when strictly necessary)

sendnn_inference/v1/core/scheduler.py
```diff
@@ -513,84 +502,171 @@ def _has_scheduling_priority(self, request):
...
+
+        # If we didn't have to remove any requests, try to add back previously
+        # paused requests (oldest first) as long as they fit within constraints
+        if not had_to_remove:
+            while self.paused_decoding_requests:
+                # Try adding the oldest paused request (first in list)
+                request_to_add = self.paused_decoding_requests[0]
+                test_requests = decoding_requests + [request_to_add]
+
+                if self._can_decode_all_requests(test_requests):
```
sducouedic: [optimization] the tkv might jump back during decoding, even when no request finishes (see visualization here: https://docs.vllm.ai/projects/spyre/en/latest/contributing/scheduler.html#visualization-decode-padding). This leads to many pausing/resuming of the same few requests in a short timespan, due to tkv jittering by a magnitude of 64. I think we need to adapt `_can_decode_all_requests` to consider a marge of 64.

sducouedic:
I noticed the median ITL increasing a lot, we should address it before merging. Possible reasons:
* logging every time we pause/resume
* creation / deletion of logitsprocessor
* tkv-jump by += 64 every time we reach a new blocks, leading to many pausing-resuming in a short timespan
* start new requests even if there is a paused one

sendnn_inference/v1/worker/spyre_input_batch.py
sducouedic: done

sendnn_inference/v1/core/scheduler.py
sducouedic: solved by using `math.ceil(request.num_computed_tokens + 1 / self.block_size)` instead of `self.kv_cache_manager.get_block_ids(request.request_id)` in predict_next_tkv

yannicks1:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=8
ENABLE_PC=0

sendnn_inference/v1/core/scheduler.py
```diff
@@ -590,84 +577,132 @@ def _has_scheduling_priority(self, request):
...
+            self.running.remove(request_to_remove)
+            self.paused_decoding_requests.append(request_to_remove)
+            logger.info("Request %s paused due to batch TKV limit ", request_to_remove.request_id)
+
+        # It shouldn't be possible to remove all requests if we started with some
+        assert not initial_had_requests or len(decoding_requests) > 0
+
+        # If we didn't have to remove any requests, try to add back previously
+        # paused requests (oldest first) as long as they fit within constraints
+        if not had_to_remove:
```
yannicks1: how expensive is this while loop? 

I guess we could/should run it regardless `if not had_to_remove`. 

Reason: it is possible that we removed the longest request above and there might actually now be space for one or more short ones to be scheduled in the current batch. With the current strategy of removing the request with the least number of decoded tokens, this can happen occasionally, but when changing the strategy of removal to e.g. removing the longest one, there could be quite some perf on the table here.. 

wdyt? @maxdebayser @sducouedic

sendnn_inference/v1/core/scheduler.py
```diff
@@ -590,84 +577,132 @@ def _has_scheduling_priority(self, request):
...
+            while self.paused_decoding_requests:
+                # Try adding the oldest paused request (first in list)
```
yannicks1: Similarly to the removal strategy, I guess also alternative re-adding strategies could be explored. 
While the current strategy 'preserves' request priority, to optimize fully for system throughput one might want to not greedily operate on this list, but rather try to add as many requests as possible. This could be achieved for e.g. by allowing to skip the head of `self.paused_decoding_requests` (we currently break out of the loop if the oldest cannot be re-added) or sorting the list by tkv (we would have to find another way of preserving some priority though...) 

But I like this simple strategy as a starting point!

sendnn_inference/v1/worker/spyre_input_batch.py
```diff
@@ -398,14 +417,114 @@ def add_request(
...
+        # Emit shift moves so every other processor (MinP, LogitBias, …) sees
+        # the correct dense-index after the gap is closed.
+        end_dense_idx = min(self._num_requests, self.max_num_reqs - 1)
+        for tmp_dense in range(dense_index, end_dense_idx):
+            self.batch_update_builder.moved.append(
+                (tmp_dense, tmp_dense + 1, MoveDirectionality.UNIDIRECTIONAL)
+            )
+
+            self.req_output_token_ids.pop(dense_index)
+            self._unregister_sampling_params(req_id, req_index)
```
yannicks1: this should be moved out of the loop!?

sendnn_inference/v1/core/scheduler.py
yannicks1, line 31: rmv, this function is now not used anymore as the volumetric check has been refactored

sendnn_inference/v1/core/scheduler.py
```diff
@@ -194,6 +190,10 @@ def __init__(self, *args, **kwargs) -> None:
         # keep a list to be able to batch prefills in the future.
         self.ongoing_prefills: list[Request] = []
 
+        # Track requests that were temporarily paused from decoding due to
+        # batch TKV constraint and moved back to waiting queue
+        self.paused_decoding_requests: list[Request] = []
```
yannicks1, line 205: minor: consider using `collections.deque` instead of python list (like we do for holdback and skip queue) to pop the head in O(1) instead of O(n).

yannicks1 (commented):
A first batch of feedback. I think in `SamplingInputBatch.pause_request()` there really is a bug that needs fixing, the rest is probably less important. 

Need to continue later as there is a lot of change:)

sendnn_inference/v1/core/scheduler.py
maxdebayser: I think this makes sense. We should add this.

sendnn_inference/v1/worker/spyre_input_batch.py
maxdebayser: oops

sendnn_inference/v1/core/scheduler.py
maxdebayser: Due to the change required to implement the `had_to_remove` suggestion the access pattern has changed a little bit

sendnn_inference/v1/sample/spyre_logits_processor.py
```diff
@@ -91,25 +155,39 @@ def update_state(self, batch_update: BatchUpdate | None):
...
+
+            for index, req_id in getattr(batch_update, "paused", []):
+                self._saved[req_id] = self.logitprocs[index]
+                self.logitprocs[index] = self._factory()
+
+            for req_id in getattr(batch_update, "finished_paused", []):
+                # Max: I think we can't assume that the request will
+                # be here because it could be a cancelled request that
+                # never made it into the batch.
+                self._saved.pop(req_id, 0)
```
joerunde: style: `self._saved.pop(req_id, None)`

since we don't care about the result, just ensuring it's gone from _saved

sendnn_inference/v1/core/scheduler.py
```diff
@@ -424,6 +424,9 @@ def schedule(self) -> "SchedulerOutput":
             self.previous_step_was_prefill = False
             running_holdback = []
 
+        if not self.previous_step_was_prefill:
+            self._handle_decode_requests_pausing()
```
joerunde, line 445: is this condition correct? What if a prefill step just finished and then added a request to the batch that made it too big?

sendnn_inference/v1/core/scheduler.py
```diff
@@ -590,84 +577,132 @@ def _has_scheduling_priority(self, request):
...
+        # the tkv should never get beyond max_model_len
+        assert next_predicted_tkv <= self.max_model_len
+
+        # check batch tkv limit: batch_size * predicted_tkv must not exceed limit
+        batch_size = len(decoding_requests)
+        predicted_batch_tkv = batch_size * next_predicted_tkv
+
+        return predicted_batch_tkv <= self.max_batch_tkv_limit
+
+    def _handle_decode_requests_pausing(self) -> None:
```
joerunde, line 719: It seems like the logic for pausing and resuming is independent, so we could split this to have more readable naming:

```
def pause_decoding_requests(self) -> None:
    ...

def resume_decoding_requests(self) -> None:
    ...
```

then we could put these in different places in the scheduler code if it makes sense, e.g. calling resume just before looking through the waiting queue.

Also if we call resume before pause, then we don't have to worry about the `if not had_to_remove` check, because if you were able to resume requests then you certainly don't need to pause them

sendnn_inference/v1/core/scheduler.py
maxdebayser: I think it is. The `previous_step_was_prefill` variable name is a bit misleading because it actually means that in this current step we're about to schedule a prefill.

sducouedic: yes `previous_step_was_prefill` means this step really, it was named `previous` because originally only `update_from_output` needed to read this variable

yannicks1: was also confused by this. should we rename it? or at least put a comment here?

sducouedic: done

sendnn_inference/v1/core/scheduler.py
sducouedic: done: `_maybe_pause_decoding_requests` and `_maybe_resume_decoding_requests` introduced

sendnn_inference/v1/core/scheduler.py
sducouedic: should be addressed now: `_maybe_resume_decoding_requests` tries all the requests in the paused queue, and runs independently of whether requests had to be removed

sendnn_inference/v1/core/scheduler.py
sducouedic: should be addressed now: `_maybe_resume_decoding_requests` tries all the requests in the paused queue

tests/e2e/test_spyre_decode_pause_scheduler_steps.py
```diff
@@ -0,0 +1,455 @@
...
+
+This tests the relaxed constraint checking where requests are scheduled if
+prefill constraints are satisfied (not future constraints). Requests that
+would violate constraints during decode will be paused at that time.
+
+The two main constraints checked at prefill time are:
+1. Max-context constraint: current tkv <= max_context_len
+2. Volumetric constraint: current_max_tkv * batch_size <= max_batch_tkv_limit
+
+Run `python -m pytest tests/e2e/test_spyre_holdback_scheduler_steps.py`.
```
joerunde: wrong file name

(But I don't really think we need this in our comments anyway)

sendnn_inference/v1/metrics/stats_logger.py
```diff
@@ -186,11 +191,103 @@ def estimate_prefill_interrupt_lower_bound(
...
+        scheduler_stats: SchedulerStats | None,
+        iteration_stats: IterationStats | None,
+        mm_cache_stats: MultiModalCacheStats | None = None,
+        engine_idx: int = 0,
+    ):
+        """Log to prometheus."""
+        if scheduler_stats is not None and scheduler_stats.kv_connector_stats:
+            sendnn_stats = scheduler_stats.kv_connector_stats.pop("sendnn-stats")
+            if sendnn_stats is not None:
+                sendnn_stats = ChunkedPrefillSpyreSchedulerStats(**sendnn_stats)
```
joerunde: sendnn_stats here might already be an instance of `ChunkedPrefillSpyreSchedulerStats` here and not a dict? I thought we set it in the scheduler like

```
            base_stats.kv_connector_stats["sendnn-stats"] = ChunkedPrefillSpyreSchedulerStats(
                decode_batch_size=decode_batch_size,
                num_paused_reqs=num_paused_reqs,
                pause_events=self.pause_events,
                resume_events=self.resume_events,
            )
```

Is this serializing to a dict under the hood or something? Otherwise the `**` unpacking should fail on a dataclass instance

sendnn_inference/v1/core/scheduler.py
```diff
@@ -600,84 +591,144 @@ def _has_scheduling_priority(self, request):
...
+        """
+        Resumes previously paused requests (oldest first) when TKV capacity is available.
+        Mutates both decoding_requests and self.running.
+
+        Returns the number of requests resumed.
+        """
+        num_resumed = 0
+
+        # Reverse iteration: pop(i) only shifts indices above i, which are already visited.
+        for i in range(len(self.paused_decoding_requests) - 1, -1, -1):
```
joerunde: Isn't this the opposite of the docstring? It claims oldest first, but we're iterating in reverse and the last item in `self.paused_decoding_requests` would be the newest paused request, not the oldest. 

Shouldn't this be a dequeue, or at least documented why it isn't?

tests/v1/core/test_scheduler_structured_outputs.py
```diff
@@ -46,6 +46,7 @@ def mocked_scheduler():
     scheduler.skipped_waiting = FCFSRequestQueue()
     scheduler.running = []
     scheduler.ongoing_prefills = []
+    scheduler.paused_decoding_requests = []
     scheduler.chunk_size = 128
     scheduler.do_interleaving = False
     scheduler.previous_step_was_prefill = False
```
joerunde: isn't this `step_is_prefill` now?

sendnn_inference/v1/core/scheduler.py
```diff
@@ -727,14 +785,31 @@ def make_stats(self, *args, **kwargs) -> SchedulerStats | None:
...
+            # Abuse the kv_connector_stats field to store the spyre stats.
+            # We can open an upstream PR to add another extensible field.
+            base_stats.kv_connector_stats["sendnn-stats"] = ChunkedPrefillSpyreSchedulerStats(
+                decode_batch_size=decode_batch_size,
+                num_paused_reqs=num_paused_reqs,
+                pause_events=self.pause_events,
+                resume_events=self.resume_events,
+            )
+        self.pause_events = 0
+        self.resume_events = 0
```
joerunde: should these be indented so they're only cleared after they're recorded on a not-None `base_stats`?

sendnn_inference/platform.py
```diff
@@ -214,7 +214,7 @@ def get_total_spyre_blocks(cls, vllm_config: VllmConfig) -> int:
 
     @classmethod
     def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
-        # 🌶️🌶️🌶️ Patch in our perf logger before the engine is created
+        # 🌶🌶🌶 Patch in our perf logger before the engine is created
```
joerunde, line 221: interesting

sendnn_inference/v1/metrics/stats_logger.py
maxdebayser: It can be deserialized to a dict when the engine core runs in a different process. The MsgPack decoder on the api server side doesn't have the full type information.

tests/v1/core/test_scheduler_structured_outputs.py
maxdebayser: Good catch!

sendnn_inference/platform.py
maxdebayser: Very strange

sendnn_inference/v1/core/scheduler.py
maxdebayser: For now I've just updated the docstring, but this is a good discussion to have.

With dequeue deleting elements in the middle would be more complicated, I think, and since we don't add new request while there are still paused requests, I'm not sure if the order matters. Wdyt, @sducouedic ?

sducouedic: ah yes, good catch. The reverse iteration is to avoid pop bug. yes I think we can just update the docstring, there is rarely more than one paused at a same time

maxdebayser:
bot:test

maxdebayser:
bot:test

sducouedic (approved):
My approval shouldn't really count, since a implemented these changes for a big part 😅 but for me it looks good

joerunde:
@maxdebayser claude tried to tell me there were a few bugs, but then it failed to exercise them so it passes the image test.

I think the only big picture thing left to talk about is metrics- if we're going to enable these custom `sendnn` metrics then shouldn't we have a bit of documentation for them? Also a unit test that ensures they at least exist would be great, that way if our slightly-hacky integration ever does get broken by vllm we'll catch it in CI

maxdebayser:
@joerunde , I've added `tests/e2e/test_spyre_decode_pause_scheduler_steps.py::test_metrics` to check that the metrics continue working.

joerunde (approved):
lpgtm!

PR #974, opened 2026-05-21
Implement per arch defaults for quantized tolerance

Daniel-Schenker:
## Description

This PR implements architecture specific default values for the absolute tolerance used for quantized models for unit tests. Power and Z saw failures due to this tolerance not being high enough. Model accuracy was verified independently. 

## Test Plan

Run sendnn_inference unit tests, observe all tests passing on Power and Z.

## Checklist


prashantgupta24 (approved):
LGTM.

Copy-pasting what Joe said in the internal channel in case we want to think about making this better in the future:

### alternative solutions:

- (medium cost) Find lower-entropy prompts that will have lower logprob differences in general
- (high cost) swap from comparing against hf implementations to comparing against fms implementations, to remove the drift from the differences in model code

PR #976, opened 2026-05-22
Remove max-model-len constraint

### LINKED ISSUE #975: Do we still need the max-context-len constraint?
I was wondering the the max-context-len constraint is still needed. I suspect that the tkv will naturally never violate max-context-len, even without the constraint.

- Requests where `prompt_len + max_tokens > max_context_len` are automatically rejected by the engine before reaching the scheduler. So we know our requests are not longer than max_context_len
- Left-padding is full blocks only (no variable in-block padding)
- Left-padding blocks are dynamically added and **removed** to match the block of the longest-decoding sequence, which itself can never go beyond `max_context_len`
- specifically, as soon as a long sequence finishes, the left-padding of the other requests are removed

I highly suspect now that this constraint isn't needed anymore, now that we removed variable left-padding. This would greatly simplify the code, and will leave us only with the volumetric constraint to address. This probably won't give any performance gain, as showed by the runs from @yannicks1, the max-context-len constraint is never fired currently anyway. So only code cleaning

sducouedic:
## Description

This PR removes max-model-len constraint. Let's run tests and see if everything works correctly. 

Closes https://github.com/torch-spyre/sendnn-inference/issues/975

## Test Plan

I updated the tests to show the situation where `max_tokens` goes beyond `max_context_len` due to left-padding, but that gets removed before the tkv can go beyond `max_context_len`. I also ran some benchmarking where half of the requests are with requested output tokens `32k - prompt_len`

## Checklist


sducouedic:
bot:test

sducouedic:
bot:test

sducouedic:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=4
AGENT=IMS

sducouedic:
bot:test

sducouedic:
bot:test

sducouedic:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=4

sducouedic:
bot:test

joerunde (approved):
lpgtm!

I agree that this check existed from before a time where we could manage the left-padding on the fly like we do now, and so it should be safe to remove.

Thank you for all the thorough tests and benchmarks!

joerunde:
bot:test

Just triple-checking some unit tests on the latest commit here 🤞

joerunde:
bot:test

CI cluster being sad :(

yannicks1 (approved):
lgtm. I am confident too that this can be removed. Thanks for addressing this!

Just a thought from a safety perspective: would it make sense to merge this PR with the pre-emption batch of PRs? (not suggesting to make it part of any other PRs, just suggesting to merge it once we merge the pre-emption). While this is nice negative diffs and code simplification, we measured no perf improvement associated with this. Hence merging it now only adds a (probably very tiny) risk of introducing a bug until pre-emption arrives and this change becomes a necessity. 
wdyt @sducouedic @joerunde @maxdebayser ?

maxdebayser (approved):

maxdebayser:
@yannicks1 I prefer self-contained incremental PRs, it makes it easier to review and merge. Also, if we find bugs later on, it makes `git bisect` more effective because it can find the specific change that caused a bug.

joerunde:
If we're just talking about delaying merging until pre-emption hits, then yeah that could be a good way to reduce our defect surface.

But on the other hand, we don't want PRs to hang out and accumulate merge conflicts either. I'd probably still lean merging this first, if we think that the pre-emption PRs are almost ready to go

PR #977, opened 2026-05-25
Hybrid hardware for running MM model to optimize image processing

nikheal2:
Image encoding is currently not supported on spyre. We are looking to utilize torch.nnpa library to run image encoding on telum II.

Added a new env var SENDNN_INFERENCE_MM_DEVICE (defaults to cpu, set to nnpa to opt in) that routes the Pixtral/SigLIP vision tower + multi-modal projector onto the telum-II chip via the torch_nnpa privateuse1 backend, while the Ministral/Granite LLM continues to run on Spyre via torch.compile(backend="sendnn").

## Test Plan

Benchmarking shows 1.75x improvement with NNPA

## Checklist


sendnn_inference/utils.py
```diff
@@ -56,3 +56,27 @@ def parse_cpu_mm_dtype(value: str) -> torch.dtype:
             f"SENDNN_INFERENCE_CPU_MM_DTYPE must be one of {list(_CPU_MM_DTYPES)}, got {value!r}"
         )
     return getattr(torch, value)
+
+
+_MM_DEVICES = ("cpu", "nnpa")
```
dilipgb: Load device data from api instead of using static mappings

sendnn_inference/utils.py
```diff
@@ -56,3 +56,27 @@ def parse_cpu_mm_dtype(value: str) -> torch.dtype:
...
+    under the name "nnpa" before any tensors are constructed on it.
+    """
+    value = value.lower()
+    if value not in _MM_DEVICES:
+        raise ValueError(
+            f"SENDNN_INFERENCE_MM_DEVICE must be one of {list(_MM_DEVICES)}, got {value!r}"
+        )
+    if value == "nnpa":
+        try:
+            import torch_nnpa  # noqa: F401  # side-effect: registers privateuse1 -> 'nnpa'
```
dilipgb: instead of noqa, can you check if device registration is successful?

sendnn_inference/envs.py
```diff
@@ -152,6 +153,14 @@ def clear_env_cache():
             _CPU_MM_DTYPE_PLATFORM_DEFAULTS.get(platform.machine(), "float16"),
         )
     ),
+    # Device used to execute the multimodal vision_tower / multi_modal_projector.
+    # "cpu" (default) keeps the historical behaviour. "nnpa" routes the encoder
+    # to the IBM Z Integrated Accelerator for AI through the torch_nnpa
+    # privateuse1 backend; the LLM forward continues to run on Spyre via
+    # torch.compile(backend="sendnn") regardless of this setting.
+    "SENDNN_INFERENCE_MM_DEVICE": lambda: parse_mm_device(
+        os.getenv("SENDNN_INFERENCE_MM_DEVICE", "cpu")
```
dilipgb: Maybe once we productize, Add it to TODO to remove env variables and use nnpa default and if not available fallback to CPU.

sendnn_inference/multimodal/mm_mappings/mistral3.py
```diff
@@ -72,6 +73,15 @@ def get_maybe_mm_embeddings(
                 # If squeezed during spec building, add it back
                 if pixel_values.ndim == 3:
                     pixel_values = pixel_values.unsqueeze(0)
+                # Move pixel_values onto the same device/dtype as the
+                # vision_tower params so the encoder forward can run there
+                # (NNPA / privateuse1 backend, or CPU). The merge with
+                # text_embeds happens later inside FMS, where the merged
+                # output ends up on text_embeds.device (CPU) automatically.
+                mm_device = envs_spyre.SENDNN_INFERENCE_MM_DEVICE
```
dilipgb: Update TODO here as well.

dilipgb (changes requested):

sendnn_inference/model_executor/model_loader/spyre.py
```diff
@@ -268,20 +268,37 @@ def load_weights(
...
         cpu_mm_dtype = envs_spyre.SENDNN_INFERENCE_CPU_MM_DTYPE
+        mm_device = envs_spyre.SENDNN_INFERENCE_MM_DEVICE
         mm_prefixes = self.mm_model_utils.mm_parameter_prefixes if self.mm_model_utils else ()
 
+        # Move the multimodal submodules whole. We must use Module.to(...) here
+        # rather than mutating param.data, because nn.Parameter.set_data refuses
+        # to swap tensor types when the backend changes (e.g. CPU TensorImpl ->
+        # NnpaTensorImpl). Module._apply rebuilds Parameter objects correctly
+        # in that case. As a bonus this also moves any buffers under the
+        # submodule.
```
joerunde: Can we use this strategy for the bf16 casts for spyre below as well, instead of looping through all the individual parameters? e.g. just using an if/else here:

```
        for module_name, module in self.fms_model.named_modules():
            if module_name in mm_module_names:
                logger.debug(
                    "Placing %s submodule on device=%s dtype=%s.",
                    module_name,
                    mm_device,
                    cpu_mm_dtype,
                )
                module.to(device=mm_device, dtype=cpu_mm_dtype)
            else:
                # cast for spyre
                module.to(dtype=torch.float16)
```

sendnn_inference/envs.py
joerunde: Since we know this is faster on `nnpa` (and assuming we don't have quality regressions, but we need to check on those) should we update this to default to `nnpa` if it's available?

The goal here would be for users to not need to set `SENDNN_INFERENCE_MM_DEVICE=nnpa` to get the speedup for free, but keeping the environment variable to allow them to turn it off if there's a problem is still good flexibility to have

nikheal2: ack

sendnn_inference/model_executor/model_loader/spyre.py
nikheal2: addressing in next commit

sendnn_inference/utils.py
nikheal2: ack, done

sendnn_inference/multimodal/mm_mappings/mistral3.py
nikheal2: addressed in next commit

nikheal2:
bot:test

joerunde:
bot:test

@nikheal2 I just added you to the allowlist to trigger these, I don't think you were in it before.
(Also not running fp8 tests to save an hour+)

nikheal2:
bot:test

nikheal2:
bot:test

nikheal2:
bot:test

joerunde:
@nikheal2 if you head over to the internal CI job queue for this, you'll see that jobs don't automatically abort if you trigger new ones on new commits. Each of these will tie up a node for ~2 hours for the tests so it would be good to either

1. not trigger a bunch if you're planning on pushing more commits
2. select a smaller set of tests (like `-m "not quantized and prefix_caching"`)
3. manually abort in-flight jobs if you trigger another big run

joerunde:
Looks like tests are passing!

This LGTM as long as we can validate that the hardware swap doesn't cause quality regressions. CC @gkumbhat, have we made any progress validating quality here?

gkumbhat:
@joerunde quality assessment in progress.

gkumbhat:
@joerunde, @nikheal2  did some small quality assessment, both image+text input and text only input cases seems to be working correctly.

tests/model_loader/test_spyre.py
```diff
@@ -71,11 +88,13 @@ def test_no_mm_utils_still_casts_bf16_decoder_params(monkeypatch):
...
 
-def test_non_bf16_non_mm_param_untouched(monkeypatch):
+def test_non_bf16_non_mm_param_cast_to_fp16(monkeypatch):
+    # The whole-model fp16 cast now downcasts non-mm fp32 params too (spyre cards
+    # don't support bf16; this path is for non-quantized models only).
     _set_cpu_mm_dtype(monkeypatch, "float32")
     fms = _FakeFmsModel([("decoder.layers.0.weight", torch.float32)])
     _cast(fms, _FakeMMUtils(("vision_tower.",)))
-    assert _dtype_of(fms, "decoder.layers.0.weight") == torch.float32
+    assert _dtype_of(fms, "decoder.layers.0.weight") == torch.float16
```
gkumbhat, line 99: is this diverging from how we test other models?

sendnn_inference/model_executor/model_loader/spyre.py
```diff
@@ -268,28 +272,48 @@ def load_weights(
...
-                if param.dtype != cpu_mm_dtype:
-                    logger.debug(
-                        "Casting vision encoder param %s to %s for CPU execution.",
-                        name,
-                        cpu_mm_dtype,
-                    )
-                    param.data = param.data.to(dtype=cpu_mm_dtype)
-            elif param.dtype == torch.bfloat16:
+        if mm_device == "nnpa" and mm_module_names and not utils_spyre.ensure_nnpa_registered():
+            logger.warning(
```
gkumbhat: should this be an error, since `mm_device=nnpa` will only be set by Z and so it should error instead of failing with a warning?

nikheal2: Idea is to fallback on CPU if nnpa is not getting registered even if it is on Z.
Does it make sense or you want to change the behaviour?

gkumbhat: I think falling back would probably be not desirable behavior, since if a user wanted that, then they would simply configure mm_device=cpu.. Am worried about things working with warning with device changed..


The real question is, if things are not working on intended device, should it error.

gkumbhat (approved):
Looks good to me!

gkumbhat (changes requested):
Looks like we have tests failing. Can you please fix those.

tests/model_loader/test_spyre.py
```diff
@@ -1,26 +1,42 @@
...
-class _FakeFmsModel:
-    """Minimal stand-in exposing only named_parameters()."""
+class _FakeFmsModel(torch.nn.Module):
+    """Minimal real nn.Module stand-in built from dotted parameter paths.
 
-    def __init__(self, specs):
-        self._params = [
-            (name, torch.nn.Parameter(torch.zeros(2, 2, dtype=dtype))) for name, dtype in specs
-        ]
+    _cast_params_for_spyre now casts the whole model with Module.to(...) and
```
gkumbhat: this isn't true anymore right?

tests/model_loader/test_spyre.py
```diff
@@ -1,26 +1,42 @@
...
+        super().__init__()
+        for path, dtype in specs:
+            *mod_parts, param_name = path.split(".")
+            parent = self
+            for part in mod_parts:
+                child = parent._modules.get(part)
+                if child is None:
+                    child = torch.nn.Module()
+                    parent.add_module(part, child)
+                parent = child
```
gkumbhat, line 36: why is this needed?

tests/model_loader/test_spyre.py
nikheal2: Yes right, I'll update docstring

tests/model_loader/test_spyre.py
nikheal2: mm submodules have to be moved  to nnpa and param.data can't swap a CPU tensor for an nnpa-backend , so _cast_params_for_spyre was changed to use named_modules() + module.to(device=...).
Earlier we did not have real submodule and couldn't exercise named_modules()/.to(). so fake had to be rebuilt as a real module tree.

sendnn_inference/model_executor/model_loader/spyre.py
```diff
@@ -268,28 +272,52 @@ def load_weights(
...
-                    logger.debug(
-                        "Casting vision encoder param %s to %s for CPU execution.",
-                        name,
-                        cpu_mm_dtype,
-                    )
-                    param.data = param.data.to(dtype=cpu_mm_dtype)
-            elif param.dtype == torch.bfloat16:
+                continue
+            if param.dtype == torch.bfloat16:
+                param.data = param.data.to(dtype=torch.float16)
```
gkumbhat: wait,  don't we want to case them to `cpu_mm_dtype` ?

sendnn_inference/model_executor/model_loader/spyre.py
```diff
@@ -268,28 +272,52 @@ def load_weights(
...
+
+        # Move the multimodal submodules whole. We must use Module.to(...) here
+        # rather than mutating param.data, because nn.Parameter.set_data refuses
+        # to swap tensor types when the backend changes (e.g. CPU TensorImpl ->
+        # NnpaTensorImpl). Module._apply rebuilds Parameter objects correctly
+        # in that case. As a bonus this also moves any buffers under the
+        # submodule.
+        for module_name, module in self.fms_model.named_modules():
```
gkumbhat, line 305: why do we need to do both module level and parameter level?

gkumbhat (changes requested):

gkumbhat (approved):
Looks good to me!

nikheal2:
bot:test

PR #979, opened 2026-05-27
Put structure output grammar check in 'can_schedule_prefill'

sducouedic:
I might be missing something, but it looks to me that the check for the request structured output to be ready can (and should) be done in `can_schedule_prefill`.


sducouedic:
bot:test

PR #982, opened 2026-05-27
[WIP] Pre-allocate all blocks in spyre scheduler

maxdebayser:
## Description

To improve device utilization, we want to allow a form of request preemption where the scheduler can remove requests from the current decoding batch but without freeing the associated KV-Cache blocks, essentially pausing the execution of the request. For that we need to be able to track how much KV cache blocks are available and reserve them to prevent overcommiting resources.

I see two design choices for this:
1) Replicate block accounting in the SpyreScheduler
2) Add behavior in the SpyreScheduler to verify block availability and pre-allocate all blocks.

With option 1) we need to be careful to track how many blocks the scheduler has assigned to each request, especially considering that some might be cached. This option is a bit simpler, but the vllm scheduler has no visibility of the blocks that are reserved.

With option 2) before callig the vllm Schedule.schedule(), we verify the availabity of all the blocks required to generate the requests max-tokens and afterwards we pre-allocate these blocks. 

This PR explores option 2.

### DIFF
--- a/sendnn_inference/v1/core/scheduler.py
+++ b/sendnn_inference/v1/core/scheduler.py
@@ -5,6 +5,7 @@ from collections import deque
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Iterable, Union
 
+
 from vllm.logger import init_logger
 from vllm.v1.core.sched.scheduler import Scheduler
 from vllm.v1.metrics.stats import SchedulerStats
@@ -151,10 +152,6 @@ class BlockAllocationParams:
     num_new_tokens: int
     num_new_computed_tokens: int
     new_computed_blocks: KVCacheBlocks | None
-    num_lookahead_tokens: int
-    num_external_computed_tokens: int
-    delay_cache_blocks: bool
-    num_encoder_tokens: int
 
 
 class ChunkedPrefillSpyreScheduler(SpyreScheduler):
@@ -278,12 +275,15 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
         # This basically replicates what the scheduler already does, but
         # scattered all over the place in `schedule()`
         if request.num_computed_tokens == 0:
+            old_log_stats = self.kv_cache_manager.log_stats
+            self.kv_cache_manager.log_stats = False
             new_computed_blocks, num_new_local_computed_tokens = (
                 self.kv_cache_manager.get_computed_blocks(request)
             )
+            self.kv_cache_manager.log_stats = old_log_stats
             num_computed_tokens = num_new_local_computed_tokens
         else:
-            new_computed_blocks = self.kv_cache_manager.empty_kv_cache_blocks
+            new_computed_blocks = self.kv_cache_manager.create_kv_cache_blocks(blocks=tuple())
             num_new_local_computed_tokens = 0
             num_computed_tokens = request.num_computed_tokens
 
@@ -297,44 +297,11 @@ class ChunkedPrefillSpyreScheduler(SpyreScheduler):
 
         num_new_tokens = num_tokens - num_computed_tokens
 
-        num_encoder_tokens = 0
-        if self.is_encoder_decoder and request.has_encoder_inputs:
-            # Max: for now I think each request gets the full encoder budget
-            # since we don't prefill more than request at a time
-            encoder_compute_budget = self.max_num_encoder_input_tokens
-            token_budget = self.max_num_scheduled_tokens
-            num_new_prompt_tokens = request.num_tokens - num_computed_tokens
-            num_new_prompt_tokens = min(num_new_prompt_tokens, token_budget)
-            (
-                encoder_inputs_to_schedule,
-                updated_new_prompt_tokens,
-                new_encoder_compute_budget,
-                external_load_encoder_input,
-            ) = self._try_schedule_encoder_inputs(
-                request,
-                num_computed_tokens,
-                num_new_prompt_tokens,
-                encoder_compute_budget,
-                shift_computed_tokens=1 if self.use_eagle else 0,
-            )
-            if updated_new_prompt_tokens != num_new_prompt_tokens:
-                # TODO: improve handling of this case
-                raise ValueError("Encoder request cannot be scheduled")
-
-            if encoder_inputs_to_schedule:
-                num_encoder_tokens = sum(
-                    request.get_num_encoder_embeds(i) for i in encoder_inputs_to_schedule
-                )
-
         return BlockAllocationParams(
             original_request_computed_tokens=request.num_computed_tokens,
             num_new_tokens=num_new_tokens,
             num_new_computed_tokens=num_new_local_computed_tokens,
             new_computed_blocks=new_computed_blocks,
-            num_lookahead_tokens=0,  # we don't support spec decoding
-            num_external_computed_tokens=0,  # we don't support KV cache transfers
-            delay_cache_blocks=False,  # we don't support KV cache loading
-            num_encoder_tokens=num_encoder_tokens,
         )
 
     def _are_blocks_available(self, request: Request, block_params: BlockAllocationParams) -> bool:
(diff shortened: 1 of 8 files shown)


sendnn_inference/v1/core/scheduler.py
```diff
@@ -253,6 +269,121 @@ def adjust_computed_tokens(
...
+        """
+        # This basically replicates what the scheduler already does, but
+        # scattered all over the place in `schedule()`
+        if request.num_computed_tokens == 0:
+            new_computed_blocks, num_new_local_computed_tokens = (
+                self.kv_cache_manager.get_computed_blocks(request)
+            )
+            num_computed_tokens = num_new_local_computed_tokens
+        else:
+            new_computed_blocks = self.kv_cache_manager.empty_kv_cache_blocks
```
sducouedic: nit but apparently we shouldn't use directly `empty_kv_cache_blocks`: https://github.com/vllm-project/vllm/blob/61288b5458b9b9136cc6cfc583bd2bfb22ccd818/vllm/v1/core/kv_cache_manager.py#L164-L168, but `create_kv_cache_blocks` with empty tuple instead

sendnn_inference/v1/core/scheduler.py
```diff
@@ -253,6 +269,121 @@ def adjust_computed_tokens(
...
+            assert request.sampling_params is not None
+            assert request.sampling_params.max_tokens is not None
+            prompt_tokens = request.num_prompt_tokens
+            max_tokens = request.sampling_params.max_tokens
+            num_tokens = prompt_tokens + max_tokens
+
+        num_new_tokens = num_tokens - num_computed_tokens
+
+        num_encoder_tokens = 0
+        if self.is_encoder_decoder and request.has_encoder_inputs:
```
sducouedic: do we support encore-decoder model at all? https://github.com/torch-spyre/sendnn-inference/blob/53f5366851e3b40f86c60c02f9ed7a953af03825/sendnn_inference/platform.py#L499
I guess when we will do, then the correct `num_encoder_tokens` will already be computed

sendnn_inference/v1/core/scheduler.py
```diff
@@ -273,9 +404,19 @@ def schedule(self) -> "SchedulerOutput":
...
         while holdback_queue:
-            if self.can_schedule_prefill(holdback_queue[0]):
+            request = holdback_queue[0]
+
+            if self.can_schedule_prefill(request):
+                block_params = self._get_block_params_for_request(request)
+                block_alloc_params[request.request_id] = block_params
+                if not self._are_blocks_available(request, block_params):
+                    # can't schedule the request
+                    break
```
sducouedic: can we move this code to `can_schedule_prefill`? It is part of the check whether there are enough blocks available, so I believe it belongs to that function

sendnn_inference/v1/core/scheduler.py
```diff
@@ -253,6 +269,121 @@ def adjust_computed_tokens(
...
+
+            if encoder_inputs_to_schedule:
+                num_encoder_tokens = sum(
+                    request.get_num_encoder_embeds(i) for i in encoder_inputs_to_schedule
+                )
+
+        return BlockAllocationParams(
+            original_request_computed_tokens=request.num_computed_tokens,
+            num_new_tokens=num_new_tokens,
+            num_new_computed_tokens=num_new_local_computed_tokens,
```
sducouedic, line 303: Anything preventing us naming `num_new_local_computed_tokens` `num_new_computed_tokens`? I am not sure I understand the meaning of "local"

sendnn_inference/v1/core/scheduler.py
```diff
@@ -253,6 +269,121 @@ def adjust_computed_tokens(
...
+            delay_cache_blocks=False,  # we don't support KV cache loading
+            num_encoder_tokens=num_encoder_tokens,
+        )
+
+    def _are_blocks_available(self, request: Request, block_params: BlockAllocationParams) -> bool:
+        """
+        Checks if there are enough blocks for this request to run
+        until it's max-tokens are reached.
+        """
+        full_num_tokens = min(block_params.num_new_tokens, self.max_model_len)
```
sducouedic, line 312: nit, but I think requests with more than max_model_len will be rejected automatically? so full_num_tokens will always be `block_params.num_new_tokens`

sendnn_inference/v1/core/scheduler.py
```diff
@@ -253,6 +269,121 @@ def adjust_computed_tokens(
...
+        num_local_computed_tokens = (
+            request.num_computed_tokens + block_params.num_new_computed_tokens
+        )
+        total_computed_tokens = min(num_local_computed_tokens, self.max_model_len)
```
sducouedic, line 316: here also, I don't think we can have more `num_local_computed_tokens` than `max_model_len`? If I am not wrong, then we could directly name `num_local_computed_tokens` `total_computed_tokens` which is easier to understand

sendnn_inference/v1/core/scheduler.py
```diff
@@ -253,6 +269,121 @@ def adjust_computed_tokens(
...
+        num_blocks_to_allocate = self.kv_cache_manager.coordinator.get_num_blocks_to_allocate(
+            request_id=request.request_id,
+            num_tokens=full_num_tokens,
+            new_computed_blocks=block_params.new_computed_blocks.blocks,
+            num_encoder_tokens=block_params.num_encoder_tokens,
+            total_computed_tokens=total_computed_tokens,
+            num_tokens_main_model=full_num_tokens,
+            apply_admission_cap=True,
```
sducouedic: I am really not sure what is `apply_admission_cap`, maybe from your investigation you figured it should be set to True. But quickly reading from the docstring: https://github.com/vllm-project/vllm/blob/c1c4db8b4b1f2178b2ca1e0638ff4cf39cf11cc5/vllm/v1/core/kv_cache_coordinator.py#L105-L108
```
            apply_admission_cap: If True, apply the recycling-aware
                per-request admission cap (SWA / chunked-local). Set only by
                the full-sequence admission gate; per-step allocation must
                leave it False so the predictor matches `allocate_new_blocks`.
```
looks to me the parameter is for sliding-window-attention or chunked-local (not sure what that one is, will investigate), but in our case I guess there is no recycling-aware specs in our case? So set to `False` according to what is said here 🤔

sendnn_inference/v1/core/scheduler.py
```diff
@@ -253,6 +269,121 @@ def adjust_computed_tokens(
...
+        )
+        return num_blocks_to_allocate <= self.kv_cache_manager.block_pool.get_num_free_blocks()
+
+    def _preallocate_blocks(
+        self, request: Request, block_params: BlockAllocationParams, outputs: SchedulerOutput
+    ):
+        # here we need to temporarily undo the effect of _update_after_schedule
```
sducouedic, line 330: general question: since `_preallocate_blocks` is called after the parent scheduler, doesn't the parent scheduler allocate already necessary blocks for the next step? should we account for this?

sducouedic (commented):
I left a couple of comments and questions. Great investigation work and putting everything together, this block allocation part is so tricky

sendnn_inference/v1/core/scheduler.py
maxdebayser: Actually, I'm going to remove all of this num_encoder_tokens accounting because I realized that it's not necessary. The input sequence already contains the placeholder tokens so the encoder tokens aren't added to the input sequence I believe. Also, the num_encoder_tokens is just passed to the `CrossAttentionManager` which doesn't do anything with it.

sendnn_inference/v1/core/scheduler.py
maxdebayser: It's tokens that aren't coming from kv cache transfer. The idea is to follow the same naming used in the upstream scheduler.

maxdebayser:
Closed in favor of https://github.com/torch-spyre/sendnn-inference/pull/984

Issue #947, opened 2026-04-22
Split sampling from model execution


joerunde:
Following up from both structured output and async scheduling, we need to move the sampling out of execute_model and into sample_tokens. This will allow the vllm engine to generate grammar bitmasks while the model is running which will boost performance.

From https://github.com/torch-spyre/sendnn-inference/pull/903#discussion_r3120759211

-------

It looks to me like this is here because we combine the token sampling in `model_executor.execute_model`, and if we were to implement `sample_tokens` in the model runner then we wouldn't have to do this. 

The engine core runs this:
```
        scheduler_output = self.scheduler.schedule()
        future = self.model_executor.execute_model(scheduler_output, non_block=True)
        grammar_output = self.scheduler.get_grammar_bitmask(scheduler_output)

        model_output = future.result()
        if model_output is None:
            model_output = self.model_executor.sample_tokens(grammar_output)
```

The important thing that I see here is that this collects the grammar bitmask asynchronously while the model is running, which seems like something that we should be doing. 

Assuming the above is correct, then:
1. At a minimum we need to put some comments here explaining this is what we're doing
2. I have a heavy preference for actually implementing `sample_tokens()` so that we can take advantage of the performance benefit

Given the time constraints of releasing this feature for `sendnn_inference==2.0.0`, we could just add some comments here and open an issue to follow up

nikheal2:
Merged!

PR #985, opened 2026-06-01
Split grammar computation from execute_model()

dilipgb:
## Description

This PR separates sampling code from Model runner and moves it under worker. This is expected to improve some performance when structured output is used along with sampling.

## Related Issues

#947 

## Test Plan

Ran e2e tests using granite-3.3-instruct-1b micro model. Tests passed. Also ran benchmarks using `vllm bench`. structured output benchmarking still pending on the issue [1828](https://github.ibm.com/ai-foundation/aiu-app-sw-tracker/issues/1828)

## Checklist


sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1493,9 +1493,15 @@ def apply_grammar_bitmask(
     ) -> None:
         """Apply grammar bitmask in-place to constrain logits for structured
         output requests.
+
+        This method checks if grammar output has been pre-computed and attached
+        to the scheduler_output. If not present, it will be computed synchronously
+        for backward compatibility.
```
joerunde: Can we push for simplicity here and not keep backwards compatibility?

tests/e2e/test_chunked_prefill_tkv_steps.py
```diff
@@ -67,7 +67,12 @@ def get_cpu_model_runner(
     max_num_seqs: int,
     max_num_batched_tokens: int,
     monkeypatch: pytest.MonkeyPatch,
-) -> ChunkedPrefillModelRunner:
```
joerunde, line 70: let's not go backwards with the type hinting- if we need a tuple here then it needs to be type hinted

tests/e2e/test_chunked_prefill_tkv_steps.py
```diff
@@ -80,7 +85,8 @@ def get_cpu_model_runner(
     )
 
     # NB: Only works because this engine is run with no multiprocessing and TP=1
-    runner: ChunkedPrefillModelRunner = engine_core.model_executor.driver_worker.worker.model_runner
+    worker = engine_core.model_executor.driver_worker.worker
+    runner: ChunkedPrefillModelRunner = worker.model_runner
```
joerunde: In the tests that are touched here I don't see the runner used any more, just the worker. Let's clean this up to only return the worker and update the name to `get_cpu_worker` to match.

If any tests do still need the model runner, they can use `worker.model_runner` to access it

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1561,49 +1592,9 @@ def execute_model(
...
-
-        # Only return outputs from the driver worker
-        if not self.is_driver_worker:
-            return self.get_empty_output()
-
-        model_output = self.sampled_output(output, is_prefill)
-        return model_output
+        # Return logits and metadata to worker for sampling
+        # Worker will call sample_tokens() with this data
+        return (logits, is_prefill, model_input, t0)
```
joerunde: This isn't a valid return type for `execute_model` though- that should be: `ModelRunnerOutput | None | Future[ModelRunnerOutput | None]`

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -770,20 +770,121 @@ def kv_cache(self) -> list[list[torch.Tensor]] | None:
...
+
+        # Apply grammar bitmask for structured output requests
+        self.model_runner.apply_grammar_bitmask(  # type: ignore[attr-defined]
+            scheduler_output,
+            logits,
+            self.model_runner.prefill_batch if is_prefill else self.model_runner.input_batch,  # type: ignore[attr-defined]
+        )
+
+        # Sample the next token
+        output = self.model_runner.model.sample(
```
joerunde: see the abstraction leaks here: this shouldn't know about the internal details of the model runner

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -770,20 +770,121 @@ def kv_cache(self) -> list[list[torch.Tensor]] | None:
...
+        )
+        sampled_ids = sampler_output.sampled_token_ids.tolist()
 
-        return EMPTY_MODEL_RUNNER_OUTPUT
+        for i, req_id in enumerate(req_ids):
+            req_state = self.model_runner.requests[req_id]  # type: ignore[attr-defined]
+            req_state.append_output_token_ids(sampled_ids[i])
 
     @SpyrePlatform.inference_mode()
+    def sample_tokens(
```
joerunde: The worker should already delegate `sample_tokens` to the model runner, which we have yet to implement. The intent with this work was to do that implementation in the model runner

As is, this creates a new signature for `sample_tokens` which collides with the existing interface: https://github.com/vllm-project/vllm/blob/v0.22.0/vllm/v1/executor/abstract.py#L229-L247. Confusing collisions like this shouldn't be merged

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -770,20 +770,121 @@ def kv_cache(self) -> list[list[torch.Tensor]] | None:
...
+        return self.model_runner.sampled_output(output, is_prefill)  # type: ignore[attr-defined]
+
     def execute_model(
         self,
         scheduler_output: "SchedulerOutput",
     ) -> ModelRunnerOutput | None:
         if self.profiler is not None:
             self.profiler.step()
-        output = self.model_runner.execute_model(scheduler_output)
-        return output if self.is_driver_worker else None
```
joerunde, line 804: I don't think this method definition should be changed as part of this PR. 

The goal here is to let the engine build the grammar bitmasks asynchronously while the model runs (see [here](https://github.com/vllm-project/vllm/blob/v0.22.0/vllm/v1/engine/core.py#L440-L441)) and then call `sample_tokens` after both of those tasks finish (see [here](https://github.com/vllm-project/vllm/blob/v0.22.0/vllm/v1/engine/core.py#L446-L448))

vLLM already takes care of this for us as long as we return `None` from the model runner. All we should have to do is make sure the state in the model runner is updated in `execute_model` such that we know how to sample the correct tokens in `sample_tokens` when that is called later with the grammar bitmask

joerunde (changes requested):
As is this diverges the interfaces from vllm which I strongly think we need to avoid.

The strategy here should be to allow the engine to build the grammar bitmasks async instead of precomputing them in the scheduler. I think the definition of done here is that we must remove this from the scheduler:
```
        outputs._spyre_grammar_output = self.get_grammar_bitmask(outputs)  # type: ignore[attr-defined]
```

Most of this code should be reusable, just organized a bit differently.

maxdebayser (changes requested):
As @joerunde mentioned, in the upstream gpu_worker, the sampling request is forwarded to the model runner, which is the best place to implement it due to the internal data structures that have to be read.

I think this PR should be named "Split sample_tokens from execute_model"

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1569,6 +1575,18 @@ def execute_model(
             logger.debug("t_forward_pass: %.2fms [prefill single chunk][batch size 1]", (t1 * 1000))
             return self.prefill_output()
 
+        # Check if we need to wait for grammar bitmask to be built asynchronously
+        # If grammar output is not ready yet, store state and return None
+        grammar_output = getattr(scheduler_output, "_spyre_grammar_output", None)
```
maxdebayser: Wouldn't it be easier to always store the state and return None? This would also eliminate the duplicated code in execute_model.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -701,6 +701,12 @@ def __init__(
 
         self._enable_prefix_caching = vllm_config.cache_config.enable_prefix_caching
 
+        # State for async sampling (when grammar bitmask is being built)
```
maxdebayser, line 766: I would prefer a new data structure here, like
```
@dataclass
class SamplingSate:
    logits
    metadata
    is_prefill
    scheduler_output
```

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1569,6 +1575,18 @@ def execute_model(
...
 
+        # Check if we need to wait for grammar bitmask to be built asynchronously
+        # If grammar output is not ready yet, store state and return None
+        grammar_output = getattr(scheduler_output, "_spyre_grammar_output", None)
+        if grammar_output is None and hasattr(scheduler_output, "grammar_req_ids"):
+            # Grammar is being built asynchronously, store state for later sampling
+            self._pending_sampling_logits = logits
+            self._pending_sampling_metadata = self.get_sampling_metadata(is_prefill)
+            self._pending_is_prefill = is_prefill
+            self._pending_scheduler_output = scheduler_output
```
maxdebayser: To err on the side of caution, I think it's better to store a deep copy of the scheduler output.

maxdebayser (changes requested):

sendnn_inference/v1/worker/spyre_model_runner.py
joerunde: +1, my understanding here is that we need to remove `scheduler_output._spyre_grammar_output`. 

That's currently set in the scheduler by having it pre-build the grammars, but the whole idea is to let the engine do that asynchronously while the model runs

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1632,6 +1650,70 @@ def prefill_output(self) -> SpyreModelRunnerOutput:
             prefix_cache_hit_len=self.get_prefix_cache_len(),
         )
 
+    def sample_tokens(self, scheduler_output: "SchedulerOutput") -> ModelRunnerOutput:
```
joerunde: Shouldn't this signature accept the grammar, and not the scheduler output?

```
    def sample_tokens(self, grammar_output: "GrammarOutput | None") -> ModelRunnerOutput:
```

we need to use the grammar output that the engine generated async and not the current `scheduler_output._spyre_grammar_output` hack

dilipgb:
Granite-3.3-8b-instruct

Without camping

```
============ Serving Benchmark Result ============
Successful requests:                     1000      
Request rate configured (RPS):           32.00     
Benchmark duration (s):                  594.77    
Total input tokens:                      121000    
Total generated tokens:                  94000     
Request throughput (req/s):              1.68      
Output token throughput (tok/s):         158.04    
Total token throughput (tok/s):          361.48    
---------------Time to First Token----------------
Mean TTFT (ms):                          276752.91 
Median TTFT (ms):                        277486.54 
P99 TTFT (ms):                           546248.91 
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          198.40    
Median TPOT (ms):                        198.68    
P99 TPOT (ms):                           212.28    
---------------Inter-token Latency----------------
Mean ITL (ms):                           196.29    
Median ITL (ms):                         106.44    
P99 ITL (ms):                            408.82    
==================================================
correct_rate(%) 100.0 
```

With sampling
```
============ Serving Benchmark Result ============
Successful requests:                     1000      
Request rate configured (RPS):           32.00     
Benchmark duration (s):                  592.07    
Total input tokens:                      121000    
Total generated tokens:                  94000     
Request throughput (req/s):              1.69      
Output token throughput (tok/s):         158.77    
Total token throughput (tok/s):          363.13    
---------------Time to First Token----------------
Mean TTFT (ms):                          274694.84 
Median TTFT (ms):                        274608.77 
P99 TTFT (ms):                           542728.94 
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          197.38    
Median TPOT (ms):                        199.02    
P99 TPOT (ms):                           208.48    
---------------Inter-token Latency----------------
Mean ITL (ms):                           195.28    
Median ITL (ms):                         105.42    
P99 ITL (ms):                            401.46    
==================================================
correct_rate(%) 100.0 
```

dilipgb:
I'll share benchmark results without this changes soon.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1462,10 +1478,24 @@ def _update_batch(self, scheduler_output: SchedulerOutput):
         if scheduler_output.finished_req_ids:
             for req_id in scheduler_output.finished_req_ids:
                 self.input_batch.remove_request(req_id)
-                # TODO: Processing multiple removals at once can break alignment
-                # of logitprocs. Refactor so that we can batch removals to the
-                # `input_batch`
+                # Clean up request state to prevent memory leak
+                # Without this, SamplingRequestState (including cached_mm_embeddings,
+                # generators, block_ids) leaks for every completed request
+                self.requests.pop(req_id, None)
```
maxdebayser, line 1743: Yeah, I also found this working on another branch. I think the comment is not necessary though.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1462,10 +1478,24 @@ def _update_batch(self, scheduler_output: SchedulerOutput):
...
+
                 self.input_batch.refresh_metadata()
+
+            # If pending sampling exists and requests were finished/aborted,
+            # the pending state may now be invalid (batch mismatch, KeyError on req_id)
+            # Clear it to prevent crashes when sample_tokens() eventually arrives
+            if self._pending_sampling_state is not None:
```
maxdebayser: The code is clear already, I think the comment is unecessary

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1488,43 +1518,217 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+                "Engine must guarantee at most one pending grammar batch globally."
+            )
+
+        # Deep copy metadata to prevent mutation issues
+        # SamplingMetadata likely contains mutable tensors (selected_token_indices,
+        # generators, sampling_tensors, temperature arrays) that are updated by
+        # refresh_metadata() on every scheduler step. If we store by reference,
+        # when sample_tokens() executes, it will use metadata from a different
+        # batch, causing wrong sampling (wrong request, wrong temperature, wrong RNG).
+        try:
```
maxdebayser: I don't think we need a try.. except here. It will either always fail or always succeed. If it fails, a more fine grained approach is needed. Also, we don't need so many comments here.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1488,43 +1518,217 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+            )
+            # Fall back to reference (risky but allows execution to continue)
+            metadata = self.get_sampling_metadata(is_prefill)
+
+        # Deep copy scheduler_output to prevent mutation issues
+        # Many schedulers recycle objects for performance. If scheduler mutates
+        # this object later, stored_scheduler_output may no longer represent the
+        # batch that produced the logits. This can cause: wrong req_ids, wrong
+        # grammar mapping, wrong logprob routing.
+        try:
```
maxdebayser: Same as above

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1488,43 +1518,217 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+        - SamplingMetadata references
+        - SchedulerOutput references
+        """
+        if self._pending_sampling_state is not None:
+            logger.warning(
+                "Clearing pending sampling state without sample_tokens() being called. "
+                "This may indicate a request cancellation, grammar build failure, or shutdown. "
+                "Leaked state: logits (GPU memory), metadata, scheduler_output."
+            )
+        self._pending_sampling_state = None
```
maxdebayser: Isn't this a bit too much just for setting `self._pending_sampling_state = None`?

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1488,43 +1518,217 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+        logits: torch.Tensor,
+        is_prefill: bool,
+        scheduler_output: "SchedulerOutput",
+    ) -> None:
+        """Defer sampling until grammar bitmask is ready.
+
+        Stores the current sampling context for later completion.
+        """
+        # Protect against double defer (batch A overwriting batch B)
+        if self._pending_sampling_state is not None:
```
maxdebayser: `assert` is shorter than `if ... raise RuntimeError()`

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1488,43 +1518,217 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+        req_ids = (
+            [r.req_id for r in scheduler_output.scheduled_new_reqs]
+            if len(scheduler_output.scheduled_new_reqs) > 0
+            else batch.sorted_requests_ids
         )
+        sampled_ids = output.sampled_token_ids.tolist()
+
+        # Verify batch/request alignment to catch silent corruption early
+        # Use explicit check instead of assert to ensure it's not removed by python -O
+        if len(req_ids) != len(sampled_ids):
```
maxdebayser: `assert` is shorter than `if ... raise RuntimeError()`

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1488,43 +1518,217 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+            logger.warning(
+                "Clearing pending sampling state without sample_tokens() being called. "
+                "This may indicate a request cancellation, grammar build failure, or shutdown. "
+                "Leaked state: logits (GPU memory), metadata, scheduler_output."
+            )
+        self._pending_sampling_state = None
 
-            def __getattr__(self, name: str):
-                return getattr(self._batch, name)
+    def apply_constraints(
```
maxdebayser: `apply_constraints` is always called right before `perform_sampling`. Maybe it would be simpler to add its code into that function.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1558,60 +1760,35 @@ def execute_model(
...
-        if not self.is_driver_worker:
-            return self.get_empty_output()
+        # Apply constraints
+        self.apply_constraints(scheduler_output, grammar_output, logits, is_prefill)
 
-        model_output = self.sampled_output(output, is_prefill)
-        return model_output
+        # Perform sampling and build output
+        sampling_metadata = self.get_sampling_metadata(is_prefill)
+        return self.perform_sampling(logits, sampling_metadata, is_prefill, scheduler_output, t0)
```
maxdebayser: I think you can always return none here and let the engine call sampling afterwards.

maxdebayser (changes requested):

sendnn_inference/v1/core/scheduler.py
```diff
@@ -212,24 +211,29 @@ def __init__(self, *args, **kwargs) -> None:
...
-        assert isinstance(model_runner_output, SpyreModelRunnerOutput), (
-            "Expecting an instance of CPSpyreModelRunnerOutput when doing chunked prefill."
-        )
-
-        # Remove completed prefills
-        self.ongoing_prefills = [
-            req for req in self.ongoing_prefills if req.num_computed_tokens < req.num_prompt_tokens
-        ]
+        # model_runner_output can be None when sampling is deferred
+        if model_runner_output is not None:
```
maxdebayser: If super().update_from_output is never called with None, why would it be called here in the derived class?

sendnn_inference/v1/core/scheduler.py
```diff
@@ -394,33 +398,13 @@ def schedule(self) -> "SchedulerOutput":
...
-                    so_req = r.structured_output_request
-                    if so_req and so_req.grammar:
-                        r.status = RequestStatus.WAITING
-
-            ready_to_prefill = [
-                r
-                for r in self.waiting
-                if r.status != RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR
-            ]
+            ready_to_prefill = [list(self.waiting)]
```
maxdebayser: Now you can just do
```
if self.waiting:
```

tests/v1/worker/mock_model.py
```diff
@@ -154,6 +154,8 @@ def execute_model(
     def execute_new_request(self, request: Request) -> ModelRunnerOutput:
         scheduler_output = self._schedule_new_request(request)
         output = self.execute_model(scheduler_output)
+        if output is None:
+            output = self.sample_tokens(grammar_output=None)
```
joerunde: Can we promote `execute_and_sample` to a test utils package that we can import and use everywhere?

tests/e2e/test_chunked_prefill_tkv_steps.py
```diff
@@ -94,6 +94,23 @@ def get_cpu_model_runner(
...
+    """Run execute_model and, if sampling was deferred, call sample_tokens.
+
+    execute_model() now always defers sampling and returns None; the caller
+    must follow up with sample_tokens().  For incomplete prefill chunks no
+    sampling is needed and execute_model returns a real (empty) output
+    directly, so we handle both cases here.
+    """
+    output = runner.execute_model(scheduler_output)
+    if output is None:
+        output = runner.sample_tokens(grammar_output=None)
```
joerunde: can we assert here that none of the requests asked for structured output?

If this is re-used for cases that do, then we'd need to generate the grammar here too

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -811,6 +817,12 @@ def execute_model(
         output = self.model_runner.execute_model(scheduler_output)
         return output if self.is_driver_worker else None
 
+    def sample_tokens(
+        self,
+        grammar_output: "GrammarOutput | None",
+    ):
```
joerunde: `-> ModelRunnerOutput:`?

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -522,6 +523,9 @@ def _warmup_spyre_dynamic_size(self, special_token_ids) -> float:
         logger.info("[WARMUP] Deploying prefill to device...")
         self.execute_model(scheduler_output)
 
+        if hasattr(self.model_runner, "clear_pending_sampling"):
```
joerunde: why all of the `hasattr`s for `clear_pending_sampling`?

For dynamic warmup, we should be able to assert that this is a `ChunkedPrefillModelRunner`, but better yet would be to add this to the base spyre model runner api and implement a no-op for the pooling runner (because it doesn't sample)

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1726,60 +1949,25 @@ def execute_model(
...
             )
 
-        # If the prompt is being prefilled we don't have to sample
-        # and generate a new token.
+        # Early return for incomplete prefill (no sampling needed)
         if is_prefill and self.check_incomplete_prefill(scheduler_output):
-            # Only return outputs from the driver worker
+            # Clear any pending sampling state since we're not sampling this iteration
+            if self._pending_sampling_state is not None:
+                self.clear_pending_sampling(reason="incomplete_prefill")
```
joerunde: Looking at this error from `sample_tokens`:
```
        assert self._pending_sampling_state is not None, (
            "sample_tokens() called but no pending sampling state exists. "
            "This indicates sample_tokens() was called without prior defer_sampling(), "
            "or the state was already cleared."
        )
```

it's not clear to me why this is needed if we haven't called `self.defer_sampling` yet in this code path

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -113,6 +115,39 @@ class SpyreModelRunnerOutput(ModelRunnerOutput):
...
+    scheduler_output: "SchedulerOutput"
+    batch_req_ids: list[str]  # Store request IDs to validate batch consistency
+
+
+class SchedulerOrderedBatchAdapter:
+    """Adapter to expose batch with request ordering matching the scheduler.
+
+    This adapter wraps a SamplingInputBatch and overrides the request ordering
+    to match the scheduler's order, which is required for grammar bitmask application.
+    The adapter also exposes the actual number of requests (not max capacity).
```
joerunde: > The adapter also exposes the actual number of requests (not max capacity).

Is that a bug in `SamplingInputBatch` or is that intended behavior?

tests/v1/worker/test_grammar_bitmask_llm.py
```diff
@@ -0,0 +1,167 @@
+# SPDX-License-Identifier: Apache-2.0
+"""End-to-end vllm.LLM test for grammar bitmask routing.
```
joerunde, line 2: @dilipgb, claude was super sus about the batch handling with `SchedulerOrderedBatchAdapter`, so we wrote up this test that uses a large batch with deterministic outputs to check if the grammars are applied correctly. 

It's currently failing on my machine, so it looks like we have correctness issues to address.

update: This fails on CI for the PR here as well. I tested against main and it passes, so this looks like a regression introduced by this PR.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+
+    def clear_pending_sampling(
+        self,
+        reason: str = "unknown",
+        finished_req_ids: list[str] | None = None,
+    ) -> None:
+        """Clear any pending sampling state.
+
+        This should be called on error paths or when aborting deferred sampling
+        to prevent memory leaks.
```
joerunde: I don't totally understand how setting `self._pending_sampling_state = None` fixes memory leaks,

Isn't this functionally the same as assigning `self._pending_sampling_state = SamplingState(...) on the next pass? The garbage collector will clean up either way

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+
+        Args:
+            reason: Why the pending state is being cleared (for debugging).
+            finished_req_ids: Request IDs that finished, if applicable.
+        """
+        if self._pending_sampling_state is not None:
+            stored_batch_req_ids = self._pending_sampling_state.batch_req_ids
+            logger.warning(
```
joerunde: this creates a ton of warnings during normal warmup which we shouldn't do

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+        # so we capture the request IDs at this point in time.
+        batch = self.prefill_batch if is_prefill else self.input_batch
+        batch_req_ids = list(
+            batch.sorted_requests_ids if hasattr(batch, "sorted_requests_ids") else batch.req_ids
+        )
+
+        self._pending_sampling_state = SamplingState(
+            # logits.clone() is needed: the underlying tensor may be a view into
+            # model output buffers that get overwritten on the next forward pass.
+            logits=logits.clone(),
```
joerunde: Is that true though? This clone seems expensive, since logits would be shape `batch x vocab`. Do tests actually fail without it?

IIUC, when we compile the fms models on spyre they allocate a new output buffer for logits for every forward pass, but I could be misunderstanding that.

sendnn_inference/v1/sample/spyre_logits_processor.py
```diff
@@ -112,5 +112,5 @@ def apply(self, logits: torch.Tensor) -> torch.Tensor:
 
         return logits
 
-    def set_prefill_index(self, idx: int) -> None:
+    def set_prefill_index(self, idx: int | None) -> None:
```
maxdebayser, line 193: Wait, why is this needed? `set_prefill_index` is called only with the result of `input_batch.add_request(request)` and that is never None.

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -44,9 +44,10 @@
     SpyrePoolingModelRunner,
     SupportedTask,
 )
+from vllm.v1.core.sched.output import GrammarOutput
 
 if TYPE_CHECKING:
```
maxdebayser: You can just remove `if TYPE_CHECKING`

sendnn_inference/v1/worker/spyre_model_runner.py
maxdebayser: This seems to be a refactoring of https://github.com/torch-spyre/sendnn-inference/pull/995. But it's not clear to me why it had to be changed to pass  `scheduler_req_ids`.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1475,6 +1524,8 @@ def add_new_request(self, request: NewRequestData):
         else:
             generator = None
 
+        structured_output_request = getattr(request, "structured_output_request", None)
```
maxdebayser: Where is the corresponding `setattr`?

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1336,7 +1379,13 @@ def _prepare_decode(
             slot_mapping.append(slot)
 
             # input token and position of the token generated in the last step
-            generation_token = req_state.output_token_ids[-1]
+            # During warmup or first decode step, output_token_ids may be empty,
```
maxdebayser: What change in this PR made this necessary?

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
-        if grammar_output is None:
-            return
 
-        class _DenseBatchAdapter:
-            def __init__(self, batch: SamplingInputBatch):
-                self._batch = batch
-                self.req_ids = batch.sorted_requests_ids
+        expected_reqs = list(scheduler_output.num_scheduled_tokens.keys())
+        actual_reqs = list(
+            batch.sorted_requests_ids if hasattr(batch, "sorted_requests_ids") else batch.req_ids
```
maxdebayser: No need for `hasattr`. `sorted_requests_ids` is a property of BaseInputBatch.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
     def apply_grammar_bitmask(
```
maxdebayser, line 1770: Why are the changes to `apply_grammar_bitmask` required? (except for passing grammar_output as an argument).

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+            logger.warning(
+                "Clearing pending sampling state without sample_tokens() being called. "
+                "Reason: %s. Finished requests: %s. Stored batch requests: %s. "
+                "This may indicate a request cancellation, grammar build failure, or shutdown. "
+                "Leaked state: logits (GPU memory), metadata, scheduler_output.",
+                reason,
+                finished_req_ids,
+                stored_batch_req_ids,
+            )
+        self._pending_sampling_state = None
```
maxdebayser: 25 lines of code just for `self._pending_sampling_state = None

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+        Returns:
+            The model runner output with sampled tokens, or None for non-driver workers.
+        """
+        output: SamplerOutput | None = self.model.sample(
+            logits=logits,
+            sampling_metadata=sampling_metadata,
+        )
+        assert output is not None, "Expected sampler output"
+
+        if t0 > 0:
```
maxdebayser: At the only place where this function is called, t0 is 0, so this whole if is dead code.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1487,6 +1538,7 @@ def add_new_request(self, request: NewRequestData):
             usable_blocks=chunk_plan.usable_cache_blocks,
             total_hit_blocks=chunk_plan.total_cache_blocks,
             block_ids=request.block_ids[0],  # we only support on kv cache group for now
+            structured_output_request=structured_output_request,
```
maxdebayser: `SamplingRequestState.structured_output_request` is set here, but is never read anywhere.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+            batch_size = len(scheduler_output.num_scheduled_tokens)
+            step_type = "[prefill last chunk]" if is_prefill else "[decode]"
+            logger.debug("t_token: %.2fms %s[batch size %d]", (t1 * 1000), step_type, batch_size)
+
+        batch = self.prefill_batch if is_prefill else self.input_batch
+        req_ids = (
+            [r.req_id for r in scheduler_output.scheduled_new_reqs]
+            if len(scheduler_output.scheduled_new_reqs) > 0
+            else batch.sorted_requests_ids
+        )
```
maxdebayser: `defer_sampling` already computed this list and assigned it to `SamplingState.batch_req_ids`

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1656,54 +1709,224 @@ def maybe_setup_new_prefill(self, scheduler_output: SchedulerOutput):
...
+        sampled_ids = output.sampled_token_ids.tolist()
+
+        assert len(req_ids) == len(sampled_ids), (
+            f"Mismatch between request IDs ({len(req_ids)}) "
+            f"and sampled tokens ({len(sampled_ids)}). "
+            f"This indicates a bug in batch construction or sampling."
+        )
+
+        for i, req_id in enumerate(req_ids):
```
maxdebayser: you can use `zip(req_ids, sampled_ids, strict=True)` to iterate and at the same time enforce that the lists have the same length.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1800,6 +1988,74 @@ def prefill_output(self) -> SpyreModelRunnerOutput:
...
+        # Validate batch consistency
+        # Compare exact lists (not sets) to catch ordering issues
+        if stored_batch_req_ids != current_batch_req_ids:
+            stored_scheduler_req_ids = list(stored_scheduler_output.num_scheduled_tokens.keys())
+            raise RuntimeError(
+                f"Batch mismatch in deferred sampling. "
+                f"Stored batch requests: {stored_batch_req_ids}, "
+                f"Current batch requests: {current_batch_req_ids}, "
+                f"Stored scheduler requests: {stored_scheduler_req_ids}"
+            )
```
maxdebayser: Just trust the stored one. If we attempt async scheduling again these won't be the same.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1800,6 +1988,74 @@ def prefill_output(self) -> SpyreModelRunnerOutput:
...
+
+        # Apply grammar bitmask constraints to logits
+        self.apply_grammar_bitmask(stored_scheduler_output, grammar_output, logits, current_batch)
+
+        # Perform sampling and build output
+        result = self.perform_sampling(
+            logits, sampling_metadata, is_prefill, stored_scheduler_output, t0=0
+        )
```
maxdebayser: The actual code of `perform_sampling` is really small once you remove the surplus lines. There's no need for it to be another function.

sendnn_inference/v1/worker/spyre_model_runner.py
dilipgb: During warmup the prefill's sample_tokens() is skipped (clear_pending_sampling discards it), so output_token_ids is never populated. Idea is to Fall back to the token supplied by the warmup harness via cached_request_data.new_token_ids.

sendnn_inference/v1/worker/spyre_model_runner.py
dilipgb: original idea was to use NewRequestData object to load the structured_output_request but later decided to directly load GrammarOutput from upstream. So this may not necessary now.

sendnn_inference/v1/worker/spyre_model_runner.py
dilipgb: same as above, I need to clean up this code path.

sendnn_inference/v1/worker/spyre_model_runner.py
dilipgb: I think, this is due to FSM failure I had earlier due to mixing up with InputBatch changes. After refactoring I figured out that is redundant, hence removing other logic. Keeping it similar to previous version.

sendnn_inference/v1/sample/spyre_logits_processor.py
maxdebayser: I see now why this is needed. I think it make sense and it's better than setting the prefill index directly from the model runner.

maxdebayser:
Thanks @dilipgb , this is looking much better now. I'll give it a spin tomorrow to see it running.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1800,6 +1840,64 @@ def prefill_output(self) -> SpyreModelRunnerOutput:
             prefix_cache_hit_len=self.get_prefix_cache_len(),
         )
 
+    def sample_tokens(self, grammar_output: "GrammarOutput") -> ModelRunnerOutput | None:
```
maxdebayser: grammar_output has to be `"GrammarOutput" | None`

tests/v1/core/test_scheduler_structured_outputs.py
```diff
@@ -434,14 +434,23 @@ def test_sparse_index_grammar_crash(
         while not structured.is_grammar_ready:
             pass
 
+    def _run(sched_output):
+        """Execute one step, passing the real grammar output from the scheduler."""
+        output = pc_model_runner.execute_model(sched_output)
+        if output is None:
+            grammar_output = getattr(sched_output, "_spyre_grammar_output", None)
```
maxdebayser: `_spyre_grammar_output` is no longer being set I think.

maxdebayser: This test is passing because sample_tokens can deal with a `grammar_output` that is `None`. Wouldn't it be better to change `execute_new_request` and `execute_running_requests` to get the actual grammar output with : `self.scheduler.get_grammar_bitmask(scheduler_output)` instead of adding this new `_run` functio?

sendnn_inference/v1/core/scheduler.py
```diff
@@ -394,33 +393,12 @@ def schedule(self) -> "SchedulerOutput":
 
         # Check new requests to prefill
         elif len(self.waiting) > 0:
-            # Try to promote grammar-waiting requests whose FSM is now
```
maxdebayser, line 464: If I'm not mistaken @joerunde said that this code is still required to prevent the scheduling of requests that are still waiting for the grammar to be ready.

dilipgb: yes, I see upstream code also has wait logic in scheduler. This block should not be removed

joerunde: Yup, there are two operations for the grammars: compiling and filling. This block should have already been holding back requests while upstream vllm compiles their grammars, which we need to keep doing. The rest of the changes in the PR make the grammar _filling_ asynchronous by moving it out of the scheduler and into the engine where the grammar fill is run concurrently with the forward pass.

Before merging this PR we should have at least one test that ensures the grammar compilation is done in a way that does not block requests from scheduling. Here is a claude-drafted test- feel free to use as is or for inspiration:

<details>
<summary>Grammar compilation test</summary>

```python
# SPDX-License-Identifier: Apache-2.0
"""Regression test: a grammar-pending request must not stall in-flight decodes.

vLLM's scheduler ships structured-output requests through a separate queue
(``self.skipped_waiting``) precisely so they can sit there with their FSM
still compiling without holding up the main waiting queue. The Spyre
``ChunkedPrefillSpyreScheduler`` drains both ``waiting`` and
``skipped_waiting`` into its own holdback each iteration to enforce its
one-prefill-at-a-time invariant — so it now owns the responsibility of
treating grammar-pending requests correctly.

If the Spyre scheduler does NOT filter grammar-pending requests after the
holdback pass, then whenever such a request sits in the queue the scheduler
hides the running decodes to switch to prefill mode but ``super().schedule()``
can't actually prefill the request — its FSM isn't ready. The result is
alternating "stalled" and "decode" steps that halve throughput for every
other request behind the structured one.

This test verifies that the unstructured requests in front of a
grammar-pending structured request continue to decode on every scheduler
step until they finish, and that once the grammar is signalled ready the
structured request can then start prefilling and producing tokens.

The test only interacts with the runner's public methods
(``execute_new_request`` / ``execute_running_requests``) and reads per-step
progress via the returned ``ModelRunnerOutput``. The grammar's readiness is
gated on a ``threading.Event`` so the test can deterministically drive both
phases — pending grammar, then signalled.
"""

from __future__ import annotations

import threading
from collections import defaultdict
from concurrent.futures import Future

import pytest
from vllm.sampling_params import StructuredOutputsParams
from vllm.tokenizers import get_tokenizer
from vllm.v1.outputs import ModelRunnerOutput
from vllm.v1.request import RequestStatus
from vllm.v1.structured_output.request import StructuredOutputRequest

from scheduling_utils import create_request_for_scheduler_test, random_prompt
from spyre_util import REFERENCE_MODELS
from v1.worker.mock_model import InstrumentedModelRunner

pytestmark = [
    pytest.mark.cpu,
    pytest.mark.worker,
    pytest.mark.skip_global_cleanup,
]


class _PermissiveStubGrammar:
    """Minimal ``StructuredOutputGrammar`` stub: accepts any tokens, never
    terminates. The scheduler advances the FSM by calling ``accept_tokens``
    each step a structured request samples; this stub keeps the scheduler
    path well-formed without our having to compile a real grammar."""

    def accept_tokens(self, request_id: str, tokens) -> bool:
        return True

    def rollback(self, num_tokens: int) -> None:
        return None

    def fill_bitmask(self, bitmask, batch_index: int) -> None:
        return None

    def is_terminated(self) -> bool:
        return False

    def reset(self) -> None:
        return None


def _event_gated_grammar_future(ready: threading.Event) -> Future:
    """Build a ``Future`` that resolves to a ``_PermissiveStubGrammar`` once
    ``ready`` is set. While the event is unset, ``Future.result(timeout=...)``
    raises ``TimeoutError`` — which is what ``StructuredOutputRequest.
    is_grammar_ready`` observes when it polls the future, so the scheduler
    sees the request as "grammar still building."""
    fut: Future = Future()
    fut.set_running_or_notify_cancel()

    def waiter() -> None:
        ready.wait()
        if not fut.done():
            fut.set_result(_PermissiveStubGrammar())

    threading.Thread(target=waiter, daemon=True).start()
    return fut


def _attach_pending_grammar(request, runner, tokenizer, ready_event):
    """Equip a Request with a structured-output request whose grammar Future
    won't resolve until ``ready_event`` is set.

    ``grammar_init`` is the public API that the engine input thread would
    normally call to kick off grammar compilation; calling it here serves
    two purposes: it initializes the structured-output backend so the
    scheduler can later allocate bitmask tensors, and it gives us a real
    Future on the request that we can then swap out for our event-gated
    one. Whatever grammar gets compiled in the background by the original
    ``grammar_init`` call is harmless — we never read it."""
    sampling_params = request.sampling_params
    assert sampling_params is not None
    sampling_params.structured_outputs = StructuredOutputsParams(regex="^b+$")
    request.structured_output_request = StructuredOutputRequest.from_sampling_params(
        sampling_params
    )
    sampling_params._validate_structured_outputs(
        runner.vllm_config.structured_outputs_config, tokenizer
    )

    runner.scheduler.structured_output_manager.grammar_init(request)
    # Replace the real (eventually-completing) Future with our event-gated
    # one. The original Future from grammar_init continues to run in the
    # background pool; its result is dropped when we overwrite the slot.
    request.structured_output_request._grammar = _event_gated_grammar_future(ready_event)

    # Request.__init__ set status to WAITING because there was no
    # structured_outputs on sampling_params at construction time. Match what
    # the engine input thread would have produced.
    request.status = RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR


def _accumulate(
    out: ModelRunnerOutput,
    output_tokens: dict[str, list[int]],
) -> None:
    for rid, sampled in zip(out.req_ids, out.sampled_token_ids):
        output_tokens[rid].extend(sampled)


def test_grammar_pending_request_does_not_stall_running_decodes(
    monkeypatch: pytest.MonkeyPatch,
):
    """Two unstructured requests are pre-filled and start decoding. A third
    structured request is added with its grammar Future unresolved. Phase 1:
    drive steps until the runner stops producing tokens — the unstructured
    requests must finish without any scheduler step being wasted while
    they were still decoding. Phase 2: signal the grammar event and verify
    the structured request can then prefill and produce its own tokens."""

    runner = InstrumentedModelRunner.build(
        monkeypatch=monkeypatch,
        max_num_seqs=4,
        max_num_batched_tokens=128,
        available_blocks=100,
    )

    model = REFERENCE_MODELS[InstrumentedModelRunner.DEFAULT_TEST_MODEL]
    tokenizer = get_tokenizer(tokenizer_name=model.name, revision=model.revision)

    unstructured_max_tokens = 8
    unstructured = [
        create_request_for_scheduler_test(
            model=model,
            request_id=i,
            add_step=0,
            max_tokens=unstructured_max_tokens,
            prompt=random_prompt(model=model, seed=i, length=8),
            use_golden_token_injection=False,
            generate_hf_results=False,
        )
        for i in range(2)
    ]

    grammar_ready = threading.Event()
    structured = create_request_for_scheduler_test(
        model=model,
        request_id=2,
        add_step=0,
        max_tokens=4,
        prompt=random_prompt(model=model, seed=100, length=8),
        use_golden_token_injection=False,
        generate_hf_results=False,
    )
    _attach_pending_grammar(structured.request, runner, tokenizer, grammar_ready)

    output_tokens: dict[str, list[int]] = defaultdict(list)
    unstructured_ids = {req.request.request_id for req in unstructured}
    structured_id = structured.request.request_id

    # Submit all three requests through the runner. The runner schedules and
    # executes one step per call; the scheduler will interleave prefills and
    # decodes as it sees fit. The structured request enters
    # WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR and should not be prefilled.
    for req in unstructured:
        _accumulate(runner.execute_new_request(request=req.request), output_tokens)
    _accumulate(runner.execute_new_request(request=structured.request), output_tokens)

    # ---- PHASE 1: drive until the runner is exhausted ----
    # "Exhausted" = three consecutive scheduler steps produced no work. A
    # healthy scheduler stops producing once both unstructured requests
    # finish; a stalled scheduler alternates between empty and decode steps
    # until the unstructured requests eventually finish anyway.
    #
    # During the loop, track every step where the runner produced nothing
    # while at least one unstructured request was still alive (started but
    # not yet at its token budget). Those are the stalls.
    stalled_steps: list[int] = []
    empty_streak = 0
    step_limit = 100  # safety cap; on a healthy scheduler we exit far earlier
    for step in range(step_limit):
        out = runner.execute_running_requests()

        alive_unstructured = {
            rid
            for rid in unstructured_ids
            if 0 < len(output_tokens[rid]) < unstructured_max_tokens
        }

        if not out.req_ids:
            empty_streak += 1
            if alive_unstructured:
                stalled_steps.append(step)
        else:
            empty_streak = 0

        _accumulate(out, output_tokens)

        if empty_streak >= 3:
            break

    # Every unstructured request should have produced all its tokens.
    for rid in unstructured_ids:
        assert len(output_tokens[rid]) >= unstructured_max_tokens, (
            f"unstructured request {rid} only produced "
            f"{len(output_tokens[rid])}/{unstructured_max_tokens} tokens before "
            f"the runner went idle — a grammar-pending request appears to be "
            f"head-of-line blocking it."
        )

    # And no step should have stalled while one was still alive.
    assert not stalled_steps, (
        f"Scheduler stalled on {len(stalled_steps)} step(s) ({stalled_steps}) "
        f"while an unstructured request was still decoding. A grammar-pending "
        f"request in the queue is head-of-line blocking running decodes — the "
        f"scheduler hid the running decodes to switch to prefill mode, but the "
        f"structured request's FSM wasn't ready so no work landed."
    )

    # The structured request must not have advanced yet — grammar still pending.
    assert len(output_tokens[structured_id]) == 0, (
        f"structured request advanced ({len(output_tokens[structured_id])} "
        f"tokens) before its grammar became ready"
    )

    # ---- PHASE 2: signal grammar ready; structured request must progress ----
    grammar_ready.set()

    for _ in range(20):
        _accumulate(runner.execute_running_requests(), output_tokens)
        if output_tokens[structured_id]:
            break

    assert output_tokens[structured_id], (
        "structured request still didn't produce any tokens after the grammar "
        "event was set"
    )
```

</details>

tests/v1/core/test_scheduler_structured_outputs.py
```diff
@@ -441,12 +441,12 @@ def test_sparse_index_grammar_crash(
             pass
 
     # Run prefill of request 1
-    pc_model_runner.execute_new_request(request=request1.request)
+    pc_model_runner.execute_new_request(request1.request)
```
maxdebayser: nit: can you restore the `request=` kw arg style? Then the diff for this file will vanish.

sendnn_inference/v1/worker/spyre_worker.py
```diff
@@ -568,9 +566,10 @@ def _cleanup_model_runner(self, request) -> None:
             **_get_extra_args(),
         )
         self.execute_model(scheduler_output)
-        # satisfy mypy
-        model_runner: ChunkedPrefillModelRunner = cast(ChunkedPrefillModelRunner, self.model_runner)
-        model_runner.tkv = 0
+        if isinstance(self.model_runner, ChunkedPrefillModelRunner):
+            self.model_runner._pending_sampling_state = None
+            # satisfy mypy
```
maxdebayser: This comment is no longer necessary.

maxdebayser:
This is looking great, @dilipgb . One last thing though, are the tests in `tests/v1/core/test_grammar_compilation.py` expected to fail on our current branch? Is it possible to run an experiment that shows this test catching a regression?

tests/v1/worker/test_grammar_bitmask_llm.py
```diff
@@ -0,0 +1,165 @@
...
+import types
+
+import pytest
+
+
+# Force eager CPU + single-process before any vllm imports.
+os.environ.setdefault("SENDNN_INFERENCE_DYNAMO_BACKEND", "eager")
+os.environ.setdefault("VLLM_ENABLE_V1_MULTIPROCESSING", "0")
+os.environ.setdefault("VLLM_USE_V1", "1")
+os.environ.setdefault("HF_HUB_OFFLINE", "1")
```
joerunde: These probably should be migrated into the test setup itself, because other tests that run before this one could pollute the environment. But we generally only ever run cpu tests or spyre tests, not a mix of both, and other tests tend to use fixtures that reset environment overrides after the test. So in practice this probably won't break anything

joerunde:
> are the tests in tests/v1/core/test_grammar_compilation.py expected to fail on our current branch? 

@maxdebayser I think these should pass on main. Grammar compilation was already non-blocking in our request queue before this PR, there just wasn't any test coverage for it

joerunde (approved):
lpgtm! I'd prefer a bit of test setup cleanup in test_grammar_bitmask_llm.py if possible

dilipgb:
> This is looking great, @dilipgb . One last thing though, are the tests in `tests/v1/core/test_grammar_compilation.py` expected to fail on our current branch? Is it possible to run an experiment that shows this test catching a regression?

@maxdebayser on main this won't fail because behaviour remains same on scheduler side. You want me to introduce a regression by removing wait logic and trigger once and see if this fails?

maxdebayser:
@dilipgb , yeah that would be great. It doesn't need to be a script or code change to be commited. Just a one-off experiment to confirm that the test would catch a regression.

dilipgb:
hi @maxdebayser I have used bob to generate some patches that fails the tests and bob suggested to remove few tests which were weak and not contributing to any changes in scheduler. I'm attaching results for same:

Remove promotion loop:
```
python3 -c "
import pathlib
f = pathlib.Path('sendnn_inference/v1/core/scheduler.py')
src = f.read_text()
old = '''            # Promote any requests whose structured-output grammar has become
            # ready, so we correctly classify ready vs not-ready requests.
            for r in list(self.waiting):
                if r.status == RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR:
                    so_req = r.structured_output_request
                    if so_req and so_req.grammar:
                        r.status = RequestStatus.WAITING

            ready_to_prefill'''
new = '            ready_to_prefill'
assert old in src, 'pattern not found'
f.write_text(src.replace(old, new, 1))
print('bug 1 introduced')
"
python3 -m pytest -v tests/v1/core/test_scheduler_structured_outputs.py
git checkout sendnn_inference/v1/core/scheduler.py
bug 1 introduced
=============================================================================================================== test session starts ================================================================================================================
platform linux -- Python 3.12.13, pytest-9.1.1, pluggy-1.6.0 -- /opt/vllm/bin/python3
cachedir: .pytest_cache
rootdir: /home/senuser/sendnn-inference
configfile: pyproject.toml
plugins: anyio-4.14.2, asyncio-1.4.0, forked-1.6.0, mock-3.15.1, timeout-2.3.1
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 5 items                                                                                                                                                                                                                                  

tests/v1/core/test_scheduler_structured_outputs.py::test_grammar_ready_request_promoted_to_waiting FAILED                                                                                                                                    [ 20%]
tests/v1/core/test_scheduler_structured_outputs.py::test_mixed_batch_only_ready_grammar_requests_promoted FAILED                                                                                                                             [ 40%]
tests/v1/core/test_scheduler_structured_outputs.py::test_grammar_not_ready_request_stays_blocked PASSED                                                                                                                                      [ 60%]
tests/v1/core/test_scheduler_structured_outputs.py::test_regular_request_not_blocked_alongside_pending_grammar PASSED                                                                                                                        [ 80%]
tests/v1/core/test_scheduler_structured_outputs.py::test_two_structured_requests_first_finishes_earlier PASSED                                                                                                                               [100%]

===================================================================================================================== FAILURES =====================================================================================================================
__________________________________________________________________________________________________ test_grammar_ready_request_promoted_to_waiting __________________________________________________________________________________________________

mocked_scheduler = <sendnn_inference.v1.core.scheduler.ChunkedPrefillSpyreScheduler object at 0x3ff7afc4d10>

    def test_grammar_ready_request_promoted_to_waiting(mocked_scheduler):
        """A structured-output request whose grammar becomes ready between
        schedule() calls must be promoted WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR →
        WAITING so the scheduler can prefill it.
    
        Regression: if the promotion loop were removed the request would stay
        blocked indefinitely even after the grammar compiled.
        """
        request = _make_structured_request("struct_req")
        assert request.status == RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR
        assert request.structured_output_request.grammar is None
    
        # Simulate grammar compilation completing.
        request.structured_output_request.grammar = Mock()
    
        mocked_scheduler.waiting.append(request)
        mocked_scheduler.schedule()
    
>       assert request.status == RequestStatus.WAITING, (
            "Request with a ready grammar was not promoted from "
            "WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR to WAITING"
        )
E       AssertionError: Request with a ready grammar was not promoted from WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR to WAITING
E       assert <RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR: 2> == <RequestStatus.WAITING: 1>
E        +  where <RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR: 2> = <vllm.v1.request.Request object at 0x3ff7b1887d0>.status
E        +  and   <RequestStatus.WAITING: 1> = RequestStatus.WAITING

tests/v1/core/test_scheduler_structured_outputs.py:135: AssertionError

```
Promote all, ignore grammar readiness:
```
python3 -c "
import pathlib
f = pathlib.Path('sendnn_inference/v1/core/scheduler.py')
src = f.read_text()
old = '                    so_req = r.structured_output_request\n                    if so_req and so_req.grammar:\n                        r.status = RequestStatus.WAITING'
new = '                    r.status = RequestStatus.WAITING'
assert old in src, 'pattern not found'
f.write_text(src.replace(old, new, 1))
print('bug 2 introduced')
"
python3 -m pytest -v tests/v1/core/test_scheduler_structured_outputs.py
git checkout sendnn_inference/v1/core/scheduler.py
bug 2 introduced
=============================================================================================================== test session starts ================================================================================================================
platform linux -- Python 3.12.13, pytest-9.1.1, pluggy-1.6.0 -- /opt/vllm/bin/python3
cachedir: .pytest_cache
rootdir: /home/senuser/sendnn-inference
configfile: pyproject.toml
plugins: anyio-4.14.2, asyncio-1.4.0, forked-1.6.0, mock-3.15.1, timeout-2.3.1
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 5 items                                                                                                                                                                                                                                  

tests/v1/core/test_scheduler_structured_outputs.py::test_grammar_ready_request_promoted_to_waiting PASSED                                                                                                                                    [ 20%]
tests/v1/core/test_scheduler_structured_outputs.py::test_mixed_batch_only_ready_grammar_requests_promoted FAILED                                                                                                                             [ 40%]
tests/v1/core/test_scheduler_structured_outputs.py::test_grammar_not_ready_request_stays_blocked FAILED                                                                                                                                      [ 60%]
tests/v1/core/test_scheduler_structured_outputs.py::test_regular_request_not_blocked_alongside_pending_grammar PASSED                                                                                                                        [ 80%]
tests/v1/core/test_scheduler_structured_outputs.py::test_two_structured_requests_first_finishes_earlier PASSED                                                                                                                               [100%]

===================================================================================================================== FAILURES =====================================================================================================================
______________________________________________________________________________________________ test_mixed_batch_only_ready_grammar_requests_promoted _______________________________________________________________________________________________

mocked_scheduler = <sendnn_inference.v1.core.scheduler.ChunkedPrefillSpyreScheduler object at 0x3ff58270830>

    def test_mixed_batch_only_ready_grammar_requests_promoted(mocked_scheduler):
        """In a mixed waiting queue only the request whose grammar is ready must
        be promoted; the one still compiling must remain blocked.
    
        Regression: a bulk-promote bug could flip all requests to WAITING
        regardless of whether their grammar compiled.
        """
        ready_req = _make_structured_request("ready", arrival_time=0)
        pending_req = _make_structured_request("pending", arrival_time=1)
    
        # Only ready_req has its grammar compiled.
        ready_req.structured_output_request.grammar = Mock()
    
        mocked_scheduler.waiting.append(ready_req)
        mocked_scheduler.waiting.append(pending_req)
        mocked_scheduler.schedule()
    
        assert ready_req.status == RequestStatus.WAITING, (
            "ready_req with compiled grammar was not promoted to WAITING"
        )
>       assert pending_req.status == RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR, (
            "pending_req was promoted despite its grammar not being ready"
        )
E       AssertionError: pending_req was promoted despite its grammar not being ready
E       assert <RequestStatus.WAITING: 1> == <RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR: 2>
E        +  where <RequestStatus.WAITING: 1> = <vllm.v1.request.Request object at 0x3ff58349e20>.status
E        +  and   <RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR: 2> = RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR

tests/v1/core/test_scheduler_structured_outputs.py:161: AssertionError
___________________________________________________________________________________________________ test_grammar_not_ready_request_stays_blocked ___________________________________________________________________________________________________
```
Return sparse padded index instead of dense re-numbered index:
```
python3 -c "
import pathlib
f = pathlib.Path('sendnn_inference/v1/core/scheduler.py')
src = f.read_text()
old = '                if r.status != RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR'
new = '                if r.status == RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR'
assert old in src, 'pattern not found'
f.write_text(src.replace(old, new, 1))
print('bug 3 introduced')
"
python3 -m pytest -v tests/v1/core/test_scheduler_structured_outputs.py
git checkout sendnn_inference/v1/core/scheduler.py
bug 3 introduced
=============================================================================================================== test session starts ================================================================================================================
platform linux -- Python 3.12.13, pytest-9.1.1, pluggy-1.6.0 -- /opt/vllm/bin/python3
cachedir: .pytest_cache
rootdir: /home/senuser/sendnn-inference
configfile: pyproject.toml
plugins: anyio-4.14.2, asyncio-1.4.0, forked-1.6.0, mock-3.15.1, timeout-2.3.1
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 5 items                                                                                                                                                                                                                                  

tests/v1/core/test_scheduler_structured_outputs.py::test_grammar_ready_request_promoted_to_waiting PASSED                                                                                                                                    [ 20%]
tests/v1/core/test_scheduler_structured_outputs.py::test_mixed_batch_only_ready_grammar_requests_promoted PASSED                                                                                                                             [ 40%]
tests/v1/core/test_scheduler_structured_outputs.py::test_grammar_not_ready_request_stays_blocked PASSED                                                                                                                                      [ 60%]
tests/v1/core/test_scheduler_structured_outputs.py::test_regular_request_not_blocked_alongside_pending_grammar PASSED                                                                                                                        [ 80%]
tests/v1/core/test_scheduler_structured_outputs.py::test_two_structured_requests_first_finishes_earlier FAILED                                                                                                                               [100%]

===================================================================================================================== FAILURES =====================================================================================================================
_______________________________________________________________________________________________ test_two_structured_requests_first_finishes_earlier ________________________________________________________________________________________________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x3ff64570290>

    def test_two_structured_requests_first_finishes_earlier(
        monkeypatch: pytest.MonkeyPatch,
    ):
        """Two structured-output requests run concurrently; request1 finishes
        before request2.  After request1 is done, request2 must still produce
        its remaining tokens — verifying that a finishing request doesn't corrupt
        the batch state for the survivor.
    
        Regression: an earlier bug in the bitmask reorder index caused an
        IndexError or wrong-token assignment once the first request left a hole
        in the batch.
        """
        pc_model_runner = InstrumentedModelRunner.build(
            monkeypatch=monkeypatch,
            max_num_batched_tokens=64,
            available_blocks=100,
        )
        pc_model_runner.scheduler.structured_output_manager._use_async_grammar_compilation = False
    
        model = REFERENCE_MODELS[InstrumentedModelRunner.DEFAULT_TEST_MODEL]
        tokenizer = get_tokenizer(tokenizer_name=model.name, revision=model.revision)
    
        prompt1 = random_prompt(model=model, seed=0, length=64)
        request1 = create_request_for_scheduler_test(
            model=model,
            request_id=0,
            add_step=0,
            max_tokens=3,
            prompt=prompt1,
            use_golden_token_injection=False,
            generate_hf_results=False,
        )
        request2 = create_request_for_scheduler_test(
            model=model,
            request_id=1,
            add_step=0,
            max_tokens=4,
            prompt=prompt1,
            use_golden_token_injection=False,
            generate_hf_results=False,
        )
    
        for request in [request1, request2]:
            assert (sampling_params := request.request.sampling_params) is not None
            sampling_params.structured_outputs = StructuredOutputsParams(regex=".*")
            request.request.structured_output_request = StructuredOutputRequest.from_sampling_params(
                sampling_params
            )
            sampling_params._validate_structured_outputs(
                model_config=pc_model_runner.vllm_config.model_config,
                structured_outputs_config=pc_model_runner.vllm_config.structured_outputs_config,
                tokenizer=tokenizer,
            )
            pc_model_runner.scheduler.structured_output_manager.grammar_init(request.request)
    
            assert (structured := request.request.structured_output_request) is not None
            while not structured.is_grammar_ready:
                pass
    
        req2_id = request2.request.request_id
        req2_max_tokens = request2.request.sampling_params.max_tokens
        req2_tokens: list[int] = []
    
        def _collect(out) -> None:
            for rid, toks in zip(out.req_ids, out.sampled_token_ids):
                if rid == req2_id:
                    req2_tokens.extend(toks)
    
        # Prefill request1, one decode, then prefill request2.
        # The prefill outputs (execute_new_request) also carry sampled tokens,
        # so collect from all steps including the request2 prefill.
        pc_model_runner.execute_new_request(request=request1.request)
        pc_model_runner.execute_running_requests()
        _collect(pc_model_runner.execute_new_request(request=request2.request))
    
        # Drive remaining decodes. request1 finishes after 2 more steps
        # (has 1 token already, max=3), leaving request2 alone.
        # request2 must still receive all its remaining tokens.
        for _ in range(6):
            _collect(pc_model_runner.execute_running_requests())
            if len(req2_tokens) >= req2_max_tokens:
                break
    
>       assert len(req2_tokens) == req2_max_tokens, (
            f"request2 received {len(req2_tokens)} tokens after request1 finished, "
            f"expected {req2_max_tokens}. "
            "A batch-state corruption after the first request finished may be the cause."
        )
E       AssertionError: request2 received 0 tokens after request1 finished, expected 4. A batch-state corruption after the first request finished may be the cause.
E       assert 0 == 4
E        +  where 0 = len([])

tests/v1/core/test_scheduler_structured_outputs.py:296: AssertionError
```

sendnn_inference/v1/core/scheduler.py
```diff
@@ -475,15 +474,15 @@ def schedule(self, throttle_prefills: bool = False) -> "SchedulerOutput":
                 if r.status != RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR
             ]
             if ready_to_prefill:
-                new_prefill_candidates = list(self.waiting)
```
maxdebayser, line 478: Looks like this was a bug in the old code

sendnn_inference/v1/core/scheduler.py
```diff
@@ -461,7 +460,7 @@ def schedule(self, throttle_prefills: bool = False) -> "SchedulerOutput":
 
         # Check new requests to prefill
         elif len(self.waiting) > 0:
-            # Try to promote grammar-waiting requests whose FSM is now
+            # Promote any requests whose structured-output grammar has become
             # ready, so we correctly classify ready vs not-ready requests.
             for r in list(self.waiting):
```
maxdebayser, line 466: As a suggestion, the for loop and the list comprehension can be combined to avoid iteration twice over the elements:
```python
            ready_to_prefill = []
            for r in list(self.waiting):
                if r.status == RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR:
                    so_req = r.structured_output_request
                    if so_req and so_req.grammar:
                        r.status = RequestStatus.WAITING
                        ready_to_prefill.append(r)
                else:
                    ready_to_prefill.append(r)
```

maxdebayser (approved):
Thanks for double-checking the tests @dilipgb . It now looks good to me. I've left a small suggestion to improve the state of a piece of code that wasn't touched by your PR. But that's optional.

PR #988, opened 2026-06-02
:bug: fix model caches

joerunde:
## Description

This updates the model caching script to ignore unused artifacts, and adds a check that we don't blow up the cache with a single model. The pooling models are currently >10GB each because we starting pulling in all of the alternative weight formats and tried to cache them.

## Test Plan

the usual GHA CI

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -264,6 +264,28 @@ jobs:
 
           wait
 
+      - name: "Check HF model cache sizes"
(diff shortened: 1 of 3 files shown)


joerunde:
Verified that the size check works, the previous run with the 2GB limit all failed since the g3.3 model artifacts are 2.3GB uncompressed. Models all seem to be functioning fine with the newly ignored artifacts 👍

rafvasq (approved):
LGTM

PR #989, opened 2026-06-03
Add FMS pooling model support

ktuvw:
## Description

Adds support for FMS models in PoolingModelRunner

## Related Issues

Relates to #1807

## Test Plan

Currently tested
1. Created a vLLM server with Qwen/Qwen3-Embedding-0.6B
2. Created an embedding with curl request to server
3. Compared vLLM Spyre embedding with CPU embedding using cosine similarity (0.999814)

Will test
1. batch accuracy
2. Qwen/Qwen3-Embedding-4B
3. vllm bench serve throughput perf

## Checklist

### DIFF
--- a/.github/ci_model_cache.yaml
+++ b/.github/ci_model_cache.yaml
@@ -13,3 +13,5 @@ models:
     revision: cf74d8acd4f198de950bf004b262e6accfed5d2c
   - repo: cross-encoder/stsb-roberta-large
     revision: 2b12c2c0088918e76151fd5937b7bba986ef1f98
+  - repo: Qwen/Qwen3-Embedding-0.6B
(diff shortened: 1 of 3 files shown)


maxdebayser (commented):
Very nice! I think what is left to do is adding the model in the know model configuration list: https://github.com/torch-spyre/sendnn-inference/blob/main/sendnn_inference/config/model_configs.yaml

maxdebayser:
Since the code path here is a bit different than for the other embedding models, running the embedding unit tests with this model would be better but we have restrictions on how many models we can test in github actions. Wdty, @joerunde ?

maxdebayser:
bot:test

maxdebayser:
bot:test

maxdebayser (approved):

PR #990, opened 2026-06-03
Introducing Sim mode

yannicks1:
## Description

Sim-mode lets us run the full plugin stack but mocks the forward pass (no op), while still:

- Exercising all the real plugin code paths so we can catch correctness/constraint bugs without execution of model forward passes on either CPU or Spyre.
- Producing a rough perf estimate for arbitrary workloads, by charging a user-supplied virtual cost per prefill chunk and per decode step. Useful for scheduler exploration and for quickly previewing the gain from features like request pausing without waiting for a real hardware run.

## How it works
When `SENDNN_INFERENCE_SIM_MODE=1`, the runner swaps `SpyreCausalLM` for a no-op `MockSpyreCausalLM`  and a `SimState` singleton accumulates a virtual clock alongside the real one. 
Each forward step charges `SENDNN_INFERENCE_SIM_PREFILL_MS` or `SENDNN_INFERENCE_SIM_DECODE_MS` regardless of shape. 
When a request finishes, its virtual lifecycle (per-token timestamps, prefill/decode durations, ITLs) is written to `sim_metrics.jsonl`, in roughly the same shape as the existing `request_metrics.jsonl` so analysis tooling carries over. Wall-clock `request_metrics.jsonl` writes are skipped in sim mode since they would all be 0.

Code reuse: The `MockSpyreCausalLM` from tests/ was reused for this implementation. It has been moved into the package sendnn_inference and is imported for the `InstrumentedModelRunner` in tests/. 

## Knobs
- `SENDNN_INFERENCE_SIM_MODE=1`: activate the feature (default 0)
- `SENDNN_INFERENCE_SIM_PREFILL_MS`: virtual cost in milliseconds per prefill chunk (default 0)
- `SENDNN_INFERENCE_SIM_DECODE_MS`: virtual cost in milliseconds per prefill chunk (default 0)


maxdebayser:
Given that we can run in eager mode and that it also exercise the full model code, what is the use case for this?

yannicks1:
> Given that we can run in eager mode and that it also exercise the full model code, what is the use case for this?

for constraint checking: This mocks the fwd call (takes ~0 ms), eager CPU execution fwd calls are taking a lot of time. 

for perform approx: By providing a rough estimate of values for prefill and decode times, I was already able to get an approx of the vllm bench perf within 1-5%. This simulated bench takes magnitudes less time than running it for real (obviously not possible to get this data with eager on CPU)

maxdebayser:
> for perform approx: By providing a rough estimate of values for prefill and decode times, I was already able to get an approx of the vllm bench perf within 1-5%. This simulated bench takes magnitudes less time than running it for real (obviously not possible to get this data with eager on CPU)

Makes sense to me. Thanks!

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1558,6 +1578,16 @@ def execute_model(
                 is_prompt=model_input.is_prompt,
             )
 
+        if envs_spyre.SENDNN_INFERENCE_SIM_MODE:
+            from sendnn_inference.v1.sim import get_sim_state
```
sducouedic: imo importing this function at every execute step isn't the best, should we do it in init and cache the function or at the beginning of the file, without checking `SENDNN_INFERENCE_SIM_MODE`

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1460,7 +1466,21 @@ def _update_batch(self, scheduler_output: SchedulerOutput):
             req_state.num_computed_tokens = num_computed_tokens
 
         if scheduler_output.finished_req_ids:
+            sim = None
+            if envs_spyre.SENDNN_INFERENCE_SIM_MODE:
+                from sendnn_inference.v1.sim import get_sim_state
```
sducouedic: same comment as in execute_model, `_update_batch` is called very often and importing get_sim_state every time seems suboptimal to me

sendnn_inference/envs.py
```diff
@@ -171,6 +171,21 @@ def clear_env_cache():
     "SENDNN_INFERENCE_MM_DEVICE": lambda: parse_mm_device(
         os.getenv("SENDNN_INFERENCE_MM_DEVICE", "auto")
     ),
+    # Sim-mode: replace the real Spyre forward with a no-op model and substitute
+    # virtual durations into request_metrics.jsonl. Lets us exercise scheduler /
+    # runner / batching logic on a laptop without AIU hardware. Use with
+    # DYNAMO_BACKEND=eager.
```
sducouedic, line 186: nit: I understand that it makes more sense to run this mock model on cpu, but is there any reason for it to not work with sendnn backend? if yes I would enforce eager backend in the code, or adapt a bit the comment to make it clear that we can use it with any backend.

tests/v1/worker/mock_model.py
```diff
@@ -26,81 +22,6 @@
 from spyre_util import REFERENCE_MODELS, patch_environment
 
 
-class MockSpyreCausalLM:
```
sducouedic, line 29: very nice, I never realized we already had the perfect class to do perf estimation. It is super cool that we can use it both for tests and quick performance evaluations for new features :)

sendnn_inference/v1/sim.py
```diff
@@ -0,0 +1,279 @@
...
+            "prefill_time_seconds": prefill_time,
+            "inference_time_seconds": inference_time,
+            "decode_time_seconds": decode_time,
+            "time_to_first_token_seconds": ttft,
+            "mean_time_per_output_token_seconds": mean_tpot,
+            "inter_token_latencies_seconds": itls,
+        }
+        self._ensure_file()
+        assert self._fp is not None
+        self._fp.write(json.dumps(record) + "\n")
```
sducouedic: isn't there also a race condition here, I guess we need a lock here as well

sendnn_inference/v1/sim.py
```diff
@@ -0,0 +1,279 @@
...
+                rec = self._records.get(rid)
+                if rec is None:
+                    rec = _RequestSimRecord(virtual_arrival=self.virtual_clock_seconds)
+                    self._records[rid] = rec
+                if is_prompt:
+                    rec.num_prefill_chunks += 1
+                    rec.last_prefill_end = end_t
+                else:
+                    rec.decode_step_ends.append(end_t)
+                rec.virtual_completion = end_t
```
sducouedic: here for tp > 1 how would it work, could it happen that with tp say 4 we record 4 times the number of prefill chunks, same for `rec.decode_step_ends.append(end_t)` would we append four times the decode time

sducouedic (commented):
This is a very cool idea, we should use it with long runs on scheduler changes

I left a few comments. But I have a general concert with tp > 1, if it could happen that decode and prefills gets recorded too many times, one for each model runner. Have you tried running with tp > 1?

sendnn_inference/v1/sim.py
yannicks1: there is no reason to run the sim model with tp > 1. tensor parallelism would happen in the forward pass which is a no-op for the sim model. But it is valid concern. let me add an assert to enforce tp=1 for the sim case to be on the save side

sendnn_inference/envs.py
yannicks1: I do not see a reason to ensure that it also work with sendnn backend. The forward pass is the only part that would run on Spyre and its precisely what we mock (no-op) in the sim case. So the only thing that running sim mode with backend sendnn would probably do is checking that Spyre/ SW stack is there (during startup) and then not use it during execution. But valid point, let me assert that backend has to be eager when in sim mode!

yannicks1:
bot:test

yannicks1:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=4

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -723,11 +723,24 @@ def __init__(
...
+
+            self._sim_state = get_sim_state()
+            self._mock_causal_lm = MockSpyreCausalLM
+
     def load_model(self) -> None:
-        self._model = SpyreCausalLM(
-            vllm_config=self.vllm_config,
-            rank=self.rank,
-        )
+        if envs_spyre.SENDNN_INFERENCE_SIM_MODE:
```
joerunde: I would prefer inheritance over putting if/else in the real model runner code, so that we're neither adding complexity nor defect surface to the real model runner:

```
class SimulatedSpyreModelRunner(ChunkedPrefillModelRunner)
```

maxdebayser: I agree, that would be cleaner.

sendnn_inference/v1/sim.py
```diff
@@ -0,0 +1,280 @@
...
+            "inference_time_seconds": inference_time,
+            "decode_time_seconds": decode_time,
+            "time_to_first_token_seconds": ttft,
+            "mean_time_per_output_token_seconds": mean_tpot,
+            "inter_token_latencies_seconds": itls,
+        }
+        with self._lock:
+            self._ensure_file()
+            assert self._fp is not None
+            self._fp.write(json.dumps(record) + "\n")
```
maxdebayser: Suggestion: make `self_.ensure_file()` a context manager

sendnn_inference/v1/sim.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+_sim_state: SimState | None = None
+
+
+def get_sim_state() -> SimState:
+    global _sim_state
+    if _sim_state is None:
+        _sim_state = SimState()
+    return _sim_state
```
maxdebayser, line 315: The only place where _sim_state is referenced is the model runner, so it can me an attribute instead of a global singleton.

maxdebayser (changes requested):
Very nice, especially test mock model code reuse. I've left a few suggestions around the code structure.

sendnn_inference/v1/sim.py
```diff
@@ -0,0 +1,280 @@
...
+        token_emit_times.extend(rec.decode_step_ends)
+        num_generation_tokens = len(token_emit_times)
+
+        if num_generation_tokens == 0:
+            # Request never produced a token (e.g., immediate cancel). Skip.
+            return
+
+        first_token_t = token_emit_times[0]
+        last_token_t = token_emit_times[-1]
+        ttft = first_token_t - rec.virtual_arrival
```
sducouedic, line 264: does the ttft still capture the waiting time in the queue this way? I see that `rec.virtual_arrival` is set with the first apparition of the request in `record_step`, which is called by the `model_runner.execute_model()` method, meaning it corresponds to the time of first chunked prefill

sendnn_inference/v1/sim.py
```diff
@@ -0,0 +1,280 @@
...
+    virtual_arrival: float
+    last_prefill_end: float | None = None
+    decode_step_ends: list[float] = field(default_factory=list)
+    virtual_completion: float | None = None
+    num_prefill_chunks: int = 0
+
+
+class SimState:
+    def __init__(self) -> None:
+        self.virtual_clock_seconds: float = 0.0
```
sducouedic, line 162: I think we should set the `virtual_clock_seconds` with the real `time.time()` and maintain only the difference to the real time. This would allow to capture all the overheads in the engine in addition to just the prefill time and decode time, if anything becomes suboptimal outside of `self.model()`, it will be captured by the sim-model:

initialization:
* `self.virtual_clock_seconds: float = time.time()`
* `self.diff_to_real_time: float = 0.0`

update in `record_step()`:
* `self.diff_to_real_time += step_seconds`
* `self.virtual_clock_seconds = time.time() + self.diff_to_real_time`  <-- we add the diff to time.time() instead of incrementing the virtual clock directly

this would also fix the issue of missing the waiting time in the ttft: ~we add `self.diff_to_real_time` to the existing `req.arrival_time` value to get the virtual arrival time~ edit: actually this wouldn't work correctly because `self.diff_to_real_time` is the current difference *now*, the `req.arrival_time` is in the past. So probably waiting time can be computed with `time.time() - req.arrival_time` (keep real time reference), then added to the ttft: `ttft = time_last_token - time_first_chunked_prefill + waiting_time`. Hope that makes sense.

yannicks1:
I am currently rethinking this. please wait with further review

sendnn_inference/v1/sim.py
yannicks1: good catch, I did add this and now rec.virtual_arrival is set upon request arrival

PR #991, opened 2026-06-03
add bob skill for vllm updates

rafvasq:
## Description

- Adds bob skill for vllm version bump automation 
  - Process mirrors the existing claude skill, slightly modified because Bob doesn't support passing arguments directly

## Related Issues

- Follow up to https://github.com/torch-spyre/sendnn-inference/pull/987 which adds a claude skill for the same thing

## Checklist


joerunde:
@rafvasq is there any way to share a markdown file for these instead of duplicating? I'd be fine if it needs to be updated a bit to not pass arguments

maxdebayser (approved):
It makes sense to me. Hopefully it also does for the agents.

PR #992, opened 2026-06-04
:sparkles: Deduplicate MM vision encoder TP ranks

gkumbhat:
## Description

1. Add a utility for sharing MM encoding across workers using shared memory. 
2. Make MM embedding only get computed on rank=0 worker, to avoid redundant computation. We then share embedding with other workers using utility above.

As the image sizes increases, so does compute requirement for vision encoder.  We have seen 20-29% improvement with this PR as compared to current main when tested with images with image dimension of 2k.  With 1K images, we also see slight improvement of about 2s (TTFT) on x86.

## Checklist


sendnn_inference/platform.py
```diff
@@ -345,7 +345,7 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 scheduler_config.max_num_batched_tokens = (
                     model_config.max_model_len * scheduler_config.max_num_seqs
                 )
-                cache_config.block_size = model_config.max_model_len  # ty: ignore[invalid-assignment]
+                cache_config.block_size = model_config.max_model_len
```
gkumbhat, line 348: type check was failing for some reason on this branch, so I just added these for tests to pass

sendnn_inference/argparse_utils.py
```diff
@@ -162,7 +162,7 @@ def patched_parse_args(
             namespace: argparse.Namespace | None = None,
         ) -> argparse.Namespace:
             result = original_parse_args(self, args, namespace)
-            assert result is not None  # type: ignore[redundant-expr]
+            assert result is not None
```
gkumbhat, line 165: type check was failing for some reason on this branch, so I just added these for tests to pass

sendnn_inference/v1/core/scheduler.py
```diff
@@ -336,7 +336,7 @@ def schedule(self) -> "SchedulerOutput":
             ready_to_prefill = [
                 r
                 for r in self.waiting
-                if r.status != RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR  # type: ignore[attr-defined]
+                if r.status != RequestStatus.WAITING_FOR_STRUCTURED_OUTPUT_GRAMMAR
```
gkumbhat, line 408: type check was failing for some reason on this branch, so I just added these for tests to pass

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1034,44 +1044,93 @@ def _prepare_chunked_prefill(self, req_id: str) -> SamplingForwardInputs:
...
+                        )
+                    else:
+                        meta = torch.zeros(4, dtype=torch.int64)
+
+                    torch.distributed.broadcast(meta, src=0)
+
+                    if self.rank != 0:
+                        shape = (int(meta[0]), int(meta[1]), int(meta[2]))
+                        dtype = _MM_EMBED_IDX_TO_DTYPE[int(meta[3])]
+                        full_embeds = read_embeddings(req_id, shape, dtype)
```
joerunde: Is there a potential for a deadlock here? If anything between the broadcasts raises then that rank will never hit the second one and the other ranks will deadlock waiting for it

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -38,6 +39,13 @@
 from sendnn_inference.platform import SpyrePlatform
 from sendnn_inference.utils import exact_div
 from sendnn_inference.v1.sample.spyre_logits_processor import build_logitsprocs_for_cb
+from sendnn_inference.v1.worker.mm_shared_memory import (
+    _DTYPE_TO_IDX as _MM_EMBED_DTYPE_TO_IDX,
+    _IDX_TO_DTYPE as _MM_EMBED_IDX_TO_DTYPE,
```
joerunde: no private imports!

Could instead expose these as methods e.g. `read_dtype()`

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1034,44 +1044,93 @@ def _prepare_chunked_prefill(self, req_id: str) -> SamplingForwardInputs:
...
-            full_embeds = self.model.get_maybe_mm_embeddings(
-                full_input_tokens,
-                mm_features=mm_features,
-                is_decode=False,
-            )
-
-            t_elapsed = time.time() - t0
+            if self.rank == 0:
+                t0 = time.time()
+                full_embeds = self.model.get_maybe_mm_embeddings(
```
joerunde: ah actually the deadlock concern is valid here too, if this raises then the other ranks will all be sitting at the first broadcast that will never come :(

This should be robust to failures both ways- either rank 0 failing to generate embeddings or other ranks failing to read them, and ideally we should have tests to cover that.

Since this is already in the forward pass though, I think that means that we'd have to crash the server anyway unless we add better handling to extract this request from the batch and raise an error for it elsewhere 🤔. (But that's still way better than deadlocking)

sendnn_inference/v1/worker/spyre_model_runner.py
gkumbhat: hmm.. good point. There is a possibility.. Need to move torch.distributed to finally, so that it always gets executed even if there is a exception.

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -1034,44 +1044,93 @@ def _prepare_chunked_prefill(self, req_id: str) -> SamplingForwardInputs:
         self.model.indices = torch.ones(1, dtype=torch.bool, device="cpu")
 
         # For multimodal requests, compute embeddings once for the full sequence
-        # and cache them, then slice per chunk. This ensures image features are
-        # correctly aligned across all chunks.
+        # and cache them, then slice per chunk.  With tensor parallelism (TP>1) each
+        # worker is a separate process; we avoid running the vision encoder world_size
+        # times by having rank 0 compute and broadcast the result to all other ranks.
```
joerunde, line 1222: This adds a big block of logic that's specific to broadcasting mm tensors. This should be encapsulated elsewhere instead of being inlined here

joerunde (changes requested):
I like how simple this is and the improvements look 🌶️🌶️🌶️!

Just needs some improvements for robustness, once we can prove this won't deadlock then we can ship it

sendnn_inference/v1/worker/spyre_model_runner.py
gkumbhat: moved outside to a single function

sendnn_inference/v1/worker/spyre_model_runner.py
gkumbhat: Fix it, so now if either generate embedding fails, or rank 0 fails, other ranks fail, we would have broadcast signal published so none of these hang.

But yeah, currently if something like that happens where we fail to read or write the embedding, then now the server itself will crash.. not ideal, but recovery mechanism can potentially be added in next iteration ?

gkumbhat:
bot:test

joerunde (approved):
ship it!

PR #993, opened 2026-06-04
Respect left padding in kv cache manager

yannicks1:
## Description

The upstream KV cache manager was over-allocating blocks during chunk 0 of left-padded prefills. Specifically one extra block per padding block was added, accounting for the masked-out padding region. The block count reconciled with reality on the next chunk (later chunks naturally fall on chunk boundaries), but during chunk 0 the KV cache manager view diverged from what the actual block table in the model runner looks like, which made test references confusing. 

### Example: 
With chunk_size=128, block_size=64, prompt length 192 → left padding = 1 block. Chunk 0 of a fresh request used to allocate [1, 2] (2 blocks) where block 2 was reserved-but-unwritten until chunk 1; after the fix it allocates [1] and chunk 1 allocates [2, 3] directly. End state is identical ([1, 2, 3] after chunk 1 either way). 

## Implementation
changes: 
- In `ChunkedPrefillSpyreScheduler.schedule()`, set `long_prefill_token_threshold = chunk_size − left_padding` for the chunk-0 step of fresh  requests with no chunk-0 cache hit. The base scheduler then asks the KV cache manager for exactly the real-content blocks of chunk 0. Subsequent chunks land on natural chunk boundaries and converge to the same final block layout.
- `adjust_computed_tokens` is now a no-op and has been removed
- The worker is unchanged

## Test Plan

Test updates: in `tests/e2e/test_spyre_pc_scheduler_steps.py`, `n_used_blocks` and `block_tables` references for the chunk-0 step of left-padded requests drop the transiently over-allocated block(s). 

## Checklist


sendnn_inference/v1/core/scheduler.py
```diff
@@ -250,8 +250,40 @@ def adjust_computed_tokens(
...
+        # If the prefix cache already covers chunk 0, no cap is needed: the
+        # base scheduler will start from chunk i>=1, which has no padding.
+        # `get_computed_blocks` records into `prefix_cache_stats` as a side
+        # effect; the base scheduler calls it again, so toggle log_stats off
+        # here to avoid double-counting.
+        prev_log_stats = self.kv_cache_manager.log_stats
+        self.kv_cache_manager.log_stats = False
+        _, prefix_len = self.kv_cache_manager.get_computed_blocks(new_prefill)
+        self.kv_cache_manager.log_stats = prev_log_stats
+        if prefix_len > 0:
```
maxdebayser: If prefix len is in blocks, shouldn't the condition here be `prefix > blocks_per_chunk` to check that were not within the first chunk?

sendnn_inference/v1/core/scheduler.py
```diff
@@ -250,8 +250,40 @@ def adjust_computed_tokens(
...
+        if not new_prefill_candidates:
+            return self.chunk_size
+
+        new_prefill = new_prefill_candidates[0]
+        # If the prefix cache already covers chunk 0, no cap is needed: the
+        # base scheduler will start from chunk i>=1, which has no padding.
+        # `get_computed_blocks` records into `prefix_cache_stats` as a side
+        # effect; the base scheduler calls it again, so toggle log_stats off
+        # here to avoid double-counting.
+        prev_log_stats = self.kv_cache_manager.log_stats
```
maxdebayser: Instead of calling the kv cache manager, you can also do `prefix = math.ceil(new_prefill.num_computed_tokens / self.block_size)`

maxdebayser (commented):
Nice, this is a detail that we completely overlooked in the past. But today while adding tests for the block reservation I noticed that the block use was off by one due to not taking the padding into account.

Overall this looks good to me, but I left a few questions.

sendnn_inference/v1/core/scheduler.py
yannicks1: I do no think so, at this point in time the new_prefill.num_computed_tokens is still 0. it only gets updated after calling super(SpyreScheduler, self).schedule() in line 401

sendnn_inference/v1/core/scheduler.py
yannicks1: Excellent catch! The `prefix_len` here represents tokens, not blocks (despite what the function name might suggest 🤷, I updated the name `prefix_len` -> `prefix_token_len`). The issue is that this value doesn't yet enforce Spyre's chunk-alignment constraint for prefixes.

Simply checking `prefix_len > 0` is insufficient because it would also match partial prefixes within chunk 0. If prefixes were guaranteed to be chunk-aligned, then `prefix_len > 0` would reliably indicate at least a full chunk 0 match. Without that guarantee, we risk incorrectly treating incomplete chunk 0 prefixes as valid cache hits. I updated the condition to `prefix_token_len >= self.chunk_size - left_padding`.

maxdebayser:
I have one last question? Do we still need need `adjust_computed_tokens` at all?  I can't think of a case where `prefix_cache_len > computed_tokens` true. But I'm also a bit tired so now, so I'm not sure.

sducouedic:
bot:test

sducouedic:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=4

yannicks1:
> I have one last question? Do we still need need `adjust_computed_tokens` at all? I can't think of a case where `prefix_cache_len > computed_tokens` true. But I'm also a bit tired so now, so I'm not sure.

I think you are right, but also have to think it through rigorously.

maxdebayser:
@yannicks1 , I've opened a PR on your branch to solve the merge conflicts: https://github.com/yannicks1/sendnn-inference/pull/4

yannicks1:
bot:test

yannicks1:
bot:bench
NUM_PROMPTS=1000
MAX_RUN_TIME=36000
IGNORE_EOS=1
CUSTOM_OUTPUT_LEN=-1
MAX_CONCURRENT=4

yannicks1:
bot:test

maxdebayser (approved):
There is an FP8 test failing, but I don't think it's related to this PR.

yannicks1:
bot:test

PR #1001, opened 2026-06-09
Revert conditional kv_cache assignment to save memory

Daniel-Schenker:
### Summary
This PR reverts the changes introduced in [#905](https://github.com/torch-spyre/sendnn-inference/pull/905), which made the kv_cache assignment conditional based on Spyre card usage.

### Background
PR #905 modified the kv_cache assignment logic (which uses dummy values when running on Spyre) from unconditional to conditional assignment. This change was motivated by profiling results indicating a ~1ms performance overhead from the assignment operation.

### Issue
While the conditional assignment provided a minor performance improvement, it introduced a significant memory overhead: torch.compile allocates an additional ~20GB of memory to accommodate the branching code paths.

### Decision
After team discussion, we determined that memory conservation is more critical than the marginal performance gain. Therefore, we are reverting PR #905 to restore the original unconditional assignment behavior.

## Checklist

### DIFF
--- a/sendnn_inference/model_executor/model_loader/spyre.py
+++ b/sendnn_inference/model_executor/model_loader/spyre.py
@@ -475,11 +475,11 @@ class SpyreCausalLM(nn.Module):
 
         # The second item in the output tuple is the KV cache.
         # However, on spyre these are ghost tensors- the data in these tensors does not reflect the
-        # actual kv cache data on the device. They exist only for proper compilation, so we don't
-        # waste any time assigning these tensors back to anything.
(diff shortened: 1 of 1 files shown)


maxdebayser:
That's sad, the ITL improvement was actually pretty good. But if it's an urgent problem we can investigate more later.

I'm intrigued by why this revert made a difference. Because the compiler only ever sees the `SpyrePlatform.is_backend_sendnn_enabled()` flag in one state during the lifetime of the server, or not? Assuming that is true, there should be no diverging code path.

PR #1005, opened 2026-06-10
:bug: Pull in FMS fixes for text-only granite-vision crash

joerunde:
## Description

- Updates the lower bound for FMS to fix granite vision crashes
- Refreshes all packages

## Checklist


pyproject.toml
```diff
@@ -54,6 +54,7 @@ git_describe_command = "git describe --dirty --tags --long --match 'v*'"
 # by accident
 override-dependencies = [
     "torch==2.11.0",
+    "torchvision==0.26.0",
```
joerunde, line 57: technically required if we want to pin `torch` this way- otherwise resolution will pick torch 2.12 + torchvision 0.27, then override torch back to 2.11 and break

joerunde:
bot:test

sducouedic:
bot:test

joerunde:
The only spyre failure here was the scoring url issue, so this should be good to go 🤞

.github/workflows/model_smoke.yml
```diff
@@ -62,9 +62,13 @@ jobs:
 
     steps:
       - name: "Lightweight disk cleanup"
+        # all rm -rf, no apt/docker — completes in < 1s and frees ~20 GB
         run: |
-          rm -rf /usr/share/swift
-          rm -rf /user/local/share/chromium
+          sudo rm -rf /usr/local/lib/android
```
maxdebayser, line 67: That's a bit odd

maxdebayser (approved):
LGTM

.github/workflows/model_smoke.yml
joerunde: that android exists, or that we remove it?

(We were occasionally running out of space during CI runs so I had to delete a bit more stuff)

PR #1009, opened 2026-06-15
Custom bench metrics

sducouedic:
## Description

This PR allows to run benchmarks with additional custom metrics in addition to the regular TTFT, ITL values. It patches `vllm bench` to reuse the base implementation, but on top of it collects additional metrics such as the waiting time in the queue, or the waiting time in the queue or the number of chunked prefill count.

Note: we can use the client patching to collect and print information from the SimModel in PR #990. Additionally the patching should also allow to inject real number of output tokens to SimModel (because as a sim model, it cannot by itself generate a realistic number of output tokens), enabling more realistic perf simulation without `--ignore-eos` and `--custom-output-len -1` parameters.

### Contributions

* `sendnn-bench serve` instead of `vllm bench`. We register a custom `spyre-chat` backend that collects per-request custom metrics from SSE responses, in addition to the existing ttft, itl values
* Injection of per-request sendnn results into the existing output JSON file created by `--save-detailed`
* using `--describe-metrics` flag saves the sendnn metrics descriptions to a `sendnn_bench_metrics_description.txt` file
* Detailed `timeline.html` which displays for each request the waiting time and individual chunked prefill times (instead of only the ttft), as well as waiting time of decodes
* test for the patching and metrics collection
* A claude skill to easily add new metrics. The user has to give a detailed description of the new metric, how and where to compute the value, then the skill allows Claude to know all the places that require modifications for the new metric to appear in the .json output file and printed result. It also tells how to adapt the tests to integrate the new metrics. Examples:
  
    > **/add-bench-metric** add a metric for the prefix cache hit percent, which is based on the number of chunks saved from cache hit over expected number of prefill chunks, for each request. The expected number of chunked prefill for a given request is math.ceil(request.num_prompt_tokens / self.chunk_size)

   > **/add-bench-metric** a list of number of left-padding blocks for the request, one value for each decode step. The list for each request should be the same length of as the decode_latencies or decode_start_times. Ignore the exisitng left_padding variables et recompute it completely: left_padding = max_num_blocks - req_num_blocks, where max_num_blocks = math.ceil(tkv / block_size), and req_num_blocks = math.ceil(req.num_computed_tokens / block_size)
* Usage docs: https://vllm--1009.org.readthedocs.build/projects/spyre/en/1009/user_guide/detailed_performance_measurement.html

### Usage

1. Export `SENDNN_INFERENCE_BENCH_METRICS_ENABLED=1` to indicate to the server to collect sendnn metrics, and also set the patchings
2. Launch the server as usual: `vllm serve {model} --max-model-len {model-len} --max-num-seqs {num_seqs}` 
3. Launch the benchmarking client `spyre-bench serve  --model ibm-granite/granite-3.3-8b-instruct --save-result --describe-metrics ...`  

Output:
```
============ Serving Benchmark Result ============
Successful requests:                     XXXX
Failed requests:                         XXXX
Maximum request concurrency:             XXXX
Benchmark duration (s):                  XXXX
Total input tokens:                      XXXX
Total generated tokens:                  XXXX
Request throughput (req/s):              XXXX
Output token throughput (tok/s):         XXXX
Peak output token throughput (tok/s):    XXXX
Peak concurrent requests:                XXXX
Total token throughput (tok/s):          XXXX
---------------Time to First Token----------------
Mean TTFT (ms):                          XXXX
Median TTFT (ms):                        XXXX
P99 TTFT (ms):                           XXXX
P100 TTFT (ms):                          XXXX
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          XXXX
Median TPOT (ms):                        XXXX
P99 TPOT (ms):                           XXXX
P100 TPOT (ms):                          XXXX
---------------Inter-token Latency----------------
Mean ITL (ms):                           XXXX
Median ITL (ms):                         XXXX
P99 ITL (ms):                            XXXX
P100 ITL (ms):                           XXXX
----------------End-to-end Latency----------------
Mean E2EL (ms):                          XXXX
Median E2EL (ms):                        XXXX
P99 E2EL (ms):                           XXXX
P100 E2EL (ms):                          XXXX
================= SenDNN Metrics =================
Total prefill chunks processed:          XXXX
---------------- Queue Wait Time -----------------
Mean Queue Wait Time (ms):               XXXX
Median Queue Wait Time (ms):             XXXX
P99 Queue Wait Time (ms):                XXXX
P100 Queue Wait Time (ms):               XXXX
------------- Chunked Prefill Count --------------
Mean Num Chunked Prefills:               XXXX
Median Num Chunked Prefills:             XXXX
P99 Num Chunked Prefills:                XXXX
P100 Num Chunked Prefills:               XXXX
------------ Chunked Prefill Latency -------------
Mean Chunk Prefill Latency (ms):         XXXX
Median Chunk Prefill Latency (ms):       XXXX
P99 Chunk Prefill Latency (ms):          XXXX
P100 Chunk Prefill Latency (ms):         XXXX
-------------- Decode Step Latency ---------------
Mean Decode Step Latency (ms):           XXXX
Median Decode Step Latency (ms):         XXXX
P99 Decode Step Latency (ms):            XXXX
P100 Decode Step Latency (ms):           XXXX
---------------- Prefix Cache Hit ----------------
Mean Prefix Cache Hit (%):               XXXX
Median Prefix Cache Hit (%):             XXXX
P99 Prefix Cache Hit (%):                XXXX
P100 Prefix Cache Hit (%):               XXXX
-------------- Left Padding Blocks ---------------
Mean Left Padding Blocks:                XXXX
Median Left Padding Blocks:              XXXX
P99 Left Padding Blocks:                 XXXX
P100 Left Padding Blocks:                XXXX
==================================================
INFO SenDNN metric descriptions written to results/sendnn_bench_metrics_description.txt
INFO Spyre metrics injected into results/spyre-chat-infqps-concurrency4-granite-3.3-8b-instruct-20260811-133931.json
```

## Related Issues

None

## Test Plan

Test the patching works correctly, that the metrics are saved to .json file correctly and also printed correctly

```bash
pytest tests/benchmarks/test_bench_metrics.py -m cpu
```

## Checklist


sducouedic:
bot:test

sducouedic:
bot:test

sducouedic:
bot:test

sendnn_inference/v1/metrics/patch_serving.py
```diff
@@ -0,0 +1,113 @@
...
+        prefix = "data: "
+        data_str = chunk.removeprefix(prefix).rstrip("\n")
+        data = json.loads(data_str)
+        data["spyre_metrics"] = spyre_metrics
+        return f"{prefix}{json.dumps(data)}\n\n"
+    except (json.JSONDecodeError, KeyError, TypeError) as e:
+        logger.warning("Failed to inject spyre_metrics into SSE chunk for %s: %s", request_id, e)
+        return chunk
+
+
```
maxdebayser, line 35: Since vllm uses starlette, is there a way to achieve the same thing by inserting a middleware? https://starlette.dev/middleware/

sducouedic: I had to iterate a bit with Claude to understand middleware, apparently the main difficulty would be to pass the metrics to the context of the middleware, this would still require patching. ie. to populate `scope["state"]` with the metrics we would need to patch the route handler of vllm

sducouedic:
bot:test

sendnn_inference/benchmarks/spyre_bench_serve.py
```diff
@@ -0,0 +1,399 @@
...
+    ]
+    chunk_lats_ms = [
+        lat * 1000 for m in metrics_list for lat in m.get("chunk_prefill_latencies_s", [])
+    ]
+    decode_lats_ms = [lat * 1000 for m in metrics_list for lat in m.get("decode_latencies_s", [])]
+    total_prefill_chunks = sum(num_chunks_list)
+
+    # Scalar summary line (mirrors vllm's plain-count header section)
+    print("{:<40} {:<10}".format("Total prefill chunks processed:", total_prefill_chunks))
+
```
maxdebayser, line 151: Is there a function equivalent to _section that we could import from vllm?

sendnn_inference/v1/metrics/stats_logger.py
```diff
@@ -21,6 +23,63 @@
...
+        with self._lock:
+            self._store[metrics.request_id] = metrics
+
+    def get_and_clear(self, request_id: str) -> SpyreRequestMetrics | None:
+        with self._lock:
+            return self._store.pop(request_id, None)
+
+
+_REGISTRY: SpyreMetricsRegistry | None = None
+_SCHEDULER: Any = None  # set by ChunkedPrefillSpyreScheduler.__init__
```
maxdebayser: Where is _SCHEDULER used? I can only see it being assigned to.

sendnn_inference/v1/metrics/stats_logger.py
```diff
@@ -21,6 +23,63 @@
...
+
+
+def enable_registry() -> SpyreMetricsRegistry:
```
maxdebayser: Actually, I also can't find uses of get_registry and enable_registry. I think I've expanded all source files here in the diff view, but ctrl-f doesn't bring up anything,

sendnn_inference/v1/metrics/patch_serving.py
```diff
@@ -0,0 +1,113 @@
...
+    _original = OpenAIServingChat.chat_completion_stream_generator
+
+    async def _patched_generator(self, request, result_generator, request_id, *args, **kwargs):
+        spyre_metrics: dict | None = None
+
+        async def _capturing_generator():
+            nonlocal spyre_metrics
+            async for res in result_generator:
+                if res.finished and res.kv_transfer_params:
```
maxdebayser, line 64: I see now why you chose `kv_transfer_params`. It's because the output processor forwards those all the way to the `RequestOutput`. Maybe this is also some feedback that we could add to Will Eaton's PR.

sendnn_inference/v1/core/scheduler.py
```diff
@@ -21,6 +23,24 @@
...
+    arrival_ts: dict[str, float] = field(default_factory=dict)
+    first_scheduled_ts: dict[str, float] = field(default_factory=dict)
+    chunk_start_times: dict[str, list[float]] = field(default_factory=dict)
+    decode_latencies: dict[str, list[float]] = field(default_factory=dict)
+    decode_start_times: dict[str, list[float]] = field(default_factory=dict)
+    tkvs: dict[str, list[int]] = field(default_factory=dict)
+    left_padding_blocks: dict[str, list[int]] = field(default_factory=dict)
+    prefill_step_start: float | None = None
+    decode_step_start: float | None = None
+
```
maxdebayser, line 56: Aren't some of these duplicating what vllm already calculates with tracing enabled? e.g. https://github.com/vllm-project/vllm/blob/225936a1dd10586798c0181696d628e7b609ea90/vllm/v1/engine/output_processor.py#L712

maxdebayser (commented):
I haven't looked deep enough into this to be sure, but I have the suspicion that some of the metrics here duplicate what is exported in vllm when [open telemetry tracing](https://docs.vllm.ai/en/v0.10.1/examples/online_serving/opentelemetry.html) is enabled. I don't know if the tracing implementation is flexible enough to add our own metrics, I suspect not, but maybe we can retrieve some of the metrics from there and aggregate them in a post processing step.

sendnn_inference/benchmarks/spyre_bench_serve.py
sducouedic: ah that would be better indeed. So I checked and there is not really a good function. The print is hardcoded for the most part in vllm: 
https://github.com/vllm-project/vllm/blob/9d6fdc2901df490549f9f73d1f10588a6d16721d/vllm/benchmarks/serve.py#L1139-L1178

there is this `process_one_metric` function (https://github.com/vllm-project/vllm/blob/9d6fdc2901df490549f9f73d1f10588a6d16721d/vllm/benchmarks/serve.py#L1251), but it is a closure inside `benchmark()` that uses local variables, and the metric is gated by `selected_percentile_metrics`. Seems simpler to keep `_section` as it is

sendnn_inference/v1/worker/spyre_model_runner.py
```diff
@@ -393,10 +393,10 @@ def vocab_size(self) -> int:
         # self.model here is probably a transformers model class
         if self.model_config.architecture in FMS_POOLING_MODEL_LIST:
             assert isinstance(self.model.config.src_vocab_size, int)
-            return self.model.config.src_vocab_size  # ty: ignore[invalid-return-type]
+            return self.model.config.src_vocab_size
```
sducouedic, line 420: for some reason CI was failing without this change

sendnn_inference/argparse_utils.py
```diff
@@ -162,7 +162,7 @@ def patched_parse_args(
             namespace: argparse.Namespace | None = None,
         ) -> argparse.Namespace:
             result = original_parse_args(self, args, namespace)
-            assert result is not None  # type: ignore[redundant-expr]
+            assert result is not None
```
sducouedic, line 165: for some reason CI was failing without this change

sendnn_inference/v1/metrics/stats_logger.py
sducouedic: `_SCHEDULER` and `_REGISTRY` aren't used at all actually, this was part of an initial plan before using the `kv_transfer_params` hack. I forgot to verify Claude removed this dead code, sorry about that

sendnn_inference/v1/metrics/patch_serving.py
sducouedic: mmh sorry I'm not sure what feedback you have in mind, that we are currently using this `kv_transfer_params` hack, but his engine notification could solve it?

sendnn_inference/v1/core/scheduler.py
sducouedic: Yes the queue_time (ie. `arrival_ts` and `first_scheduled_ts`) is kind of duplicated (I think it is the only one). I was aware of that, but didn't find an obvious way to recover this value in `_free_request()` (I didn't spend too much effort though), but since we track the time (start_time + latency), my thought was that it was simpler to just directly track them in the scheduler additionally, so we have everything gathered in the same place

sducouedic:
@maxdebayser We (I and Claude) had a detailed look at the metrics and investigated if there are potential duplicates:
- (I just added a NOTE comment in the code for this one) **`queued_time_s`** (ie. `first_scheduled_ts` - `arrival_ts`): it is actually different to `queued_time` of the engine ([link](https://github.com/vllm-project/vllm/blob/225936a1dd10586798c0181696d628e7b609ea90/vllm/v1/engine/output_processor.py#L730)): their is defined with `metrics.scheduled_ts - metrics.queued_ts` and ours use `request.arrival_time` which is the API server timestamp. The difference is actually non-neglectible (typically when there is a chunked prefill, the request has to wait until the and of the chunked prefill to be added to the waiting queue), and the latter is better for interpreting the TTFT. (opened an [issue ](https://github.com/vllm-project/vllm/issues/29180)to vllm some time ago). 
- **`chunk_start_times` and `chunk_latencies`**, no duplicates because in vllm there is just a single aggregate "prefill_time", here each chunked prefill is measured separately
- **`decode_start_times` and `decode_latencies`**, there is the `decode_time` metric value ([link](https://github.com/vllm-project/vllm/blob/06fae6911406524ecd89fcce5c6095f920f6b730/vllm/v1/engine/output_processor.py#L732)) but measure an aggregate time between first token and last token. Otherwise there is this `IterationStats.inter_token_latencies_iter` ([link](https://docs.vllm.ai/en/latest/api/vllm/v1/metrics/stats/#vllm.v1.metrics.stats.IterationStats)), but it is still not the same because interleaving makes prefill time contributing to the ITL, altering the values we want to capture

These two metrics are a bit more debatable: 
- `prefix_cache_hit_pct` that gives for each request the number of "saved" chunked prefill (saved thanks to cache hit) / number of chunked prefill that would be necessary without cache hit
- `first_scheduled_ts` that is a timestamp manually done in our scheduler, but that exists already in `RequestStateStats.scheduled_ts` 

For the former, it was more of a demonstration of how to use `/add-bench-metric` it is not really needed now, perhaps I should remove it because it is not reusing the existing implementation at all. But in my opinion it is still worth keeping it as it is, because it is only four localized added lines, and it provides the per-request cache hit all in one place (in the bench output).

For the latter, to obtain the `scheduled_ts` from `RequestStateStats`, Claude says it wouldn't worth it (lot's of changes to capture this value alone):
<details>

> It would take more than it's worth. RequestStateStats lives on RequestState in the OutputProcessor — a separate object in a separate part of the engine, only accessible after the request finishes, and only if log_stats=True. The scheduler's _free_request only receives the Request object, which has no link back to it.
> 
> To use scheduled_ts from there you'd need to either:
> 
> Pass RequestState or its stats into _free_request somehow (breaks the vLLM interface), or
> Read FinishedRequestStats.queued_time in FileStatLogger.record() and ship it separately — but that's back in the stats logger, not in _free_request where the rest of the packing happens, so you'd need a second IPC channel or re-introduce some shared state.
> The current approach — recording arrival_ts from request.arrival_time and first_scheduled_ts from the first time a request appears in ongoing_prefills — is simpler and self-contained. The one real difference from queued_time is that vLLM's version uses monotonic scheduled_ts - queued_ts (event timestamps), whereas yours uses wall-clock first_scheduled_ts - arrival_ts, so they're not numerically identical, but they measure the same thing.
> 
> The observation is correct in principle but the plumbing cost to eliminate the duplication is high enough that keeping the current approach is the right call. Worth a comment in the code noting the overlap so the next person doesn't re-investigate.

</details>

So at the end the OpenTelemetry traces are quite limited in information (total prefill time and total decode time), not really enough for this level of detail

sendnn_inference/benchmarks/spyre_bench_serve.py
```diff
@@ -0,0 +1,509 @@
...
+    result["spyre_decode_latencies_s"] = [m.get("decode_latencies_s", []) for m in metrics_list]
+    result["spyre_decode_start_times_s"] = [m.get("decode_start_times_s", []) for m in metrics_list]
+    result["spyre_tkvs"] = [m.get("tkvs", []) for m in metrics_list]
+    result["spyre_prefix_cache_hit_pct"] = [
+        m.get("prefix_cache_hit_pct", 0.0) for m in metrics_list
+    ]
+    result["spyre_left_padding_blocks"] = [m.get("left_padding_blocks", []) for m in metrics_list]
+    result["spyre_pause_latencies_s"] = [m.get("pause_latencies_s", []) for m in metrics_list]
+    result["spyre_pause_start_times_s"] = [m.get("pause_start_times_s", []) for m in metrics_list]
+    result["spyre_was_missing_blocks"] = [m.get("was_missing_blocks", False) for m in metrics_list]
```
yannicks1: hmm, is it possible that `metrics_list` here ordered by request completion time, but the file constructed by vLLM is ordered by submission time?

sendnn_inference/v1/metrics/patch_serving.py
```diff
@@ -0,0 +1,113 @@
...
+def _patch_chat() -> None:
+    try:
+        from vllm.entrypoints.openai.chat_completion.serving import OpenAIServingChat
+    except ImportError:
+        logger.warning("Could not import OpenAIServingChat — chat serving patch skipped")
+        return
+
+    _original = OpenAIServingChat.chat_completion_stream_generator
+
+    async def _patched_generator(self, request, result_generator, request_id, *args, **kwargs):
```
yannicks1, line 58: these positional args are a bit fragile if upstream signature changes. Can we add a compat test to detect this?

sendnn_inference/benchmarks/spyre_bench_serve.py
```diff
@@ -0,0 +1,509 @@
...
+    else:
+        try:
+            candidates = [
+                os.path.join(result_dir, f)
+                for f in os.listdir(result_dir)
+                if f.endswith(".json")
+                and os.path.getmtime(os.path.join(result_dir, f)) >= run_started_at
+            ]
+        except OSError:
+            candidates = []
```
yannicks1: this seems risky. can we just always set an `explicit_name`?

sendnn_inference/benchmarks/spyre_bench_serve.py
```diff
@@ -0,0 +1,509 @@
...
+
+    stdout_buf = io.StringIO()
+    stderr_buf = io.StringIO()
+    orig_stdout, orig_stderr = sys.stdout, sys.stderr
+    done = {"v": False}
+
+    class _StdoutSplitter:
+        def write(self, s):
+            if not done["v"]:
+                if s.strip() == "=" * 50:
```
yannicks1: also risky if that formatting changes upstream. We could add an compat test to detect this too.

sendnn_inference/benchmarks/spyre_bench_serve.py
sducouedic: ah this is a very good point, I'll check that

sendnn_inference/v1/metrics/patch_serving.py
sducouedic: added some tests to catch upstream changes in CI

sendnn_inference/benchmarks/spyre_bench_serve.py
sducouedic: Now I force the user to set `--result-dir` and `--result-filename` whenever `--save-result` is set

sendnn_inference/benchmarks/spyre_bench_serve.py
sducouedic: Added a compat test so that if this changes in upstream, the CI will detect it

sducouedic:
Thanks for the review @yannicks1 I addressed your comments

yannicks1 (approved):
lgtm, thanks for this addition. will be very useful to get a better understanding of HW perf!

PR #1011, opened 2026-06-18
Fix decode deployment

yannicks1:
## Description

We found that there is an overhead present for the first decode pass only. By doing an additional decode pass during warmup this ITL spike can be eliminated. This ensure the compiled decode program is also installed on the device before runtime. Note that this mirrors the prefill deployment. 

## Test Plan

Tested with several model on the device. Also did profiling and saw the additional ops disappear after this PR.  

## Checklist


yannicks1:
bot:test

joerunde (approved):
Assuming this works on spyre, let's ship it!

(We should have thought of this back when we did the one extra prefill)

yannicks1:
bot:test

maxdebayser (approved):

PR #1013, opened 2026-06-19
Feature: support --load-format dummy for random-init model loading

yannicks1:
## Description

Adds support for loading any causal-LM model with `--load-format dummy`, so weights are random-initialized and no checkpoint needs to be downloaded.

Note: Pooling and multimodal are out of scope (separate code paths). 

## Related Issues

This substantially simplifies model enablement and bench-marking without an available checkpoint. 

## Test Plan

added `tests/e2e/test_load_format_dummy.py`

## Checklist

### DIFF
--- a/sendnn_inference/model_executor/model_loader/spyre.py
+++ b/sendnn_inference/model_executor/model_loader/spyre.py
@@ -72,6 +72,7 @@ class SpyreCausalLM(nn.Module):
         self.parallel_config = vllm_config.parallel_config
         self.cache_config = vllm_config.cache_config
         self.scheduler_config = vllm_config.scheduler_config
+        self.load_config = vllm_config.load_config
         self.dtype = self.get_dtype()
 
         # Wrappers for utils for multimodal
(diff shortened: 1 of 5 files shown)


maxdebayser:
Do you still need a config.json in the repository? Will it be downloaded automatically if not present locally?

yannicks1:
> Do you still need a config.json in the repository? Will it be downloaded automatically if not present locally?

Yes, you either point `--model` at a local directory containing a config.json, or you pass a valid HF model identifier and it'll fetch config.json on demand (no weight files) if it isn't already in the cache.

maxdebayser (approved):

sendnn_inference/model_executor/model_loader/spyre.py
```diff
@@ -171,16 +172,30 @@ def load_weights(
...
-                revision=model_config.revision,
+        # `--load-format dummy` skips the checkpoint download and routes through
+        # FMS's `hf_configured` path, which fetches only config.json and then
+        # random-inits the model via `reset_parameters()`.
+        variant: str | None = None
+        if self.load_config.load_format == "dummy":
+            logger.info(
+                "Loading model %s with random weights.",
+                model_config.model,
             )
```
sducouedic, line 183: nit but in line 258, `model_path` is used in `spyre_mm.maybe_get_mm_utils`
if dummy model isn't working with multimal model, maybe we should raise an error

```suggestion
        if self.load_config.load_format == "dummy":
            if self.is_multimodal:
                raise ValueError("dummy format not supported for multimodal models")
            logger.info(
                "Loading model %s with random weights.",
                model_config.model,
)
```

tests/e2e/test_load_format_dummy.py
```diff
@@ -0,0 +1,39 @@
...
+@pytest.mark.cpu
+def test_load_format_dummy_generates_tokens(
+    model: ModelInfo,
+    max_model_len: int,
+    max_num_seqs: int,
+    max_num_batched_tokens: int,
+):
+    """The dummy load path random-inits weights and still produces tokens
+    (output content is meaningless and not checked)."""
+    envs_spyre.override("SENDNN_INFERENCE_DYNAMO_BACKEND", "eager")
```
sducouedic, line 19: why not using `monkeypatch.setenv` here? this is how we usually do right?

sducouedic (commented):
Just the monkeypatch question for me, otherwise look good

tests/e2e/test_load_format_dummy.py
joerunde: tomato tomato?

Functionally the difference here is that `.override` for the environment variables we manage ourselves ensures that the cached value is cleared so that we don't accidentally read a stale value that was put into the cache by a different test. IDK if that actually matters here, since all tests running on GHA would be only using the eager backend anyway, and this would be skipped on runs for spyre hardware

joerunde:
@yannicks1 do we need a guard or warning log if this is used with pooling models?

tests/e2e/test_load_format_dummy.py
sducouedic: okay, I had a look and `envs_spyre.override` was also used in other tests files, I thought we were only using monkey patching, so found it odd. If monkey patching is generally better practice to avoid these leaking environment variables, I would still change that but up to you @yannicks1

yannicks1: agree, leaving it as also consistent with other e2e tests (test_profiler.py, test_stats_logger.py .. )

PR #1017, opened 2026-06-25
SWA Granite HF registration

yannicks1:
## Description

Drops the dependency on the granite-swa-model transformers [fork](https://github.com/Bharat-Runwal/transformers/tree/granite-swa-model). Routes `granite_swa` through the generic FMS hf_pretrained/hf_configured path. Pairs with [this](https://github.com/foundation-model-stack/foundation-model-stack/pull/538) FMS registration PR. 

## Related Issues
none

## Test Plan
none

## Checklist


maxdebayser (approved):

joerunde (approved):
We could consider adding a test here that loads a model config with `granite_swa` model_type to flex these code paths. (That might require whipping up a super tiny dummy model, which our good friend claude is also pretty good at.)

Not totally required for the initial merge here, but I would lean towards including because we've shipped a lot of regressions in the past for models like mistral that we didn't have load-path test cases for in the github CI

yannicks1:
Thanks for reviewing! Added some tests and hit auto-merge :)

PR #1021, opened 2026-06-25
Remove Qwen3 models from CI tests and docs

maxdebayser:


sendnn_inference/config/model_configs.yaml
```diff
@@ -338,30 +338,6 @@ models:
           - prompt_len: 512
             batch_size: 64
 
-  Qwen/Qwen3-Embedding-0.6B:
```
joerunde, line 341: should we instead put an if guard in the docs generator to remove this from the docs? That way if somebody does want to test qwen models they'll still work?

maxdebayser: Actually there is already a `EXCLUDED_MODELS` list. I'm adding the models there now and restoring the configurations.

joerunde (approved):
ship it

PR #1023, opened 2026-06-29
Force pooling models to load in float16

maxdebayser:
Since [transfomers 5](https://github.com/huggingface/transformers/releases/tag/v5.0.0#:~:text=Default%20dtype%20update) the model load in the dtype that is specified in config.json instead of float32

The full mask expansion silences a torch_sendnn warning but it doesn't seem to be necessary.  ~I'm still benchmarking to see if it affects the performance.~ And it's slower, so for now we let sendnn print the warning.

### DIFF
--- a/sendnn_inference/v1/worker/spyre_model_runner.py
+++ b/sendnn_inference/v1/worker/spyre_model_runner.py
@@ -672,14 +672,10 @@ class SpyrePoolingModelRunner(
 
                 hidden_states = outputs[0]
             else:
-                mask = model_input.input_masks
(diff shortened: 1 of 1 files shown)


maxdebayser:
bot:test

joerunde:
Should we retain the same behavior by explicitly loading in fp32? Just wondering if there was a reason we weren't using fp16 before, but it could have just been a miss

Daniel-Schenker (approved):
lgtm

joerunde:
spyre tests passed- ship it!

PR #1024, opened 2026-06-30
Fix override for unknown pooling models

maxdebayser:
For unknown models we cap the max-model-len and the max-num-seqs to defaults that are known to work with several generative models. But this doesn't work well for pooling models. The defaults provided by the warmup shapes are cood enough.


joerunde (approved):
lpgtm, but let's get the full integration tests run on it

joerunde:
bot:test
MARKERS="spyre and scoring"

PR #1026, opened 2026-07-06
chore: bumps vllm to v0.24.0

gkumbhat:
Defers tokenizer registry patch from module-level to check_and_update_config to fix a circular import introduced in vLLM 0.24.0, where torch_utils.py now imports is_pin_memory_available at module-level (triggering platform resolution before torch_utils finishes initializing).

## Description

Enable `0.24.0` vLLM

1. vLLM upstream PR https://github.com/vllm-project/vllm/pull/42656 seems to have slightly changed how blocks gets prepended, which seems to fail the tests. 
2. vLLM upstream PR https://github.com/vllm-project/vllm/pull/40299 seems to have fixed the mistral model file parsing issue, thus with move to `0.24.0` we can remove the patch we added for that entirely.

## Checklist


sendnn_inference/platform.py
```diff
@@ -901,5 +912,10 @@ def _compute_config_format(namespace: argparse.Namespace) -> str:
...
 # suppressed since it's expected for models not in the registry.
-# This must be done at import time to be applied to spawned worker processes.
-SpyrePlatform._patch_tokenizer_registry_get_config()
+#
+# In vLLM >= 0.24.0, torch_utils.py imports is_pin_memory_available() at
+# module-level, which triggers platform resolution before torch_utils finishes
+# initializing. Importing vllm.tokenizers.registry here would cause a circular
+# import, so the call is deferred to SpyrePlatform.check_and_update_config.
+# Workers use model execution paths, not the tokenizer registry, so they do
+# not need this patch applied at module-level.
```
joerunde: @gkumbhat this entire comment block should be moved to `check_and_update_config` then

Also- I'm really dubious about this working since the previous comment explains that this had to be here in order to be picked up by the spawned worker processes. If those _don't_ run `check_and_update_config` because the config is simply serialized and sent to them, then the patch won't apply

joerunde:
bot:test

gkumbhat:
bot:test

joerunde (approved):
ship it

PR #1027, opened 2026-07-07
Model config_format and tokenizer_mode conditional defaults fix

Daniel-Schenker:
## Description

This PR updates the logic used to apply conditional defaults for `config_format` and `tokenizer_mode`. 

The previous implementation was incorrectly assigning mistral formatting to Llama models from HF as it was recursively searching the model files for a `params.json` file which was present in a subdirectory of some Llama models.

The new implementation uses `is_mistral_model_repo` which checks for `consolidated*.safetensors` file which is a more accurate check for mistral models.

## Test Plan

Tested with various supported models and it correctly detects mistral for supported mistral family models while defaulting to auto for non-mistral models.

## Checklist


joerunde (approved):
lpgtm

PR #1028, opened 2026-07-10
Set attn backend to eager for pooling models

maxdebayser:
With transformers 5 the default became "sdpa" and that is causing very long compilation times as well as
compiler crashes


gkumbhat (approved):
Looks good to me!

PR #1031, opened 2026-07-15
Improve model config/tokenizer format defaults

Daniel-Schenker:
## Description

This PR improves upon the faulty logic used in https://github.com/torch-spyre/sendnn-inference/pull/1027. 

The previous fix attempted to exclude Llama models from mistral-format detection by checking whether params.json existed only inside an original/ subdirectory, using two calls to `any_pattern_in_repo_files`. This did not work as the vLLM implementation of `any_pattern_in_repo_files` delegates to `list_filtered_repo_files`, which matches patterns using `fnmatch.fnmatch(os.path.basename(file), pattern)`. This always returns false since `os.path.basename` is applied to every file path before fnmatch runs. Llama models with only original/params.json were still being assigned mistral format.

fix: correctly detect root-level params.json for mistral format detection. 

Replace both `any_pattern_in_repo_files calls` with a single `list_repo_files` call followed by an exact string membership test.

## Checklist

### DIFF
--- a/sendnn_inference/platform.py
+++ b/sendnn_inference/platform.py
@@ -21,7 +21,6 @@ import os
 from typing import TYPE_CHECKING, cast, Literal
 
 import torch
-import huggingface_hub
 from vllm.logger import init_logger
 from vllm.utils.argparse_utils import FlexibleArgumentParser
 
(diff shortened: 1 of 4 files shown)


joerunde (approved):
looks reasonable to me!

🤞 for all the integration tests

PR #1038, opened 2026-07-24
:bug: Fix granite vision not working with tp=1

gkumbhat:
## Description

Currently TP=1 was not supported for separate mm encoder processing. This would lead to a weird situation for granite vision model, which we run on TP=1 and have `SENDNN_INFERENCE_ASYNC_MM_ENCODER=1` by default.

In this PR I have removing that condition to only run separate encoder process when we have TP=1.. This is safe to do (as I tested) and going forward it will only be controlled by env variable and not TP condition.

## Checklist

### DIFF
--- a/sendnn_inference/__init__.py
+++ b/sendnn_inference/__init__.py
@@ -30,12 +30,6 @@ if importlib.util.find_spec("torch_nnpa") is not None:
     os.environ.setdefault("TORCH_DEVICE_BACKEND_AUTOLOAD", "0")
 
 import importlib.metadata  # noqa: E402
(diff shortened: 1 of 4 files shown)


Daniel-Schenker (approved):
lgtm

sendnn_inference/__init__.py
```diff
@@ -45,6 +39,14 @@ def register():
     return "sendnn_inference.platform.SpyrePlatform"
 
 
+import json  # noqa: E402
```
gkumbhat, line 42: To avoid circular import error when running smoke test.

PR #1040, opened 2026-07-27
Add platform-based model filtering

R3hankhan123:
## Description
Introduce a `platforms` field on each model entry in model_configs.yaml so the registry can filter supported models by architecture (x86_64, s390x, ppc64le) at runtime using platform.machine().
## Checklist

### DIFF
--- a/sendnn_inference/config/model_configs.yaml
+++ b/sendnn_inference/config/model_configs.yaml
@@ -3,6 +3,8 @@
 #
 # Each model entry contains:
 # - architecture: Pattern for matching HuggingFace model configs
+# - platforms: List of supported CPU architectures (e.g. [x86_64, s390x, ppc64le]).
+#              If omitted, the model is supported on all platforms.
 # - static_batching_configs: List of static batching configurations (optional)
 # - continuous_batching_configs: List of continuous batching configurations (optional)
 #   - Each CB config may have a nested device_config for device-specific settings
@@ -177,6 +179,7 @@ models:
   # Granite 3.3 8B Instruct
   ibm-granite/granite-3.3-8b-instruct:
     architecture: *granite_33_8b_architecture
+    platforms: [x86_64, ppc64le, s390x]
 
     # Continuous batching configurations
     continuous_batching_configs:
@@ -200,6 +203,7 @@ models:
       <<: *granite_33_8b_architecture
       quantization_config:
         format: float-quantized
+    platforms: [x86_64, ppc64le, s390x]
 
     # Continuous batching configurations only
     continuous_batching_configs:
@@ -218,6 +222,7 @@ models:
   # Llama 3.1 8B Instruct
   meta-llama/Llama-3.1-8B-Instruct:
     architecture: *llama3_8b_architecture
+    platforms: [x86_64, ppc64le]
     continuous_batching_configs:
       - tp_size: 1
         max_model_len: 3072
@@ -237,6 +242,7 @@ models:
       <<: *granite_4_8b_architecture
       model_type: granitemoehybrid
       num_experts_per_tok: 0
+    platforms: [x86_64, ppc64le, s390x]
 
     # Continuous batching configurations
     continuous_batching_configs:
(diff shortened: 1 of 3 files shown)


dilipgb (approved):
LGTM

R3hankhan123:
bot:test
MARKERS="spyre and not quantized and not multi and chunked_prefill"

R3hankhan123:
bot:test
MARKERS="spyre and not quantized and not multi and chunked_prefill"

dilipgb:
@joerunde please review. Can we mark this PR for 2.5.2 release? Is there a milestone or something which we can add so we don't cut release?

sendnn_inference/config/model_configs.yaml
```diff
@@ -185,6 +188,7 @@ models:
       <<: *granite_33_8b_architecture
       quantization_config:
         format: float-quantized
+    platforms: [x86_64, ppc64le, s390x]
```
joerunde, line 206: cc @gkumbhat do we want `x86_64` or `amd64`?

joerunde:
Looks like the spyre-ci is failing- are the failures related or are they a different problem?

joerunde:
> Can we mark this PR for 2.5.2 release? Is there a milestone or something which we can add so we don't cut release?

We haven't been tracking release milestones, just coordinating together on when anything needs releasing. The group of people coordinating that right now should be you, @Daniel-Schenker, @gkumbhat, and maybe @yannicks1 (for any SWA support items).

sendnn_inference/config/model_configs.yaml
gkumbhat: Different vendors unfortunately use different nomenclatures. But here, since we would detect the version by `platform.machine()`, `x86_64`  is fine to use for linux based systems.

sendnn_inference/config/model_registry.py
```diff
@@ -145,10 +146,19 @@ def find_matching_model(self, vllm_model_config: "VllmModelConfig") -> ModelConf
             logger.debug("No HF config available for matching")
             return None
 
+        current_platform = platform.machine()
         best_match: ModelConfig | None = None
         best_field_count = -1
 
         for model_name, model_config in self._models.items():
+            if model_config.platforms and current_platform not in model_config.platforms:
+                logger.debug(
```
gkumbhat: @R3hankhan123 what is intended behavior here. With this we are just not loading the config and logging the message at debug level. This means, it won't necessarily restrict model from loading (might result in other forms of errors - which can get confusing)

R3hankhan123: If we want to restrict the model then https://github.com/torch-spyre/sendnn-inference/blob/74278cb4c9f8571d43c181f51e3cf10cc2a435cb/sendnn_inference/envs.py#L25 should be defaulted to true then?

gkumbhat: Ah missed this variable exist. But I guess, I am missing how this helps. Like if someone is trying to run particular model on unsupported platform, than with this if condition, they _might_ get compilation failure. instead of model working correctly.

So may be if a platform wants to restrict based on known configs, then they should set this env variable to True on their side.

R3hankhan123: Yes because when setting this env variable it was preventing me to deploy unsupported models on s390x

gkumbhat: right and thats what we want right?

gkumbhat: or do you just want to show user warning but still allow them to deploy? In which case, I would recommend changing this log message to warning

R3hankhan123: I would defer this to @dilipgb

sendnn_inference/config/model_registry.py
```diff
@@ -145,10 +146,19 @@ def find_matching_model(self, vllm_model_config: "VllmModelConfig") -> ModelConf
...
         best_field_count = -1
 
         for model_name, model_config in self._models.items():
+            if model_config.platforms and current_platform not in model_config.platforms:
+                logger.warning(
+                    "Skipping model '%s': not supported on platform '%s'",
+                    model_name,
+                    current_platform,
+                )
+                continue
```
joerunde: Motivation is a documentation only change- let's update the warning message here and remove this continue to not break any existing deployments

gkumbhat (approved):
Looks good to me!

PR #1043, opened 2026-07-31
Add logic for ppc64le to use logical cores when using psutil cpu_count

ktuvw:
## Description

For ppc64le we want to use logical cpus to determine cpus per worker in multimodal models

## Related Issues

#1945

## Test Plan

Start a vllm server on a machine with <36 physical cores and check cpus per worker is set correctly

## Checklist

### DIFF
--- a/sendnn_inference/platform.py
+++ b/sendnn_inference/platform.py
@@ -667,9 +667,7 @@ class SpyrePlatform(Platform):
                 else:
                     use_logical_cpus = False
                 cpu_count = float(psutil.cpu_count(logical=use_logical_cpus))
-                detection_message = (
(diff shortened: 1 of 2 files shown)


ktuvw:
@gkumbhat looks like it's failing at tests/utils/test_platform_cpu_count.py due to 
```
assert "physical" in msg
E       AssertionError: assert 'physical' in 'Detected 6.0 CPUs from psutil.cpu_count(logical=False)'
```
Can I take out this assert or should I add logic like {"logical" if use_logical_cpus else "physical"} to the print?

gkumbhat (approved):
Looks good to me! 

@ktuvw I think adding the if condition would be better to match the logic of platform.py.

Can you also fix the typing issue please.

gkumbhat (approved):
Looks good to me!

PR #1045, opened 2026-08-06
:sparkles: Add precomputed noise pool for sampler

gkumbhat:
## Description

- Add optional exponential-noise pool for the custom `SpyreSampler` to avoid slow per-step `Tensor.exponential_()` random number generation on `s390x` platform
    - `ExponentialNoisePool` - fills a buffer of Exp(1) noise once at model load. `draw()` returns a `random-offset` slice, `draw_row()` returns a seed-selected slice.
    - `SpyreTopKTopPSampler` - pins `forward_native`, sources sampling noise from the pool (falls back to upstream random_sample when no pool)
    - `SpyreSampler` - swaps the pooled top-k/top-p sampler into vLLM's Sampler; all other sampler behavior inherited unchanged 
- Oversized batches `(batch × vocab` > `pool`) fall back to upstream random_sample with a one-time warning instead of erroring
- Wire `SpyreSampler` into `SpyreCausalLM` via `_build_sampler`. This behaves identically to upstream when the pool is disabled
- New env variables:
    - `SENDNN_INFERENCE_USE_NOISE_POOL` (bool, default 0) - enable the pool
    - `SENDNN_INFERENCE_NOISE_POOL_MULTIPLIER` (int, default 32) - pool size = multiplier × max_num_seqs × vocab_size
    - `SENDNN_INFERENCE_SAMPLER_TIMING` (int, default 0) - log average sampling latency every N calls

- Pool dtype follows the existing platform defaults for consistency and correctness

### Notes

If request contains seed, then we use per-request generator to pick a deterministic pool offset (reproducible per seed, skips bulk generation) instead of regenerating noise. So this is still cheaper than random generation and aligns with non-seeded request for performance.

## Test Plan

- New `tests/v1/sample/test_spyre_noise_pool.py` (CPU-marked), covering:
    - Pool draw shape, dtype, and value membership
    - Pool dtype follows the requested dtype
    - Pool-too-small raises, and non-positive pool size raises
    - Oversized batch falls back to on-the-fly noise instead of raising
    - Peaked distribution samples deterministically (argmax)
    - Uniform distribution covers the vocab without dominance
    - Seeded row is reproducible across runs with the same seed
    - Seeded row reads pool values (seed-selected offset), not fresh noise
    - Different seeds generally produce different tokens
    - No-pool path matches upstream `random_sample` bit-for-bit given the same RNG state
    - `SpyreSampler` correctly wires the pool into `SpyreTopKTopPSampler`

## Checklist

### DIFF
--- a/sendnn_inference/envs.py
+++ b/sendnn_inference/envs.py
@@ -30,12 +30,25 @@ if TYPE_CHECKING:
     SENDNN_INFERENCE_TP_MM_SHARING: bool = True
     SENDNN_INFERENCE_LONG_OUT_PRIO: bool = False
     SENDNN_INFERENCE_PAUSING_ENABLED: bool = True
+    SENDNN_INFERENCE_USE_NOISE_POOL: bool = False
+    SENDNN_INFERENCE_NOISE_POOL_MULTIPLIER: int = 32
+    SENDNN_INFERENCE_SAMPLER_TIMING: int = 0
 
 logger = init_logger(__name__)
 
 _cache: dict[str, Any] = {}
 
 _CPU_MM_DTYPE_PLATFORM_DEFAULTS = {"s390x": "float32", "ppc64le": "float32"}
+
+
+def get_noise_pool_dtype() -> torch.dtype:
+    """Resolve the dtype for the sampler's exponential-noise pool.
+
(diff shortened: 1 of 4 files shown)


gkumbhat:
closed by #1046

PR #1046, opened 2026-08-17
feat(sampling): Add async ring buffer and Spyre-optimized samplers

dmholtz:
## Description

This PR optimizes the Spyre sampling path in three stages to remove the main latency bottlenecks from token generation while preserving correctness.
To ease the review, each stage is a separate commit.

### Problem
The default sampling path spends too much time in the critical decoding loop on CPU-heavy work:
- Exponential noise generation is computed synchronously in the sampling path.
- Softmax normalization remains expensive for large vocabularies.
- Tensor-parallel CPU ranks redundantly sample and diverge from one another.

### Solution
1. Stage 1: Async noise pre-sampling
   - Introduce an asynchronous ring buffer that pre-generates exponential noise on a background thread.
   - Replace synchronous `exponential_()` calls in the critical path with zero-copy buffer access.
   - Add `SpyreTopKTopPSampler` and `SpyreSampler` integration for the Spyre path.

2. Stage 2: Single-rank sampling on TP
   - Only TP rank 0 performs sampling and broadcasts the selected token IDs to the remaining ranks.
   - This removes redundant CPU work and keeps all ranks aligned on the same tokens.

3. Stage 3: Log-space Gumbel sampling
   - Reformulate sampling as `argmax(logits - log_noise)` instead of `argmax(log(softmax(logits)) + log_noise)`.
   - This preserves sampling behavior while removing softmax from the critical path.

### Key Insight
The three optimizations are complementary: async noise generation removes expensive background work from the hot loop, TP rank gating removes redundant computation and divergence, and log-space Gumbel sampling removes softmax entirely. Together they reduce sampling latency without changing output distribution.
Performance tests show that OTPS increase by >30% for Granite 4.1 while staying in a fixed, thight compute budget.

## Related Issues

n/a

## Test Plan

- Unit tests to verify ring-buffer correctness and sampling correctnes
- Performance tests

## Checklist

- [x] I have read the contributing guidelines
- [x] My code follows the project's code style (`bash format.sh`)
- [x] I have added tests for my changes
- [ ] I have updated the documentation where needed

### DIFF
--- a/sendnn_inference/model_executor/model_loader/spyre.py
+++ b/sendnn_inference/model_executor/model_loader/spyre.py
@@ -17,12 +17,12 @@ from vllm.logger import init_logger
 from vllm.model_executor.model_loader.weight_utils import download_weights_from_hf
 from vllm.v1.outputs import SamplerOutput
 from vllm.v1.sample.metadata import SamplingMetadata
-from vllm.v1.sample.sampler import Sampler
 
 import sendnn_inference.envs as envs_spyre
 import sendnn_inference.multimodal as spyre_mm
 import sendnn_inference.utils as utils_spyre
 from sendnn_inference.platform import SpyrePlatform
+from sendnn_inference.v1.sample.spyre_sampler import SpyreSampler
 
 try:
     import backends.dynamo_tracer  # ty: ignore[unresolved-import] # noqa
@@ -113,7 +113,8 @@ class SpyreCausalLM(nn.Module):
     ) -> None:
         super().__init__()
 
-        self.sampler = Sampler()
+        # SpyreSampler is a vLLM Sampler subclass that uses top-k/top-p sampling implementations optimized for Spyre platform.
+        self.sampler = SpyreSampler(vllm_config=vllm_config)
 
         # boolean tensor of length batch size with indices:
         # True for unfinished sequences and
--- a/sendnn_inference/v1/sample/spyre_topk_topp_sampler.py
+++ b/sendnn_inference/v1/sample/spyre_topk_topp_sampler.py
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the sendnn-inference project
+
+import warnings
+
+import torch
+import torch.nn as nn
+from vllm.config.model import LogprobsMode
+from vllm.v1.sample.ops.topk_topp_sampler import apply_top_k_top_p
+
+from sendnn_inference.v1.sample.async_ring_buffer import AsyncExponential_RingBuffer
+
+
+class SpyreTopKTopPSampler(nn.Module):
+    """Top-k/top-p sampler optimized for Spyre hardware via asynchronous noise pre-sampling.
+
+    This removes CPU-bound noise generation from the latency-critical sampling path by
+    pre-drawing noise into a ring buffer that the decoder can consume via zero-copy views
+    during token selection. The buffer is pre-allocated based on vocab_size and multiples
+    of max_batch_size to support zero-copy access patterns.
+    """
+
+    def __init__(
+        self,
+        vocab_size: int,
+        max_batch_size: int,
+        logprobs_mode: LogprobsMode = "raw_logprobs",
+    ):
+        """Initialize the SpyreTopKTopPSampler with a asynchronous exponential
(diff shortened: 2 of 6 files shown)


tests/v1/sample/test_spyre_topk_topp_sampler.py
```diff
@@ -0,0 +1,76 @@
+# SPDX-License-Identifier: Apache-2.0
```
dilipgb, line 1: nitpick: correct the file name: test_spyre_topk_topp_sampler

dilipgb (changes requested):

dilipgb:
Otherwise looks good to me. Pele tests passed fine. Maybe we wanted to run e2e tests on card once? @gkumbhat can you please take a look at this once?

tests/v1/sample/test_spyre_topk_topp_sampler.py
dmholtz: Done

dmholtz:
Thank you for your review @dilipgb, let's run another round of tests to get this PR green

dilipgb:
@joerunde would take look at this PR once?

tdoublep:
Can we rebase the PR against the new version of vLLM to check there is no interaction?

dmholtz:
@tdoublep I rebased this PR to main, it went smooth without any conflicts. Could you please re-run the tests to confirm the behavior with vLLM v0.27.1?

tdoublep (approved):
LGTM - thanks for the awesome work!

tdoublep:
@dilipgb Please take a final look (you requested changes, which is blocking merge). Tests are passing against latests main

sendnn_inference/model_executor/model_loader/spyre.py
```diff
@@ -112,8 +114,18 @@ def __init__(
         rank: int,
     ) -> None:
         super().__init__()
-
-        self.sampler = Sampler()
+        vllm_config_compatible = SpyreSampler.is_vllm_config_compatible(vllm_config)
+        if not vllm_config_compatible:
+            warnings.warn(
```
R3hankhan123: Can we use vllm's logger here instead of warnings

sendnn_inference/v1/sample/spyre_topk_topp_sampler.py
```diff
@@ -0,0 +1,87 @@
...
+        self,
+        logits: torch.Tensor,
+        generators: dict[int, torch.Generator],
+        k: torch.Tensor | None,
+        p: torch.Tensor | None,
+    ) -> tuple[torch.Tensor, torch.Tensor | None]:
+        """Apply top-k/top-p filtering and sample tokens using pre-drawn noise."""
+
+        if generators:
+            warnings.warn(
```
R3hankhan123: Can we use vllm's logger here instead of warnings

R3hankhan123 (commented):
1-2 nits rest it looks good to me

sendnn_inference/v1/sample/spyre_topk_topp_sampler.py
dmholtz: @R3hankhan123 Thank you for the suggestion. I intentionally used warnings.warn() here because this is intended to be a Python warning that callers can filter/control, rather than diagnostic logging. Does vLLM's logger support the same (to avoid spamming the logs in every forward)? Since this is a NIT, I would leave it unchanged.

sendnn_inference/model_executor/model_loader/spyre.py
dmholtz: This is part of the constructor, so my argument from below does not really apply here...

dmholtz: @R3hankhan123 I pushed a commit to replace the warnings.warn call here. Please re-run the CI.

sendnn_inference/v1/sample/spyre_topk_topp_sampler.py
R3hankhan123: yup you can use warning_once function call in vllm's logger

dmholtz:
@R3hankhan123 Thanks for your review, I addressed both NITs.

R3hankhan123 (approved):
LGTM

dilipgb (approved):
LGTM

